PR #24, opened 2016-05-10
Added new asList() to ParseItemList

gertjanal:


src/main/java/io/parsingdata/metal/data/ParseItemList.java
```diff
@@ -82,4 +85,19 @@ public String toString() {
         return isEmpty() ? "" : ">" + head + tail.toString();
     }
 
+    /**
+     * Return all {@link ParseGraph}s as list by iterating recursively.
+     * @return all {@link ParseGraph}s in <code>this</code> {@link ParseItemList}
+     */
+    public List<ParseGraph> asList() {
```
rdvdijk: Should probably be part of a `metal-utils`. @jvdb We should start that and put this and put some of @ccreeten's cool util methods in there as well.

rdvdijk: Why not just `List<ParseItem>`?

jvdb:
I agree with @rdvdijk, this is a great thing to have but needs to be put into a util library. So closed but kept :)

PR #32, opened 2016-05-24
Add a Let token to hold calculated values

gertjanal:
Token based on a zero byte value. Might not be exactly what we need, but works for now. If we agree on the impl, I'll add some tests


jvdb:
Thanks for implementing this approach. It provides some good insights into the problem/solution space. As we've discussed, this won't be the actual implementation for Metal since regular `ParseValue` instances need to directly map to parsed values due to serialization/deserialization and traceability issues.

So I will close this PR, but keep it for reference purposes :+1:

PR #33, opened 2016-05-25
Added Len for counting the length of the byte array

gertjanal:
Couldn't find any byte length expression, let me know if I was right so I can create tests

### DIFF
--- a/src/main/java/io/parsingdata/metal/expression/value/reference/Len.java
+++ b/src/main/java/io/parsingdata/metal/expression/value/reference/Len.java
@@ -20,7 +20,7 @@ public class Len implements ValueExpression {
     @Override
     public OptionalValue eval(final Environment env, final Encoding enc) {
         final ParseValue value = env.order.get(_name);
-        return OptionalValue.of(value == null ? null : ConstantFactory.createFromNumeric(value.getValue().length, new Encoding(true)));
+        return OptionalValue.of(value == null || value.getValue() == null ? null : ConstantFactory.createFromNumeric(value.getValue().length, new Encoding(true)));
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


src/main/java/io/parsingdata/metal/expression/value/reference/Len.java
```diff
@@ -0,0 +1,31 @@
...
+    private final String _name;
+
+    public Len(final String name) {
+        _name = checkNotNull(name, "name");
+    }
+
+    @Override
+    public OptionalValue eval(final Environment env, final Encoding enc) {
+        final ParseValue value = env.order.get(_name);
+        return OptionalValue.of(value == null ? null : ConstantFactory.createFromNumeric(value.getValue().length, new Encoding(true)));
```
gertjanal: null check

src/main/java/io/parsingdata/metal/expression/value/reference/Len.java
```diff
@@ -0,0 +1,35 @@
...
+
+import io.parsingdata.metal.data.Environment;
+import io.parsingdata.metal.data.ParseValue;
+import io.parsingdata.metal.encoding.Encoding;
+import io.parsingdata.metal.expression.value.ConstantFactory;
+import io.parsingdata.metal.expression.value.OptionalValue;
+import io.parsingdata.metal.expression.value.Value;
+import io.parsingdata.metal.expression.value.ValueExpression;
+
+public class Len implements ValueExpression {
```
jvdb: This should probably extend `UnaryValueExpression`.

gertjanal: updated

jvdb:
Implements part of #37.

PR #36, opened 2016-06-13
Keep encoding when using ConstantFactory (old repo)

gertjanal:
Please see tests and let me know if this change makes sense to you. I think it is an improvement as the output bytes now have the same encoding as the input bytes.


src/test/java/io/parsingdata/metal/EndiannessTest.java
```diff
@@ -23,27 +23,75 @@
...
+import io.parsingdata.metal.expression.value.Value;
+import io.parsingdata.metal.expression.value.ValueExpression;
 import io.parsingdata.metal.token.Token;
 
 public class EndiannessTest {
 
     @Test
     public void andAcrossByteBoundaryLE() throws IOException {
-        final Token t = def("x", con(2), eq(and(self, con(0x03, 0xff)), con(0x01, 0x1b)));
+        final Token t = def("x", con(2), eq(and(self, con(le(), 0xff, 0x03)), con(le(), 0x1b, 0x01)));
```
gertjanal: Please verify

PR #43, opened 2016-06-17
Keep encoding when using ConstantFactory

gertjanal:
In my opinion, encoding should not be changed when `and` is used


jvdb:
Clients should always mind the encoding! But thanks for the ideas!

PR #44, opened 2016-06-17
Html hexviewer for Metal Def tokens

gertjanal:
Re-opened.

Java code to generate a html hexviewer.

After generating the html, open the source file to load the hex editor
b45bace8-3317-11e6-86b0-e3106556b228

This code should probably be merged to metal-util


gertjanal:
Y U NO merge? :disappointed: :cry:

jvdb:
Looks awesome! I think this should be an application/library all by itself!

Issue #46, opened 2016-06-22
Clean up multimodule pom and Maven warnings


jvdb:
Some of the issues:

1. Separate descriptions for the different artifacts (looks better in the generated site).
2. Fix warnings around SCM properties and inheritance of the values it uses.
3. Fix Javadoc warnings that were introduced in the 3.1.0 release.

jvdb:
Apparently the links between the parent site and the submodule sites are intended to be broken(?):

https://maven.apache.org/plugins/maven-site-plugin/faq.html#Why_dont_the_links_between_parent_and_child_modules_work_when_I_run_mvn_site

PR #47, opened 2016-06-22
Cleanup46

jvdb:
Should resolve all cleanup issues discussed in #46. Please check that `mvn clean verify site` produces *zero* errors and warnings.


core/src/main/java/io/parsingdata/metal/data/selection/ByName.java
```diff
@@ -39,11 +40,12 @@ public static ParseValue get(final ParseGraph graph, final String name) {
     }
 
     /**
+     * @param graph The graph to search
```
rdvdijk: This needs actual JavaDoc.

PR #53, opened 2016-07-05
Remove recursion in ParseGraphList and ByOffset classes

gertjanal:


jvdb:
Thanks for the input! Currently we're not changing the recursive style, but we may add an alternative implementation at some point.

Issue #16, opened 2016-03-31
Support arbitrary precision integer values on all arithmetic and addressing operations


jvdb:
This may prove to be a challenge in some areas. E.g., Java requires integer indices on arrays.

PR #57, opened 2016-07-12
Implementation of multiexpressions

### LINKED ISSUE #20: Let ValueExpression.eval() return a collection of Values instead of a single Value
This will make it easier to generalize complex operations and predicates. An example is a Zip operation that operates on two collections.

Furthermore it will make the combinators more modular -- instead of having separate First, Last, etc. implementations, Ref can simply return all instances which can then be modified by generic First, Last, etc. modifiers.

jvdb:
This will resolve #20. There is a detailed discussion there as well.


core/src/main/java/io/parsingdata/metal/Shorthand.java
```diff
@@ -16,58 +16,18 @@
...
-import io.parsingdata.metal.expression.value.Reducer;
-import io.parsingdata.metal.expression.value.UnaryValueExpression;
-import io.parsingdata.metal.expression.value.Value;
-import io.parsingdata.metal.expression.value.ValueExpression;
-import io.parsingdata.metal.expression.value.arithmetic.Add;
-import io.parsingdata.metal.expression.value.arithmetic.Div;
-import io.parsingdata.metal.expression.value.arithmetic.Mod;
-import io.parsingdata.metal.expression.value.arithmetic.Mul;
-import io.parsingdata.metal.expression.value.arithmetic.Neg;
+import io.parsingdata.metal.expression.comparison.*;
```
rdvdijk: This probably changed because you are now using IDEA instead of Eclipse, correct?

(I'm fine with this, by the way.)

core/src/main/java/io/parsingdata/metal/expression/comparison/ComparisonExpression.java
```diff
@@ -37,11 +38,18 @@ public ComparisonExpression(final ValueExpression current, final ValueExpression
...
-        if (!opv.isPresent()) { return false; }
-        return compare(ocv.get(), opv.get());
+        final OptionalValueList ocl = _current == null ? OptionalValueList.create(OptionalValue.of(env.order.current())) : _current.eval(env, enc);
+        if (ocl.isEmpty()) { return false; }
+        final OptionalValueList opl = _predicate.eval(env, enc);
+        if (ocl.size != opl.size) { return false; }
+        return compare(ocl, opl);
+    }
+
+    private boolean compare(final OptionalValueList currents, final OptionalValueList predicates) {
```
rdvdijk: This private, recursive method has the same name the abstract method that should be implemented. That is confusing. Consider renaming this method `compareList`.

rdvdijk: `currents`? Why not simply `values`?

core/src/main/java/io/parsingdata/metal/Shorthand.java
jvdb: Yes. It also has a different import ordering. I suppose we can fix this but I'm fine with slowly migrating to IntelliJ default.

core/src/main/java/io/parsingdata/metal/expression/value/BinaryValueExpression.java
```diff
@@ -16,35 +16,48 @@
...
+import io.parsingdata.metal.data.OptionalValueList;
 import io.parsingdata.metal.encoding.Encoding;
 
+import static io.parsingdata.metal.Util.checkNotNull;
+
 public abstract class BinaryValueExpression implements ValueExpression {
 
-    private final ValueExpression _lop;
-    private final ValueExpression _rop;
+    private final ValueExpression lop;
```
rdvdijk: Agreed about the rename, however: make `public` while you're at it.

core/src/main/java/io/parsingdata/metal/expression/value/BinaryValueExpression.java
```diff
@@ -16,35 +16,48 @@
...
-        final OptionalValue lv = _lop.eval(env, enc);
-        if (!lv.isPresent()) { return lv; }
-        final OptionalValue rv = _rop.eval(env, enc);
-        if (!rv.isPresent()) { return rv; }
-        return eval(lv.get(), rv.get(), env, enc);
+    public OptionalValueList eval(final Environment env, final Encoding enc) {
+        return evalLists(lop.eval(env, enc), rop.eval(env, enc), env, enc);
+    }
+
+    private OptionalValueList evalLists(final OptionalValueList lvl, final OptionalValueList rvl, final Environment env, final Encoding enc) {
```
rdvdijk: I know you're a fan of these short variable names, but consider using simply `left` and `right` in code like this instead.

rdvdijk: (As a side note, `lvl` looks like `1v1` :smile:)

core/src/main/java/io/parsingdata/metal/expression/value/BinaryValueExpression.java
```diff
@@ -16,35 +16,48 @@
...
+        if (lvl.isEmpty()) { return makeListWithEmpty(rvl.size); }
+        if (rvl.isEmpty()) { return makeListWithEmpty(lvl.size); }
```
rdvdijk: These two lines of codes create behavior that should be documented *somewhere* in the Javadoc of this class (either on the class level Javadoc or on the public `eval` above).

core/src/main/java/io/parsingdata/metal/expression/value/Elvis.java
```diff
@@ -32,25 +35,27 @@
  * evaluates to that value, else it evaluates to the value of <code>ref("bar")</code>.
  */
 public class Elvis implements ValueExpression {
-    private final ValueExpression _lop;
-    private final ValueExpression _rop;
+    private final ValueExpression lop;
+    private final ValueExpression rop;
```
rdvdijk: (See my earlier comment, make these `public`.)

core/src/main/java/io/parsingdata/metal/expression/value/Elvis.java
```diff
@@ -32,25 +35,27 @@
...
-        final OptionalValue eval = _lop.eval(env, enc);
-        if (eval.isPresent()) {
-            return eval;
-        }
-        return _rop.eval(env, enc);
+    public OptionalValueList eval(final Environment env, final Encoding enc) {
+        return eval(lop.eval(env, enc), rop.eval(env, enc));
+    }
+
+    private OptionalValueList eval(final OptionalValueList llist, final OptionalValueList rlist) {
```
rdvdijk: Another style: `llist` and `rlist`. Sorry about the style nitpicking, the code would benefit from unifying these variable names throughout the codebase.

core/src/main/java/io/parsingdata/metal/expression/value/Elvis.java
```diff
@@ -32,25 +35,27 @@
...
+        if (llist.isEmpty()) { return rlist; }
+        if (rlist.isEmpty()) { return llist; }
```
rdvdijk: My earlier comment holds here as well, these two lines constitute explicit behavior that should be properly documented.

core/src/main/java/io/parsingdata/metal/expression/value/FoldLeft.java
```diff
@@ -20,33 +20,35 @@
...
     @Override
-    public OptionalValue eval(final Environment env, final Encoding enc) {
-        final OptionalValue init = _init != null ? _init.eval(env, enc) : OptionalValue.empty();
-        final ParseValueList values = env.order.getAll(_name).reverse();
-        if (values.isEmpty()) { return init; }
-        if (init.isPresent()) { return fold(env, enc, _reducer, init, values); }
-        return fold(env, enc, _reducer, OptionalValue.of(values.head), values.tail);
+    public OptionalValueList eval(final Environment env, final Encoding enc) {
+        final OptionalValueList init = _init != null ? _init.eval(env, enc) : OptionalValueList.EMPTY;
+        if (init.size > 1) { throw new RuntimeException("Init may not contain more than a single value."); }
```
rdvdijk: Use `IllegalArgumentException`, not just a plain `RunTimeException`.

core/src/main/java/io/parsingdata/metal/expression/value/FoldLeft.java
```diff
@@ -20,33 +20,35 @@
...
+        if (values.isEmpty() || values.containsEmpty()) { return init; }
+        if (!init.isEmpty()) { return OptionalValueList.create(fold(env, enc, _reducer, init.head, values)); }
+        return OptionalValueList.create(fold(env, enc, _reducer, values.head, values.tail));
     }
 
-    private OptionalValue fold(final Environment env, final Encoding enc, final Reducer reducer, final OptionalValue head, final ParseValueList tail) {
+    private OptionalValue fold(final Environment env, final Encoding enc, final Reducer reducer, final OptionalValue head, final OptionalValueList tail) {
         if (!head.isPresent() || tail.isEmpty()) { return head; }
-        return fold(env, enc, reducer, reducer.reduce(con(head.get()), con(tail.head)).eval(env, enc), tail.tail);
+        return fold(env, enc, reducer, reducer.reduce(con(head.get()), con(tail.head.get())).eval(env, enc).head, tail.tail);
```
rdvdijk: Regarding second argument to `reducer.reduce`: `con(tail.head.get())).eval(env, enc).head`:

So in theory the `ValueExpression` returned by the `Reducer` could result in multiple values. Here only the `head` of that list is passed along to the next `fold`.

- Passing on `OptionalValueList` into `fold` makes no sense, right? :wink:
- Maybe we should add a sanity check here that the `Reducer` never returns an expression that results in multiple values?
- Should this behavior be documented somehow?

core/src/main/java/io/parsingdata/metal/expression/value/FoldRight.java
```diff
@@ -16,37 +16,38 @@
...
     @Override
-    public OptionalValue eval(final Environment env, final Encoding enc) {
-        final OptionalValue init = _init != null ? _init.eval(env, enc) : OptionalValue.empty();
-        final ParseValueList values = env.order.getAll(_name);
-        if (values.isEmpty()) { return init; }
-        if (init.isPresent()) { return fold(env, enc, _reducer, init, values); }
-        return fold(env, enc, _reducer, OptionalValue.of(values.head), values.tail);
+    public OptionalValueList eval(final Environment env, final Encoding enc) {
+        final OptionalValueList init = _init != null ? _init.eval(env, enc) : OptionalValueList.EMPTY;
+        if (init.size > 1) { throw new RuntimeException("Init may not contain more than a single value."); }
```
rdvdijk: `RunTimeException` -> `IllegalArgumentException`

core/src/main/java/io/parsingdata/metal/expression/value/FoldRight.java
```diff
@@ -16,37 +16,38 @@
...
+        if (values.isEmpty() || values.containsEmpty()) { return init; }
+        if (!init.isEmpty()) { return OptionalValueList.create(fold(env, enc, _reducer, init.head, values)); }
+        return OptionalValueList.create(fold(env, enc, _reducer, values.head, values.tail));
     }
 
-    private OptionalValue fold(final Environment env, final Encoding enc, final Reducer reducer, final OptionalValue head, final ParseValueList tail) {
+    private OptionalValue fold(final Environment env, final Encoding enc, final Reducer reducer, final OptionalValue head, final OptionalValueList tail) {
         if (!head.isPresent() || tail.isEmpty()) { return head; }
-        return fold(env, enc, reducer, reducer.reduce(con(tail.head), con(head.get())).eval(env, enc), tail.tail);
+        return fold(env, enc, reducer, reducer.reduce(con(tail.head.get()), con(head.get())).eval(env, enc).head, tail.tail);
```
rdvdijk: (See my earlier comment about only passing on the `head` into the next `reduce`.)

core/src/main/java/io/parsingdata/metal/expression/value/reference/First.java
```diff
@@ -16,35 +16,30 @@
...
 import io.parsingdata.metal.encoding.Encoding;
-import io.parsingdata.metal.expression.value.OptionalValue;
 import io.parsingdata.metal.expression.value.ValueExpression;
 
+import static io.parsingdata.metal.Util.checkNotNull;
+
 public class First implements ValueExpression {
 
-    private final String _name;
+    private final ValueExpression _op;
```
rdvdijk: (In other new/changed code you had already used field names without underscore, but not here?)

core/src/main/java/io/parsingdata/metal/expression/value/FoldLeft.java
jvdb: Good point, I will look into this. Looks like a quick rework to handle lists, but perhaps it should be more constrained.

core/src/main/java/io/parsingdata/metal/expression/value/reference/Offset.java
```diff
@@ -16,32 +16,21 @@
...
-        final ParseValue ref = env.order.get(_name);
-        return ref != null ? OptionalValue.of(ConstantFactory.createFromNumeric(ref.getOffset(), ref.getEncoding())) : OptionalValue.empty();
-    }
+    public Offset(final ValueExpression op) { super(op); }
 
     @Override
-    public String toString() {
-        return getClass().getSimpleName() + "(" + _name + ")";
+    public OptionalValue eval(final Value op, final Environment env, final Encoding enc) {
+        if (op instanceof ParseValue) {
```
rdvdijk: Bummer about the `instanceof` and cast code smell, but I see why.

(Maybe discuss this off-line.)

core/src/main/java/io/parsingdata/metal/expression/value/reference/Ref.java
```diff
@@ -32,8 +33,8 @@ public Ref(final String name) {
     }
 
     @Override
-    public OptionalValue eval(final Environment env, final Encoding enc) {
-        return OptionalValue.of(env.order.get(_name));
+    public OptionalValueList eval(final Environment env, final Encoding enc) {
+        return OptionalValueList.create(env.order.getAll(_name));
```
rdvdijk: This diff a prime example of why we wanted all this :+1:

core/src/main/java/io/parsingdata/metal/token/Def.java
```diff
@@ -44,20 +44,21 @@ public Def(final String name, final ValueExpression size, final Expression pred,
 
     @Override
     protected ParseResult parseImpl(final String scope, final Environment env, final Encoding enc) throws IOException {
-        final OptionalValue size = _size.eval(env, enc);
-        if (!size.isPresent()) {
+        final OptionalValueList size = _size.eval(env, enc);
+        if (size.size != 1) { throw new RuntimeException("Size must yield a single value."); }
```
rdvdijk: `RuntimeException` :arrow_right: `IllegalArgumentException`

core/src/main/java/io/parsingdata/metal/token/Def.java
```diff
@@ -44,20 +44,21 @@ public Def(final String name, final ValueExpression size, final Expression pred,
...
+        if (!size.head.isPresent()) {
             return new ParseResult(false, env);
         }
-        // TODO: Handle value expression results as BigInteger (METAL-15)
-        final int dataSize = size.get().asNumeric().intValue();
+        // TODO: Handle value expression results as BigInteger
```
rdvdijk: Do we have a GitHub issue for this `TODO`?

core/src/main/java/io/parsingdata/metal/token/Def.java
```diff
@@ -44,20 +44,21 @@ public Def(final String name, final ValueExpression size, final Expression pred,
...
+        final int dataSize = size.head.get().asNumeric().intValue();
         if (dataSize < 0) {
             return new ParseResult(false, env);
         }
         final byte[] data = new byte[dataSize];
         if (env.input.read(env.offset, data) != data.length) {
             return new ParseResult(false, env);
         }
-        final Environment newEnv = new Environment(env.order.add(new ParseValue(scope, _name, this, env.offset, data, enc)), env.input, env.offset + size.get().asNumeric().intValue());
+        final Environment newEnv = new Environment(env.order.add(new ParseValue(scope, _name, this, env.offset, data, enc)), env.input, env.offset + size.head.get().asNumeric().intValue());
```
rdvdijk: Use `dataSize` here instead of calling `size.head.get().asNumeric().intValue()` again.

core/src/main/java/io/parsingdata/metal/token/Nod.java
```diff
@@ -37,8 +37,10 @@ public Nod(final ValueExpression size, final Encoding enc) {
 
     @Override
     protected ParseResult parseImpl(final String scope, final Environment env, final Encoding enc) throws IOException {
-        final OptionalValue ov = _size.eval(env, enc);
-        return ov.isPresent() ? new ParseResult(true, new Environment(env.order, env.input, env.offset + ov.get().asNumeric().longValue())) : new ParseResult(false, env);
+        final OptionalValueList ov = _size.eval(env, enc);
+        if (ov.isEmpty()) { return new ParseResult(false, env); }
+        if (ov.size > 1) { throw new RuntimeException("A single size value must be provided."); }
```
rdvdijk: `RuntimeException` :arrow_right: `IllegalArgumentException`

core/src/main/java/io/parsingdata/metal/token/Sub.java
```diff
@@ -40,17 +40,38 @@ public Sub(final Token op, final ValueExpression addr, final Encoding enc) {
 
     @Override
     protected ParseResult parseImpl(final String scope, final Environment env, final Encoding enc) throws IOException {
-        final OptionalValue ov = _addr.eval(env, enc);
-        if (!ov.isPresent()) { return new ParseResult(false, env); }
-        final long ref = ov.get().asNumeric().longValue();
-        if (env.order.hasGraphAtRef(ref)) { return new ParseResult(true, new Environment(env.order.add(new ParseRef(ref, this)), env.input, env.offset)); }
-        final ParseResult res = _op.parse(scope, new Environment(env.order.addBranch(this), env.input, ref), enc);
+        final OptionalValueList addrs = _addr.eval(env, enc);
```
rdvdijk: `addrs` :arrow_right: `addresses` (or `offsets`).

core/src/main/java/io/parsingdata/metal/token/Sub.java
```diff
@@ -40,17 +40,38 @@ public Sub(final Token op, final ValueExpression addr, final Encoding enc) {
...
+        if (addrs.isEmpty() || addrs.containsEmpty()) { return new ParseResult(false, env); }
+        final ParseResult res = iterate(scope, addrs, new Environment(env.order.addBranch(this), env.input, env.offset), enc);
```
rdvdijk: Ah, so here is the single `addBranch` (and not deeper into the iteration below).

core/src/main/java/io/parsingdata/metal/token/Sub.java
```diff
@@ -40,17 +40,38 @@ public Sub(final Token op, final ValueExpression addr, final Encoding enc) {
...
+    }
+
+    private ParseResult parse(final String scope, final long ref, final Environment env, final Encoding enc) throws IOException {
+        if (env.order.hasGraphAtRef(ref)) {
+            return new ParseResult(true, new Environment(env.order.add(new ParseRef(ref, this)), env.input, env.offset));
+        }
+        final ParseResult res = _op.parse(scope, new Environment(env.order, env.input, ref), enc);
+        if (res.succeeded()) {
+            return res;
+        }
```
rdvdijk: Three lines of code for a single return, I'm so proud. :relieved:

core/src/test/java/io/parsingdata/metal/ComparisonExpressionSemanticsTest.java
```diff
@@ -61,7 +62,11 @@
             { "1 == 1(eq)", valCom(1, eq(ref("a"))), stream(1, 1), enc(), true },
             { "2 == 1(eq)", valCom(1, eq(ref("a"))), stream(1, 2), enc(), false },
             { "1 == 1 with self", valCom(1, eq(self, ref("a"))), stream(1, 1), enc(), true },
-            { "1 == 2 with self", valCom(1, eq(self, ref("a"))), stream(1, 2), enc(), false }
+            { "1 == 2 with self", valCom(1, eq(self, ref("a"))), stream(1, 2), enc(), false },
+            { "1, 2 == 1", listCom(eq(ref("a"), ref("b")), expTrue()), stream(1, 2, 1, 2), enc(), false },
```
rdvdijk: Maybe put the expectation in the test message? e.g. `1, 2 != 1`

core/src/test/java/io/parsingdata/metal/ErrorsTest.java
```diff
@@ -16,35 +16,36 @@
...
+import org.junit.rules.ExpectedException;
+
+import java.io.IOException;
+
+import static io.parsingdata.metal.Shorthand.*;
+import static io.parsingdata.metal.TokenDefinitions.any;
+import static io.parsingdata.metal.util.EncodingFactory.enc;
+import static io.parsingdata.metal.util.EnvironmentFactory.stream;
 
-@RunWith(JUnit4.class)
```
rdvdijk: (Kudos :smile:)

core/src/test/java/io/parsingdata/metal/ErrorsTest.java
```diff
@@ -54,4 +55,17 @@ public void ioError() throws IOException {
         t.parse(env, enc());
     }
 
+    @Test
+    public void multiValueInRepN() throws IOException {
+        thrown.expect(RuntimeException.class);
```
rdvdijk: OK, one more time. `IllegalArgumentException`, please :wink:

core/src/test/java/io/parsingdata/metal/ReducersTest.java
```diff
@@ -16,42 +16,31 @@
...
+import static io.parsingdata.metal.util.EncodingFactory.le;
+import static io.parsingdata.metal.util.EnvironmentFactory.stream;
 
 public class ReducersTest extends ParameterizedParse {
 
     @Parameters(name="{0} ({4})")
     public static Collection<Object[]> data() {
         return Arrays.asList(new Object[][] {
             { "[1, 2, 3, 6] a, a, a, addAll(a)", reduceAddA, stream(1, 2, 3, 6), enc(), true },
+            { "[1, 2, 3, 3] a, a, a, addAllOffset(a)", reduceAddOffsetA, stream(1, 2, 3, 3), enc(), true },
```
rdvdijk: 0+1+2 = 3. Got it. (Took me while :smirk:)

core/src/test/java/io/parsingdata/metal/ReferenceValueExpressionSemanticsTest.java
```diff
@@ -56,14 +49,21 @@
...
-            { "[1, 2, 3] a, a, first(b)", refList("a", "a", first("b")), stream(1, 2, 3), enc(), false },
+            { "[1, 2, 1] a, a, first(a)", refList("a", "a", first(ref("a"))), stream(1, 2, 1), enc(), true },
+            { "[1, 2, 3] a, a, first(a)", refList("a", "a", first(ref("a"))), stream(1, 2, 3), enc(), false },
+            { "[1, 2, 3] a, a, first(b)", refList("a", "a", first(ref("b"))), stream(1, 2, 3), enc(), false },
             { "[1, 2, 3] a, a, ref(b)", refList("a", "a", ref("b")), stream(1, 2, 3), enc(), false },
-            { "[1, 2, 0] a, b, offset(a)", refList("a", "b", offset("a")), stream(1, 2, 0), enc(), true },
-            { "[1, 2, 1] a, a, offset(a)", refList("a", "a", offset("a")), stream(1, 2, 1), enc(), true },
-            { "[1, 2, 2] a, b, offset(z)", refList("a", "b", offset("z")), stream(1, 2, 2), enc(), true },
-            { "[1, 2, 3] a, b, offset(c)", refList("a", "b", offset("c")), stream(1, 2, 3), enc(), false }
+            { "[1, 2, 3] a, a, last(a)", refList("a", "a", last(ref("a"))), stream(1, 2, 2), enc(), true },
```
rdvdijk: Test description does not match the stream data `[1, 2, 3]` vs `stream(1, 2, 2)`.

core/src/test/java/io/parsingdata/metal/SubStructTableTest.java
```diff
@@ -0,0 +1,90 @@
...
+                             * count:  ^
+                             * pointers:  ^, ^, ^
+                             * ref1:      +----------------^^--^^
+                             * ref2:         +-----^^--^^
+                             * ref3:            +---------------------^^--^^
+                             */
+        final ParseResult res = table.parse(env, enc());
+        assertTrue(res.succeeded());
+        assertEquals(4, res.getEnvironment().offset);
+        final ParseGraph order = res.getEnvironment().order;
```
rdvdijk: I think this test would benefit from explaining/testing what the intermediate graphs are.

What is `order.head` and what is `order.head.asGraph().head`? Maybe test those graphs' definitions (their classes, at least).

core/src/test/java/io/parsingdata/metal/SubStructTest.java
```diff
@@ -98,7 +87,7 @@ public void linkedListWithSelfReference() throws IOException {
         final ParseGraph first = out.head.asGraph();
         checkBranch(first, 0, 0);
 
-        final ParseRef ref = first.tail.head.asGraph().head.asRef();
+        final ParseRef ref = first.tail.head.asGraph().head.asGraph().head.asRef();
```
rdvdijk: Why is there an extra graph here?

core/src/main/java/io/parsingdata/metal/token/Def.java
jvdb: Yes: #16. I'll add a reference to it.

core/src/test/java/io/parsingdata/metal/expression/value/BinaryValueExpressionListSemanticsTest.java
```diff
@@ -0,0 +1,79 @@
...
+import static io.parsingdata.metal.util.EnvironmentFactory.stream;
+
+public class BinaryValueExpressionListSemanticsTest extends ParameterizedParse {
+
+    @Parameterized.Parameters(name="{0} ({4})")
+    public static Collection<Object[]> data() {
+        return Arrays.asList(new Object[][] {
+            { "a, a, b, b, last(a+b)", pred2(eq(last(add(ref("a"), ref("b"))))), stream(1, 2, 3, 4, 6), enc(), true },
+            { "a, a, b, b, first(a+b)", pred2(eq(first(add(ref("a"), ref("b"))))), stream(1, 2, 3, 4, 4), enc(), true },
+            { "a, a, a, b, b, last(a+b)", pred3(eq(last(add(ref("a"), ref("b"))))), stream(1, 2, 3, 4, 5, 8), enc(), true },
```
rdvdijk: `last(add(ref("a"), ref("b")))`, that should not be 8, in this case, right?

- `ref("a")` :arrow_right: `[1,2,3]`
- `ref("b")` :arrow_right: `[4,5]`
- `add(ref("a"), ref("b"))` :arrow_right: `[1+4,2+5,3+ω]` =  `[5,7,ω]`
- `last(add(ref("a"), ref("b")))` :arrow_right: ω ?

Or is it 'left' padded with `ω`, because the `head` of the lists point to `3` and `5`?

- `add(ref("a"), ref("b"))` :arrow_right: `[1+ω,2+4,3+5]` =  `[ω, 6,8]`

core/src/test/java/io/parsingdata/metal/expression/value/BinaryValueExpressionListSemanticsTest.java
```diff
@@ -0,0 +1,79 @@
...
+            { "a, a, a, b, b, first(a+b)", pred3(eq(first(add(ref("a"), ref("b"))))), stream(1, 2, 3, 4, 5, 5), enc(), false },
```
rdvdijk: So this parse fails because of the the first in this list is empty (`ω`)?

core/src/test/java/io/parsingdata/metal/expression/value/BinaryValueExpressionListSemanticsTest.java
```diff
@@ -0,0 +1,79 @@
...
+            { "a, a, b, b, first(offset(a)+offset(b))", pred2(eq(first(add(offset(ref("a")), offset(ref("b")))))), stream(1, 2, 3, 4, 2), enc(), true },
+            { "a, a, a, b, b, last((offset(a)+offset(b))", pred3(eq(last(add(offset(ref("a")), offset(ref("b")))))), stream(1, 2, 3, 4, 5, 6), enc(), true },
+            { "a, a, a, b, b, first(offset(a)+offset(b))", pred3(eq(first(add(offset(ref("a")), offset(ref("b")))))), stream(1, 2, 3, 4, 5, 3), enc(), false },
+            { "a, a, a, b, b, last(a)+first(b)", pred3(eq(add(last(ref("a")), first(ref("b"))))), stream(1, 2, 3, 4, 5, 7), enc(), true },
+            { "a, a, a, b, b, first(a)+last(b)", pred3(eq(add(first(ref("a")), last(ref("b"))))), stream(1, 2, 3, 4, 5, 6), enc(), true },
+            { "a, a, b, b, last(not(a)+not(b))", pred2(eq(last(add(not(ref("a")), not(ref("b")))))), stream(1, 240, 3, 15, -1), enc(), true },
+            { "a, a, b, b, first(not(a)+not(b))", pred2(eq(first(add(not(ref("a")), not(ref("b")))))), stream(230, 2, 25, 4, -1), enc(), true },
+            { "a, a, a, b, b, last((not(a)+not(b))", pred3(eq(last(add(not(ref("a")), not(ref("b")))))), stream(1, 2, 200, 4, 55, -1), enc(), true },
```
rdvdijk: Unmatched parentheses in description: `last((not(a)+not(b))`.

core/src/test/java/io/parsingdata/metal/SubStructTest.java
jvdb: I think this was added because of Sub-nodes in the graph.

core/src/test/java/io/parsingdata/metal/expression/value/BinaryValueExpressionListSemanticsTest.java
jvdb: It's right-padded, but the immutable lists are stored reversed :)

core/src/test/java/io/parsingdata/metal/expression/value/BinaryValueExpressionListSemanticsTest.java
jvdb: Yes.

core/src/test/java/io/parsingdata/metal/expression/value/BinaryValueExpressionListSemanticsTest.java
```diff
@@ -0,0 +1,79 @@
...
+            { "a, a, a, b, b, first(a+b)", pred3(eq(first(add(ref("a"), ref("b"))))), stream(1, 2, 3, 4, 5, 5), enc(), false },
+            { "a, a, b, b, last(offset(a)+offset(b))", pred2(eq(last(add(offset(ref("a")), offset(ref("b")))))), stream(1, 2, 3, 4, 4), enc(), true },
+            { "a, a, b, b, first(offset(a)+offset(b))", pred2(eq(first(add(offset(ref("a")), offset(ref("b")))))), stream(1, 2, 3, 4, 2), enc(), true },
+            { "a, a, a, b, b, last((offset(a)+offset(b))", pred3(eq(last(add(offset(ref("a")), offset(ref("b")))))), stream(1, 2, 3, 4, 5, 6), enc(), true },
+            { "a, a, a, b, b, first(offset(a)+offset(b))", pred3(eq(first(add(offset(ref("a")), offset(ref("b")))))), stream(1, 2, 3, 4, 5, 3), enc(), false },
+            { "a, a, a, b, b, last(a)+first(b)", pred3(eq(add(last(ref("a")), first(ref("b"))))), stream(1, 2, 3, 4, 5, 7), enc(), true },
+            { "a, a, a, b, b, first(a)+last(b)", pred3(eq(add(first(ref("a")), last(ref("b"))))), stream(1, 2, 3, 4, 5, 6), enc(), true },
+            { "a, a, b, b, last(not(a)+not(b))", pred2(eq(last(add(not(ref("a")), not(ref("b")))))), stream(1, 240, 3, 15, -1), enc(), true },
```
rdvdijk: How can the result be `-1` if the encoding is unsigned?

`~240 + ~15 = 15 + 240 = 255`, which is not `-1` (unless it is signed)

core/src/test/java/io/parsingdata/metal/expression/value/ElvisExpressionTest.java
```diff
@@ -50,27 +44,78 @@
...
         final ParseResult result = choice.parse(stream(3), enc());
-        final OptionalValue eval = elvisExpression.eval(result.getEnvironment(), enc());
+        final OptionalValueList eval = elvisExpression.eval(result.getEnvironment(), enc());
 
-        assertFalse(eval.isPresent());
+        assertNotNull(eval);
+        assertTrue(eval.isEmpty());
+    }
+
+    private Token createListToken(final Token left, final Token right) {
```
rdvdijk: Looks like a using `seq` directly would be much simpler?

core/src/test/java/io/parsingdata/metal/expression/value/ElvisExpressionTest.java
```diff
@@ -50,27 +44,78 @@
...
+        assertEquals(4, eval.head.get().asNumeric().intValue());
+        assertEquals(3, eval.tail.head.get().asNumeric().intValue());
+    }
+
+    @Test
+    public void elvisListDifferentLengths() throws IOException {
+        final ParseResult result = createListToken(seq(any("a"), any("a")), seq(any("b"), any("b"), any("b"))).parse(stream(1, 2, 3, 4, 5), enc());
+        assertTrue(result.succeeded());
+        final ValueExpression elvis = elvis(ref("a"), ref("b"));
+        final OptionalValueList eval = elvis.eval(result.getEnvironment(), enc());
```
rdvdijk: Let me work this out:

- `ref("a") = [2,1]`
- `ref("b") = [5,4,3]`
- `elvis(ref("a"), ref("b")) = [2,1,3]`

Got it. :-)

core/src/main/java/io/parsingdata/metal/expression/value/BinaryValueExpression.java
jvdb: See #58, it's also part of milestone 4.0.0.

core/src/main/java/io/parsingdata/metal/expression/comparison/ComparisonExpression.java
jvdb: See #59, it's also part of milestone 4.0.0.

core/src/test/java/io/parsingdata/metal/expression/value/BinaryValueExpressionListSemanticsTest.java
jvdb: I'm always confused about the byte notation for 255, so I tend to write -1 instead.

core/src/test/java/io/parsingdata/metal/SubStructTest.java
rdvdijk: Let me attempt to visualize this in ASCII:

```
---  Graph(Sub) --- Graph(Sub) -- Graph(Sub)
         |              |             |
     Graph(Seq)        Ref        Graph(Seq)
```

Right?

core/src/test/java/io/parsingdata/metal/token/NodTest.java
```diff
@@ -29,6 +29,7 @@
 import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
 
 public class NodTest {
```
rdvdijk: Unrelated: this needs a test for negative size(s).

jvdb: That's actually a serious bug in the code itself (not just the test). See #61.

core/src/main/java/io/parsingdata/metal/expression/value/reference/Offset.java
jvdb: See #63. May not be for 4.0.0 though.

PR #66, opened 2016-07-14
#55: Introduce enum for signedness to replace boolean argument.

### LINKED ISSUE #55: Use an enum for Encoding signedness instead of boolean
Currently we use a `boolean` to set the signedness of an Encoding:

```java
    public Encoding(final boolean signed) {
        this(signed, DEFAULT_CHARSET, DEFAULT_BYTE_ORDER);
    }
    // ...
    public Encoding(final boolean signed, final Charset charset, final ByteOrder byteOrder) {
        _signed = signed;
        _charset = charset;
        _byteOrder = byteOrder;
    }
```

Using an `enum` similar to `ByteOrder` would be much cleaner, and easier on the eyes.

Proposal: `Signedness.SIGNED` and `Signedness.UNSIGNED`.

rdvdijk:
This will resolve #55.

### DIFF
--- a/core/src/main/java/io/parsingdata/metal/encoding/Encoding.java
+++ b/core/src/main/java/io/parsingdata/metal/encoding/Encoding.java
@@ -20,11 +20,11 @@ import java.nio.charset.Charset;
 
 public class Encoding {
 
-    private static final boolean DEFAULT_SIGNED = false;
+    private static final Sign DEFAULT_SIGNED = Sign.UNSIGNED;
     private static final Charset DEFAULT_CHARSET = Charset.forName("ISO646-US");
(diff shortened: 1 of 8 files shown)


core/src/main/java/io/parsingdata/metal/encoding/Encoding.java
```diff
@@ -64,7 +68,6 @@ public ByteOrder getByteOrder() {
 
     @Override
     public String toString() {
-        return getClass().getSimpleName() + "(" + (_signed ? "SIGNED" : "UNSIGNED") + "," + _charset + "," + _byteOrder + ")";
+        return getClass().getSimpleName() + "(" + _sign + "," + _charset + "," + _byteOrder + ")";
```
jvdb: This is in itself almost reason enough to make this change :)

PR #68, opened 2016-07-15
Nod size can not be zero or negative.

### LINKED ISSUE #61: Nod doesn't check for < 0 sizes

rdvdijk:
This resolves #61.

### DIFF
--- a/core/src/main/java/io/parsingdata/metal/token/Nod.java
+++ b/core/src/main/java/io/parsingdata/metal/token/Nod.java
@@ -40,7 +40,9 @@ public class Nod extends Token {
         final OptionalValueList ov = _size.eval(env, enc);
         if (ov.isEmpty()) { return new ParseResult(false, env); }
         if (ov.size > 1) { throw new IllegalStateException("Size may not evaluate to more than a single value."); }
-        return ov.head.isPresent() ? new ParseResult(true, new Environment(env.order, env.input, env.offset + ov.head.get().asNumeric().longValue())) : new ParseResult(false, env);
+        final long size = ov.head.get().asNumeric().longValue();
+        if (size <= 0) { throw new IllegalStateException("Size must be larger than 0, but is: " + size); }
+        return ov.head.isPresent() ? new ParseResult(true, new Environment(env.order, env.input, env.offset + size)) : new ParseResult(false, env);
     }
 
     @Override
--- a/core/src/main/java/io/parsingdata/metal/token/Nod.java
+++ b/core/src/main/java/io/parsingdata/metal/token/Nod.java
@@ -38,11 +38,20 @@ public class Nod extends Token {
     @Override
     protected ParseResult parseImpl(final String scope, final Environment env, final Encoding enc) throws IOException {
         final OptionalValueList ov = _size.eval(env, enc);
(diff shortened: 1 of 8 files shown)


core/src/main/java/io/parsingdata/metal/token/Nod.java
```diff
@@ -40,7 +40,9 @@ protected ParseResult parseImpl(final String scope, final Environment env, final
         final OptionalValueList ov = _size.eval(env, enc);
         if (ov.isEmpty()) { return new ParseResult(false, env); }
         if (ov.size > 1) { throw new IllegalStateException("Size may not evaluate to more than a single value."); }
-        return ov.head.isPresent() ? new ParseResult(true, new Environment(env.order, env.input, env.offset + ov.head.get().asNumeric().longValue())) : new ParseResult(false, env);
+        final long size = ov.head.get().asNumeric().longValue();
+        if (size <= 0) { throw new IllegalStateException("Size must be larger than 0, but is: " + size); }
+        return ov.head.isPresent() ? new ParseResult(true, new Environment(env.order, env.input, env.offset + size)) : new ParseResult(false, env);
```
jvdb: Two things:

First, the way it handles unexpected values is very inconsistent:
- `[ ]` -> `false`
- `[ x, y ]` -> `IllegalStateException`
- `[ ω ]` -> `false`
- `[ -1 ]` -> `IllegalStateException`

Second, there is a bug in the handling of `[ ω ]` as the check for it is on line 45 while `ov.head.get()` is executed without check on line 43.

rdvdijk: Good point. I now see that `Def` results in `false` for negative sizes. I think we should have the same behavior here.

rdvdijk: @jvdb Can you check this new implementation?

core/src/main/java/io/parsingdata/metal/token/Nod.java
```diff
@@ -38,9 +38,20 @@ public Nod(final ValueExpression size, final Encoding enc) {
     @Override
     protected ParseResult parseImpl(final String scope, final Environment env, final Encoding enc) throws IOException {
         final OptionalValueList ov = _size.eval(env, enc);
-        if (ov.isEmpty()) { return new ParseResult(false, env); }
-        if (ov.size > 1) { throw new IllegalStateException("Size may not evaluate to more than a single value."); }
-        return ov.head.isPresent() ? new ParseResult(true, new Environment(env.order, env.input, env.offset + ov.head.get().asNumeric().longValue())) : new ParseResult(false, env);
+        if (ov.isEmpty()) {
```
rdvdijk: @jvdb Is this `ov.isEmpty()` check actually needed now, by the way? I see that in `Def` the check is slightly different:

```java
        final OptionalValueList size = _size.eval(env, enc);
        if (size.size != 1) { throw new IllegalStateException("Size may not evaluate to more than a single value."); }
        if (!size.head.isPresent()) {
            return new ParseResult(false, env);
        }
        // TODO: Handle value expression results as BigInteger (#16)
        final int dataSize = size.head.get().asNumeric().intValue();
        if (dataSize < 0) {
            return new ParseResult(false, env);
        }
```

PR #75, opened 2016-07-19
Make immutable fields public, improve naming, remove getters from ParseResult

### LINKED ISSUE #58: Make private immutable fields public

jvdb:
Resolves #58: Many immutable public fields in `Token` and `Expression` implementations are now public. The argument names in `Shorthand` and public constructors have been improved. Public field names also have better names. `ParseResult` no longer has getters (so `result.succeeded()` and `result.getEnvironment()` become `result.succeeded` and `result.environment` respectively).

### DIFF
--- a/core/src/main/java/io/parsingdata/metal/expression/comparison/ComparisonExpression.java
+++ b/core/src/main/java/io/parsingdata/metal/expression/comparison/ComparisonExpression.java
@@ -38,11 +38,11 @@ public abstract class ComparisonExpression implements Expression {
 
     @Override
     public boolean eval(final Environment env, final Encoding enc) {
-        final OptionalValueList ocl = value == null ? OptionalValueList.create(OptionalValue.of(env.order.current())) : value.eval(env, enc);
-        if (ocl.isEmpty()) { return false; }
+        final OptionalValueList ovl = value == null ? OptionalValueList.create(OptionalValue.of(env.order.current())) : value.eval(env, enc);
+        if (ovl.isEmpty()) { return false; }
         final OptionalValueList opl = predicate.eval(env, enc);
-        if (ocl.size != opl.size) { return false; }
-        return compare(ocl, opl);
+        if (ovl.size != opl.size) { return false; }
+        return compare(ovl, opl);
     }
 
     private boolean compare(final OptionalValueList currents, final OptionalValueList predicates) {
--- a/core/src/main/java/io/parsingdata/metal/expression/comparison/ComparisonExpression.java
+++ b/core/src/main/java/io/parsingdata/metal/expression/comparison/ComparisonExpression.java
@@ -16,8 +16,6 @@
 
 package io.parsingdata.metal.expression.comparison;
 
-import static io.parsingdata.metal.Util.checkNotNull;
-
 import io.parsingdata.metal.data.Environment;
 import io.parsingdata.metal.data.OptionalValueList;
 import io.parsingdata.metal.encoding.Encoding;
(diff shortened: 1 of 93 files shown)


core/src/main/java/io/parsingdata/metal/expression/comparison/ComparisonExpression.java
```diff
@@ -26,21 +24,23 @@
...
-        _predicate = checkNotNull(predicate, "predicate");
+    public ComparisonExpression(final ValueExpression value, final ValueExpression predicate) {
+        this.value = value;
+        this.predicate = checkNotNull(predicate, "predicate");
     }
 
     @Override
     public boolean eval(final Environment env, final Encoding enc) {
-        final OptionalValueList ocl = _current == null ? OptionalValueList.create(OptionalValue.of(env.order.current())) : _current.eval(env, enc);
+        final OptionalValueList ocl = value == null ? OptionalValueList.create(OptionalValue.of(env.order.current())) : value.eval(env, enc);
```
rdvdijk: `ocl` does really make sense anymore, maybe `ovl`?

core/src/main/java/io/parsingdata/metal/expression/value/reference/Count.java
```diff
@@ -29,15 +29,15 @@
...
-    public Count(final ValueExpression op) {
-        this.op = checkNotNull(op, "op");
+    public Count(final ValueExpression operand) {
+        this.operand = checkNotNull(operand, "operand");
     }
 
     @Override
     public OptionalValueList eval(final Environment env, final Encoding enc) {
-        final OptionalValueList ovl = this.op.eval(env, enc);
+        final OptionalValueList ovl = this.operand.eval(env, enc);
```
rdvdijk: In some cases I see unnecessary `this.` prefixes. In this case it was already there, and it can be safely removed.

To compare, `toString()` below doesn't use the prefix.

core/src/main/java/io/parsingdata/metal/token/Cho.java
```diff
@@ -27,31 +27,31 @@
 
 public class Cho extends Token {
 
-    private final Token[] _tokens;
+    public final Token[] tokens;
```
rdvdijk: An array is mutable. We should expose a clone of the array, or a collection:

```java
public List<Token> tokens() {
    return Arrays.asList(tokens);
}
```

(Same holds for `Seq.)

core/src/main/java/io/parsingdata/metal/token/Seq.java
```diff
@@ -27,31 +27,31 @@
 
 public class Seq extends Token {
 
-    private final Token[] _tokens;
+    public final Token[] tokens;
```
rdvdijk: (See my comment in `Cho`.)

core/src/main/java/io/parsingdata/metal/token/Token.java
```diff
@@ -16,24 +16,24 @@
...
 import io.parsingdata.metal.encoding.Encoding;
 
+import java.io.IOException;
+
 public abstract class Token {
 
     public final static String DEFAULT_NAME = "W";
 
-    private final Encoding _enc;
+    public final Encoding enc;
```
rdvdijk: Create a `enc` -> `encoding` and `env` -> `environment` issue for the 5.0.0 milestone.

core/src/main/java/io/parsingdata/metal/expression/value/reference/Count.java
jvdb: I found some more in `OptionalValueList` and `ParseGraph`, fixed those too.

core/src/main/java/io/parsingdata/metal/token/Cho.java
jvdb: Wow, great catch! I've just turned it into a `return tokens.clone()`. We have to rethink all those data structures at some point anyway, so I would propose to keep it simple for now.

PR #81, opened 2016-08-02
#80: ParseValues from Def with explicit encoding do not have a name

### LINKED ISSUE #80: ParseValues from Def with explicit encoding do not have a name

gertjanal:
Resolves #80


rdvdijk:
A bug that was found when @gertjanal was updating Traces to Metal 4.0.0.

PR #84, opened 2016-09-08
Use junit-quickcheck to test ConstantFactory's long and BigInteger conversion code.

### LINKED ISSUE #83: Improve value conversion code tests
As well as possibly other parts that are suitable to this approach.

jvdb:
This will resolve #83.

I hesitated due to loss of JDK7 testing on Travis. But we'll lose that quickly anyway as we move toward Java8 with Metal itself. So seems a reasonable step.

### DIFF
--- a/core/src/test/java/io/parsingdata/metal/expression/value/ConstantFactoryTest.java
+++ b/core/src/test/java/io/parsingdata/metal/expression/value/ConstantFactoryTest.java
@@ -17,32 +17,32 @@ public class ConstantFactoryTest {
 
     public static final BigInteger TWOS_DIFF = BigInteger.valueOf(2).add(BigInteger.valueOf(Long.MAX_VALUE)).add(BigInteger.valueOf(Long.MAX_VALUE));
 
-    @Property(trials = 1000000)
-    public void longConstant(final long input) {
+    @Test public void longConstant0() { checkLongConstant(0L); }
+    @Property public void longConstant8(@InRange(min = "-128", max = "127") final long input) { checkLongConstant(input); }
+    @Property public void longConstant16(@InRange(min = "-32768", max = "32767") final long input) { checkLongConstant(input); }
+    @Property public void longConstant24(@InRange(min = "-8388608", max = "8388607") final long input) { checkLongConstant(input); }
+    @Property public void longConstant32(@InRange(min = "-2147483648", max = "2147483647") final long input) { checkLongConstant(input); }
+    @Property public void longConstant40(@InRange(min = "-549755813888", max = "549755813887") final long input) { checkLongConstant(input); }
+    @Property public void longConstant48(@InRange(min = "-140737488355328", max = "140737488355327") final long input) { checkLongConstant(input); }
+    @Property public void longConstant56(@InRange(min = "-36028797018963968", max = "36028797018963967") final long input) { checkLongConstant(input); }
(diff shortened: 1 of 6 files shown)


rdvdijk:
We might be able to check for JDK version inside these tests, and re-enable Java 7 support for all other tests...

core/src/test/java/io/parsingdata/metal/expression/value/ConstantFactoryTest.java
```diff
@@ -0,0 +1,55 @@
...
+            final long maxValue = BigInteger.valueOf(2).pow(i-1).longValue();
+            if ((maxValue + input) >= 0) {
+                return BigInteger.valueOf(input).add(BigInteger.valueOf(2*maxValue));
+            }
+        }
+        return BigInteger.valueOf(input).add(TWOS_DIFF);
+    }
+
+    @Property
+    public void bigIntegerConstant(@InRange(min = "-1000000000000000000", max = "1000000000000000000") final BigInteger input) {
```
rdvdijk: Shouldn't we have the full range of sizes here as well? (e.g. `bigIntegerConstant8`, `bigIntegerConstant16`, etc.?)

core/src/test/java/io/parsingdata/metal/expression/value/ConstantFactoryTest.java
```diff
@@ -0,0 +1,55 @@
...
+    private void checkLongConstant(final long input) {
+        assertEquals(input, ConstantFactory.createFromNumeric(input, signed()).asNumeric().longValue());
+        if (input >= 0) {
+            assertEquals(input, ConstantFactory.createFromNumeric(input, enc()).asNumeric().longValue());
+        } else {
+            assertEquals(0, calculateUnsignedValue(input).compareTo(ConstantFactory.createFromNumeric(input, enc()).asNumeric()));
+        }
+    }
+
+    private BigInteger calculateUnsignedValue(final long input) {
```
rdvdijk: This deserves a comment.

core/src/test/java/io/parsingdata/metal/expression/value/ConstantFactoryTest.java
jvdb: @rdvdijk `BigInteger` has arbitrary sizes so "full range" is not realistic :)

But I see your point. Combining is also possible -- turn the values in the other tests into `BigInteger`s and then perform that test. Then add two specific tests for larger numbers (both positive and negatives values >64 bits).

rdvdijk:
How about we add this to the `ConstantFactoryTest` file, re-enable Java 7 and see if Travis agrees?

`assumeTrue()` documentation states:
> If called with an expression evaluating to false, the test will halt and be ignored.

```java
    @Before
    public void checkJDK() {
        final String specVersion = System.getProperty("java.specification.version");
        final String[] specVersionParts = specVersion.split("\\.");
        final int majorVersion = Integer.parseInt(specVersionParts[0]);
        final int minorVersion = Integer.parseInt(specVersionParts[1]);
        assumeTrue("Tests only run on Java >= 8 (java.specification.version is: " + specVersion + ")", majorVersion >= 1 && minorVersion >= 8);
    }
```

jvdb:
We can try but I don't think it'll work -- the class references junit-quickcheck _before_ executing the code so it will give the same error. But like stated earlier -- we will lose 1.7 compatibility down the road anyway, so no point in investing heavily in keeping it.

jvdb:
This issue has instead been addressed by #116.

Issue #85, opened 2016-09-08
Clean up creating of ParseResult instances


rdvdijk:
The `Token` implementations are now littered with calls to the `ParseResult` constructor.

Let's investigate if we can clean this up by providing static factory methods instead.

PR #86, opened 2016-09-08
#85 Provide and use ParseResult factory methods.

rdvdijk:
@jvdb What do you think?

(I branched the `dynamic-callbacks` branch here, by the way.)

### DIFF
--- a/core/src/main/java/io/parsingdata/metal/token/Seq.java
+++ b/core/src/main/java/io/parsingdata/metal/token/Seq.java
@@ -24,7 +24,7 @@ import io.parsingdata.metal.encoding.Encoding;
 import java.io.IOException;
 
 import static io.parsingdata.metal.Util.checkContainsNoNulls;
-import static io.parsingdata.metal.data.ParseResult.fail;
+import static io.parsingdata.metal.data.ParseResult.failure;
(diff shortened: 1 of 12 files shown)


core/src/main/java/io/parsingdata/metal/token/Seq.java
```diff
@@ -39,15 +41,22 @@ public Seq(final String name, final Encoding enc, final Token... tokens) {
         return tokens.clone();
     }
 
+    /**
```
rdvdijk: Oops!

jvdb:
Looks great! My only nitpick that `success()` is a noun and `fail()` a verb. So it should probably either be `success()` and `error()` or `failure()` (nouns) or `succeed()` and `fail()` (verbs).

But I see how you came to the current ones, they're (by themselves) the best options...

rdvdijk:
Verb and noun, well I'll be damned. Totally missed that. How about we go with `success()` and `failure()`, just because they're both 7 characters long? 😉

jvdb:
Agreed, those are probably the best same-type words :)

rdvdijk:
@jvdb Renamed.

PR #89, opened 2016-09-14
Add dynamic callbacks mechanism as alternative for StructSink

### LINKED ISSUE #12: Let StructSinks be attached dynamically
Right now a callback can be assigned to a token by wrapping it in a `Str`. Instead of this static mapping, a more dynamic approach is more useful, which will allow dynamically turning callbacks on and off and allow adding callbacks to any token at any time.

Together with #11 this will make `Str` obsolete.

jvdb:
Resolves #12.

### DIFF
--- a/core/src/main/java/io/parsingdata/metal/data/Environment.java
+++ b/core/src/main/java/io/parsingdata/metal/data/Environment.java
@@ -16,10 +16,10 @@
 
 package io.parsingdata.metal.data;
 
+import static io.parsingdata.metal.Util.checkNotNull;
+
 import io.parsingdata.metal.data.callback.TokenCallbackList;
 import io.parsingdata.metal.token.Token;
-
-import static io.parsingdata.metal.Util.checkNotNull;
 
 public class Environment {
 
@@ -35,24 +35,20 @@ public class Environment {
         this.callbacks = checkNotNull(callbacks, "callbacks");
     }
 
-    public Environment(final ParseGraph order, final ByteStream input, final long offset) {
-        this(order, input, offset, TokenCallbackList.EMPTY);
-    }
-
     public Environment(final ByteStream input, final long offset, final TokenCallbackList callbacks) {
         this(ParseGraph.EMPTY, input, offset, callbacks);
     }
 
     public Environment(final ByteStream input, final long offset) {
-        this(ParseGraph.EMPTY, input, offset, TokenCallbackList.EMPTY);
+        this(input, offset, TokenCallbackList.EMPTY);
     }
 
     public Environment(final ByteStream input, final TokenCallbackList callbacks) {
-        this(ParseGraph.EMPTY, input, 0L, callbacks);
+        this(input, 0L, callbacks);
     }
 
     public Environment(final ByteStream input) {
-        this(ParseGraph.EMPTY, input, 0L, TokenCallbackList.EMPTY);
+        this(input, 0L);
     }
 
     public Environment addBranch(final Token token) {
--- a/core/src/main/java/io/parsingdata/metal/data/Environment.java
+++ b/core/src/main/java/io/parsingdata/metal/data/Environment.java
@@ -56,18 +56,18 @@ public class Environment {
         this(ParseGraph.EMPTY, input, 0L, TokenCallbackList.EMPTY);
     }
 
-    public void handleCallbacks(final Token token, final Environment environment, final Encoding encoding, final ParseItem item) {
-        handleCallbacks(callbacks, token, environment, encoding, item);
+    public void handleCallbacks(final Token token, final ParseResult result, final Encoding encoding) {
+        handleCallbacks(callbacks, token, result, encoding);
     }
 
-    private void handleCallbacks(final TokenCallbackList callbacks, final Token token, final Environment environment, final Encoding encoding, final ParseItem item) {
+    private void handleCallbacks(final TokenCallbackList callbacks, final Token token, final ParseResult result, final Encoding encoding) {
         if (callbacks.isEmpty()) {
             return;
         }
         if (callbacks.head.token == token) {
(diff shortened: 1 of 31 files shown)


rdvdijk:
Coverage went down, I think we can remove some dead code from the `TraceCallbackList`, add an explicit test for the `BaseCallback` and the new `Environment` constructors.

core/src/main/java/io/parsingdata/metal/data/Environment.java
```diff
@@ -16,29 +16,68 @@
...
+    }
+
+    public Environment add(final ParseRef parseRef) {
+        return new Environment(order.add(parseRef), input, offset, callbacks);
     }
 
     @Override
     public String toString() {
-        return "stream: " + input + "; offset: " + offset + "; order: " + order;
+        return "stream: " + input + "; offset: " + offset + "; order: " + order + "; callbacks: " + callbacks;
```
rdvdijk: (Apparently there is a new way of reviewing on GitHub, let's see!)

We need a test for `toString()` with actual callbacks. The current `toString()` tests `Environment`s without callbacks.

core/src/main/java/io/parsingdata/metal/data/ParseResult.java
```diff
@@ -26,6 +29,28 @@ public ParseResult(final boolean succeeded, final Environment environment) {
         this.environment = environment;
     }
 
+    public void handleCallbacks(final Token token) {
+        handleCallbacks(environment.callbacks, token);
+    }
+
+    private void handleCallbacks(final TokenCallbackList callbacks, final Token token) {
```
rdvdijk: I don't think we have explicit tests yet for this code, add some.

core/src/main/java/io/parsingdata/metal/data/callback/BaseCallback.java
```diff
@@ -0,0 +1,38 @@
...
+ * limitations under the License.
+ */
+
+package io.parsingdata.metal.data.callback;
+
+import io.parsingdata.metal.data.Environment;
+import io.parsingdata.metal.data.ParseResult;
+import io.parsingdata.metal.token.Token;
+
+public class BaseCallback implements Callback {
```
rdvdijk: Add some tests for this class.

core/src/main/java/io/parsingdata/metal/data/callback/Callback.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+package io.parsingdata.metal.data.callback;
+
+import io.parsingdata.metal.data.ParseResult;
+import io.parsingdata.metal.token.Token;
+
+public interface Callback {
+
+    void handle(final Token token, final ParseResult result);
```
rdvdijk: Let's get into the habit of adding Javadoc to new code...

core/src/main/java/io/parsingdata/metal/data/callback/TokenCallbackList.java
```diff
@@ -0,0 +1,65 @@
...
+
+    public static TokenCallbackList create(final TokenCallback head) {
+        return EMPTY.add(checkNotNull(head, "head"));
+    }
+
+    public TokenCallbackList add(final TokenCallback head) {
+        return new TokenCallbackList(checkNotNull(head, "head"), this);
+    }
+
+    public TokenCallbackList add(final TokenCallbackList list) {
```
rdvdijk: I think this method is dead code, remove it.

core/src/test/java/io/parsingdata/metal/ToStringTest.java
```diff
@@ -105,7 +105,7 @@ public void encoding() {
     @Test
     public void data() {
         final Environment environment = stream(1, 2);
-        final String envString = "stream: InMemoryByteStream(2); offset: 0; order: graph(EMPTY)";
+        final String envString = "stream: InMemoryByteStream(2); offset: 0; order: graph(EMPTY); callbacks: ";
```
rdvdijk: (See my earlier "we need a test for toString()" :wink:)

core/src/test/java/io/parsingdata/metal/data/ByNameTest.java
```diff
@@ -25,7 +26,7 @@
 
     @Test
     public void getValueOnEmpty() {
-        assertNull(ByName.getValue(ParseGraph.EMPTY, "name"));
+        TestCase.assertNull(ByName.getValue(ParseGraph.EMPTY, "name"));
```
rdvdijk: Not sure what happened here, but use a static import as before. Also: I think we shouldn't use `TestCase.*` but `Assert.*`.

core/src/main/java/io/parsingdata/metal/data/callback/BaseCallback.java
jvdb: It is tested, but because it is not abstract, coverage requires a test that doesn't inherit from `BaseCallback` in order to execute the `handleSuccess()` and `handleFailure()` methods. I've reverted it to an abstract class.

Clients can trivially recreate the adapter pattern if they need it. Providing an abstract class seems cleaner for Metal I think.

core/src/main/java/io/parsingdata/metal/data/callback/Callback.java
jvdb: I've added #9 to the 5.0.0 milestone. It's better to document the important stuff first I think.

core/src/main/java/io/parsingdata/metal/data/ParseResult.java
jvdb: Excellent point! I had a hard time coming up with a good test however, which indicated that it probably shouldn't be a public API. To fix this I moved the method to the `Token` base class, which greatly simplified the implementation.

PR #93, opened 2016-09-28
Implement public API to retrieve all root items for a specific Token

### LINKED ISSUE #88: Add API to retrieve a collection's base item
An example:

`seq(a, b, c)` will create three `ParseGraph` objects when successfully parsed:

1. Containing `c`,
2. Containing `b` with `c` as a child,
3. Containing `a` with `b` as a child.

It is often desirable (e.g., in callbacks or when processing complete results) to only receive the item from 3, which contains the entire collection. An API should be created that makes this possible.

For example, `ByToken.getRoot(token)` or `ByToken.getBase(token)`.

jvdb:
Resolves #88.

### DIFF
--- a/core/src/main/java/io/parsingdata/metal/data/selection/ByToken.java
+++ b/core/src/main/java/io/parsingdata/metal/data/selection/ByToken.java
@@ -59,11 +59,6 @@ public final class ByToken {
         return results;
     }
 
-    /**
-     * @param graph The graph to search
-     * @param definition The token to match the ParseItem's definition field
-     * @return All values with the provided name in this graph
-     */
     public static ParseValueList getAllValues(final ParseGraph graph, final Token definition) {
         checkNotNull(graph, "graph");
         checkNotNull(definition, "definition");
@@ -91,7 +86,7 @@ public final class ByToken {
         final ParseItemList result = (item.getDefinition() == definition && parent.getDefinition() != definition ? ParseItemList.create(item) : ParseItemList.EMPTY);
         if (item.isGraph() && !item.asGraph().isEmpty()) {
             return result.add(getAllRootsRecursive(item.asGraph().head, item.asGraph(), definition))
-                .add(getAllRootsRecursive(item.asGraph().tail, item.asGraph(), definition));
+                         .add(getAllRootsRecursive(item.asGraph().tail, item.asGraph(), definition));
         }
         return result;
     }
--- a/core/src/main/java/io/parsingdata/metal/data/selection/ByToken.java
+++ b/core/src/main/java/io/parsingdata/metal/data/selection/ByToken.java
@@ -85,7 +85,11 @@ public final class ByToken {
         return getAllRootsRecursive(graph, definition);
     }
 
-    private static ParseItemList getAllRootsRecursive(final ParseGraph graph, final Token definition) {
-        return null;
+    private static ParseItemList getAllRootsRecursive(final ParseItem item, final Token definition) {
+        if (item.getDefinition() == definition) { return ParseItemList.create(item); }
+        if (item.isGraph() && !item.asGraph().isEmpty()) {
+            return getAllRootsRecursive(item.asGraph().head, definition).add(getAllRootsRecursive(item.asGraph().tail, definition));
+        }
+        return ParseItemList.EMPTY;
     }
 }
--- a/core/src/main/java/io/parsingdata/metal/data/selection/ByToken.java
+++ b/core/src/main/java/io/parsingdata/metal/data/selection/ByToken.java
@@ -16,13 +16,13 @@
 
 package io.parsingdata.metal.data.selection;
 
+import static io.parsingdata.metal.Util.checkNotNull;
+
 import io.parsingdata.metal.data.ParseGraph;
 import io.parsingdata.metal.data.ParseItem;
 import io.parsingdata.metal.data.ParseItemList;
 import io.parsingdata.metal.data.ParseValueList;
 import io.parsingdata.metal.token.Token;
-
-import static io.parsingdata.metal.Util.checkNotNull;
 
 public final class ByToken {
 
@@ -79,9 +79,13 @@ public final class ByToken {
         return tailResults;
     }
     
-    public static ParseItemList getAllRoots(ParseGraph graph, Token definition) {
+    public static ParseItemList getAllRoots(final ParseGraph graph, final Token definition) {
         checkNotNull(graph, "graph");
         checkNotNull(definition, "definition");
+        return getAllRootsRecursive(graph, definition);
+    }
+
+    private static ParseItemList getAllRootsRecursive(final ParseGraph graph, final Token definition) {
         return null;
     }
 }
--- a/core/src/main/java/io/parsingdata/metal/data/selection/ByToken.java
+++ b/core/src/main/java/io/parsingdata/metal/data/selection/ByToken.java
@@ -77,9 +77,10 @@ public final class ByToken {
     public static ParseItemList getAllRoots(final ParseGraph graph, final Token definition) {
         checkNotNull(graph, "graph");
         checkNotNull(definition, "definition");
-        return (graph.getDefinition() == definition && !graph.isEmpty() ? ParseItemList.create(graph) : ParseItemList.EMPTY)
-            .add(getAllRootsRecursive(graph.head, graph, definition))
-            .add(getAllRootsRecursive(graph.tail, graph, definition));
+        final ParseItemList result = graph.getDefinition() == definition ? ParseItemList.create(graph) : ParseItemList.EMPTY;
+        if (graph.isEmpty()) { return result; }
+        return result.add(getAllRootsRecursive(graph.head, graph, definition))
+                     .add(getAllRootsRecursive(graph.tail, graph, definition));
     }
 
     private static ParseItemList getAllRootsRecursive(final ParseItem item, final ParseGraph parent, final Token definition) {
--- a/core/src/main/java/io/parsingdata/metal/data/selection/ByToken.java
+++ b/core/src/main/java/io/parsingdata/metal/data/selection/ByToken.java
@@ -82,14 +82,18 @@ public final class ByToken {
     public static ParseItemList getAllRoots(final ParseGraph graph, final Token definition) {
         checkNotNull(graph, "graph");
         checkNotNull(definition, "definition");
-        return getAllRootsRecursive(graph, definition);
+        return (graph.getDefinition() == definition && !graph.isEmpty() ? ParseItemList.create(graph) : ParseItemList.EMPTY)
+            .add(getAllRootsRecursive(graph.head, graph, definition))
+            .add(getAllRootsRecursive(graph.tail, graph, definition));
     }
 
-    private static ParseItemList getAllRootsRecursive(final ParseItem item, final Token definition) {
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/parsingdata/metal/data/selection/ByToken.java
```diff
@@ -78,5 +78,18 @@ private static ParseValueList getAllValuesRecursive(final ParseGraph graph, fina
...
     }
+    
+    public static ParseItemList getAllRoots(final ParseGraph graph, final Token definition) {
+        checkNotNull(graph, "graph");
+        checkNotNull(definition, "definition");
+        return getAllRootsRecursive(graph, definition);
+    }
 
+    private static ParseItemList getAllRootsRecursive(final ParseItem item, final Token definition) {
+        if (item.getDefinition() == definition) { return ParseItemList.create(item); }
```
rdvdijk: We've discussed this off-line, but I'll also write this here for future reference: We will now miss results from self-referencing `Token`s. In that case a graph can contain another graph with the same `Token` definition that can still be considered a 'root' graph.

And by self-referencing tokens I mean either:

*a*) anonymous `Token`s that parse `this` (e.g. linked lists), or;
*b*) `Token`s that eventually `sub()` to themselves (currently implemented using *a*).

core/src/main/java/io/parsingdata/metal/data/selection/ByToken.java
```diff
@@ -78,5 +78,22 @@ private static ParseValueList getAllValuesRecursive(final ParseGraph graph, fina
         if (head.isGraph()) { return tailResults.add(getAllValuesRecursive(head.asGraph(), definition)); }
         return tailResults;
     }
+    
+    public static ParseItemList getAllRoots(final ParseGraph graph, final Token definition) {
+        checkNotNull(graph, "graph");
+        checkNotNull(definition, "definition");
+        return (graph.getDefinition() == definition && !graph.isEmpty() ? ParseItemList.create(graph) : ParseItemList.EMPTY)
+            .add(getAllRootsRecursive(graph.head, graph, definition))
+            .add(getAllRootsRecursive(graph.tail, graph, definition));
```
rdvdijk: I like this alignment of the two `.add`s, you could do that below as well, although that might look a bit too weird?

core/src/main/java/io/parsingdata/metal/data/selection/ByToken.java
```diff
@@ -78,5 +78,22 @@ private static ParseValueList getAllValuesRecursive(final ParseGraph graph, fina
...
+    }
+
+    private static ParseItemList getAllRootsRecursive(final ParseItem item, final ParseGraph parent, final Token definition) {
+        final ParseItemList result = (item.getDefinition() == definition && parent.getDefinition() != definition ? ParseItemList.create(item) : ParseItemList.EMPTY);
```
rdvdijk: This is a pretty long line.

An idea: we introduce a `boolean hasDefinition(Token definition)` on `ParseItem`. This line would become:

```java
final ParseItemList result = (item.hasDefinition(definition) && !parent.hasDefinition(definition) ? ParseItemList.create(item) : ParseItemList.EMPTY);
```

That's still pretty long, though... Nevermind 😉

core/src/main/java/io/parsingdata/metal/data/selection/ByToken.java
```diff
@@ -78,5 +78,22 @@ private static ParseValueList getAllValuesRecursive(final ParseGraph graph, fina
         if (head.isGraph()) { return tailResults.add(getAllValuesRecursive(head.asGraph(), definition)); }
         return tailResults;
     }
+    
+    public static ParseItemList getAllRoots(final ParseGraph graph, final Token definition) {
+        checkNotNull(graph, "graph");
+        checkNotNull(definition, "definition");
+        return (graph.getDefinition() == definition && !graph.isEmpty() ? ParseItemList.create(graph) : ParseItemList.EMPTY)
```
rdvdijk: Are the `!graph.isEmpty()` checks actually needed?

core/src/test/java/io/parsingdata/metal/data/selection/ByTokenTest.java
```diff
@@ -229,4 +250,35 @@ public void getSubRef() throws IOException {
...
+        assertEquals(2, seqs.size);
+        assertEquals(subSeq, seqs.head.getDefinition());
+        assertEquals(subSeq, seqs.tail.head.getDefinition());
+        final ParseValue c1 = seqs.head.asGraph().head.asValue();
+        assertEquals(3, c1.asNumeric().intValue());
+        final ParseValue c2 = seqs.tail.head.asGraph().head.asValue();
+        assertEquals(3, c2.asNumeric().intValue());
+        assertNotEquals(seqs.head.asGraph().head, seqs.tail.head.asGraph().head);
+    }
+
```
rdvdijk: You need test here to confirm that we can now retrieve graphs created by self-referencing tokens.

jvdb: Agreed! I mentioned this in the commit message :)

core/src/main/java/io/parsingdata/metal/data/selection/ByToken.java
jvdb: Otherwise there are 'NullPointerException's on terminators...

core/src/test/java/io/parsingdata/metal/data/selection/ByTokenTest.java
jvdb: I've added a `Sub`-test but not one that uses self-referencing, mainly because that causes a problem: the `definition` of a `Token` contained in a custom `Token` (anonymous or not) actually doesn't point to the outside class, but the embedded `Token`, which is difficult to access and reference.

jvdb: Added self-referencing test using a custom `Token` implementation that exposes its inner `Token`.

core/src/main/java/io/parsingdata/metal/data/selection/ByToken.java
jvdb: Additionally, it's factored slightly differently in the final commit to deal with an edge case (bonus: full coverage!)

core/src/test/java/io/parsingdata/metal/data/selection/ByTokenTest.java
rdvdijk: Not sure I fully understand this:

> the `definition` of a `Token` contained in a custom `Token` (anonymous or not) actually doesn't point to the outside class, but the embedded `Token`, which is difficult to access and reference

You'll have to explain that to me later, I guess that's easier to explain face-to-face.

core/src/test/java/io/parsingdata/metal/data/selection/ByTokenTest.java
```diff
@@ -229,4 +252,93 @@ public void getSubRef() throws IOException {
...
+        assertEquals(smallSeq, seqs.tail.head.getDefinition());
+        final ParseValue c1 = seqs.head.asGraph().head.asValue();
+        assertEquals(3, c1.asNumeric().intValue());
+        final ParseValue c2 = seqs.tail.head.asGraph().head.asValue();
+        assertEquals(3, c2.asNumeric().intValue());
+        assertNotEquals(seqs.head.asGraph().head, seqs.tail.head.asGraph().head);
+    }
+
+    private HashSet<ParseItem> makeSet(ParseItemList seqs) {
+        final HashSet<ParseItem> items = new HashSet<>();
```
rdvdijk: Interface on the left, implementation on the right.

```java
final Set<ParseItem> items = new HashSet<>();
```

Also make this method return `Set<ParseItem>` and make the `seqs` argument `final`.

core/src/test/java/io/parsingdata/metal/data/selection/ByTokenTest.java
```diff
@@ -229,4 +252,93 @@ public void getSubRef() throws IOException {
...
+        for (ParseItemList current = seqs; current != null && !current.isEmpty(); current = current.tail) {
+            items.add(current.head);
+        }
+        return items;
+    }
+
+    @Test
+    public void getAllRootsMultiSub() throws IOException {
+        final ParseResult result = rep(seq(smallSeq, sub(smallSeq, currentOffset))).parse(stream(1, 2, 1, 2, 1, 2, 1, 2), enc());
```
rdvdijk: Neat 😉

core/src/test/java/io/parsingdata/metal/data/selection/ByTokenTest.java
```diff
@@ -229,4 +252,93 @@ public void getSubRef() throws IOException {
...
+        assertEquals(3, c1.asNumeric().intValue());
+        final ParseValue c2 = seqs.tail.head.asGraph().head.asValue();
+        assertEquals(3, c2.asNumeric().intValue());
+        assertNotEquals(seqs.head.asGraph().head, seqs.tail.head.asGraph().head);
+    }
+
+    private HashSet<ParseItem> makeSet(ParseItemList seqs) {
+        final HashSet<ParseItem> items = new HashSet<>();
+        for (ParseItemList current = seqs; current != null && !current.isEmpty(); current = current.tail) {
+            items.add(current.head);
```
rdvdijk: I don't think that any of our `ParseItem`s implement a `hashCode()`, are you sure this actually works as expected?

core/src/test/java/io/parsingdata/metal/data/selection/ByTokenTest.java
```diff
@@ -229,4 +252,93 @@ public void getSubRef() throws IOException {
...
+        final ParseResult result = rep(seq(smallSeq, sub(smallSeq, currentOffset))).parse(stream(1, 2, 1, 2, 1, 2, 1, 2), enc());
+                                                                                           /* 1: +--------+
+                                                                                           /* 2:       +--------+
+                                                                                           /* 3:             +--------+ */
+        assertTrue(result.succeeded);
+        final ParseItemList seqs = ByToken.getAllRoots(result.environment.order, smallSeq);
+        assertEquals(6, seqs.size); // Three regular and three subs.
+        final HashSet<ParseItem> items = makeSet(seqs);
+        assertEquals(seqs.size, items.size()); // Check that there are no duplicate results.
+        for (ParseItem item : items) {
```
rdvdijk: `final ParseItem`

core/src/test/java/io/parsingdata/metal/data/selection/ByTokenTest.java
```diff
@@ -229,4 +252,93 @@ public void getSubRef() throws IOException {
...
+        public CustomToken() {
+            super("", enc());
+            token = seq(any("a"), opt(this));
+        }
+
+        @Override
+        protected ParseResult parseImpl(String scope, Environment env, Encoding enc) throws IOException {
+            return token.parse(scope, env, enc);
+        }
+    }
```
rdvdijk: I think I now see your point. The `CustomToken` instance itself never ends up as a `definition` of any `ParseItem` in the graph, but the nested `token` does.

jvdb: Indeed!

PR #96, opened 2016-10-07
Fix cycle detection and resolving of ParseRefs

### LINKED ISSUE #92: ByOffset's hasGraphAtRef() and findRef() methods are inconsistent
As shown in the fix to #90, the subgraph returned by `ByOffset`'s methods `hasGraphAtRef()` and `findRef()` are inconsistent at best. This needs a serious rethink to make sure there is both an understandable (for users) and implementable approach to cycle detection and handling.

One idea is to ignore subgraphs that are related to `Sub`, so as to prevent cycle prone traversal in cycle detection itself.

jvdb:
Resolves #92.

Besides the offset, the `Token` that has created a subgraph is also considered when looking for existing subgraphs that can lead to a cycle.

Additionally, a bug has been fixed that would not (immediately) detect a cycle because `Sub`s were traversed when determining a subgraphs lowest offset.


core/src/main/java/io/parsingdata/metal/data/selection/ByType.java
```diff
@@ -16,26 +16,27 @@
...
-
-import static io.parsingdata.metal.Util.checkNotNull;
+import io.parsingdata.metal.data.ParseItemList;
 
 public final class ByType {
 
     private ByType() {}
 
-    public static ParseGraphList getRefs(final ParseGraph graph) {
+    public static ParseItemList getRefs(final ParseGraph graph) {
```
rdvdijk: Now that this has changed from `ParseGraphList` to `ParseItemList`, there is only one place where we still use `ParseGraphList`: the `ByType.getGraphs()` method. We do not use that method anymore (except in tests), so here is an idea: let's remove `ParseGraphList`. Or is there a use case for `getGraphs()`?

If you agree, let's create a separate issue.

core/src/main/java/io/parsingdata/metal/data/selection/ByOffset.java
```diff
@@ -19,53 +19,60 @@
...
-        if (graphs.head.containsValue() && graphs.head.getLowestOffsetValue().getOffset() == ref) {
-            return graphs.head;
+    public static ParseItem findRef(final ParseItemList items, final long ref) {
+        checkNotNull(items, "graphs");
+        if (items.isEmpty()) { return null; }
+        if (items.head.isValue() && items.head.asValue().getOffset() == ref) {
+            return items.head;
         }
-        return null;
+        if (items.head.isGraph() && containsLocalValue(items.head.asGraph()) && ByOffset.getLowestOffsetValue(items.head.asGraph()).getOffset() == ref) {
```
jvdb: Well, if we remove the check then we'll need a bunch of `null` checks around all places where the `getLowest*` methods recurse or call each other.

On the other hand, the whole logic has turned into six methods and is quite tangled. So I will have a look to see whether it can be simplified (with or without the check).

core/src/main/java/io/parsingdata/metal/data/selection/ByOffset.java
```diff
@@ -19,53 +19,60 @@
...
 
-    public static ParseGraph findRef(final ParseGraphList graphs, final long ref) {
-        checkNotNull(graphs, "graphs");
-        if (graphs.isEmpty()) { return null; }
-        final ParseGraph res = findRef(graphs.tail, ref);
-        if (res != null) { return res; }
-        if (graphs.head.containsValue() && graphs.head.getLowestOffsetValue().getOffset() == ref) {
-            return graphs.head;
+    public static ParseItem findRef(final ParseItemList items, final long ref) {
+        checkNotNull(items, "graphs");
```
jvdb: I've decided to use this when working with a list such as `ParseItemList`, to clearly communicate that, as part of list processing, we're essentially _popping_ off the top value for handling.

In cases where we get a `ParseGraph` passed in, I prefer to use `graph.head` and `graph.tail` to communicate that we're traversing a data structure.

core/src/main/java/io/parsingdata/metal/data/selection/ByOffset.java
jvdb: This ended up removing a lot of code! See what you think...

core/src/main/java/io/parsingdata/metal/data/selection/ByType.java
jvdb: This is tangled with the main issue in several ways. In fact, the test code that uses it has already been removed since it were dodgy tests that wanted to work around *really* inspecting the data structure (instead doing some heuristic checks). I rewrote the test code to do a proper `ParseGraph` inspection. But I suppose `ParseGraphList` may as well go now...

jvdb: And it's gone...

jvdb:
Very small decrease in code coverage caused by removal of a lot of code that was covered :-)

core/src/main/java/io/parsingdata/metal/data/transformation/Reversal.java
```diff
@@ -23,13 +23,17 @@
...
+    }
+
+    private static ParseGraph reverse(final ParseGraph oldGraph, final ParseGraph newGraph) {
         if (oldGraph.isEmpty()) { return newGraph; }
         return reverse(oldGraph.tail, new ParseGraph(reverseItem(oldGraph.head), newGraph, oldGraph.definition));
     }
 
     private static ParseItem reverseItem(final ParseItem item) {
-        return item.isGraph() ? item.asGraph().reverse() : item;
+        return item.isGraph() ? Reversal.reverse(item.asGraph(), ParseGraph.EMPTY) : item;
```
rdvdijk: No need for the `Reversal.` prefix.

PR #99, opened 2016-10-10
Idea: bit support in Metal

gertjanal:
Working code, but intended as an idea for future implementation.

``` java
final Token bits = bits("entry", con(8), // read 8 bytes
    bit("State", con(3), expTrue()), // read 3 bits
    bit("Reserved", con(17), eqNum(con(0))), // read 17 bits, with predicate
    bit("Offset", con(44), expTrue())); // read 44 bits
```


jvdb:
Thanks for the concept. I think that for a full implementation in Metal, there are some issues with this approach: `Bit`s can only be nested in `Bits` and `Bits` only accepts `Bit`s. This breaks the conceptual idea of the combinator approach.

Next, I think we explore an alternative approach where the `size` argument of a `Def` is interpreted as "size in bits" instead of "size in bytes" as now. This is a major breaking change, but would be more conceptually compatible with the typical Metal approach.

Issue #11, opened 2016-03-31
Let all tokens have an (optional) name


jvdb:
Replacing the need for using `Str` to name something explicitly.

Together with #12 this will make `Str` obsolete.

PR #100, opened 2016-10-11
Remove Str Token

### LINKED ISSUE #6: Test whether StructSink receives the correct subgraph
Especially when dealing with recursive or repeating structures, it's important that a `StructSink` on a single such structure receives only the current one.

### LINKED ISSUE #87: Remove Str
Remove the `Str` token. The `Str` exists for three reasons:

1. Attach a callback to some token. Addressed by #12.
2. Create a named scope. Addressed by #11.
3. Allow retrieving a subgraph of some collection type (e.g. `Seq` or `Rep`) without also receiving all smaller subgraphs it contains. Addressed by #88.

rdvdijk:
Resolves #87. Resolves #6.

Now that issues #12 (_attach callback to token_), #11 (_all tokens can have a name_) and #88 (_get root graphs_) have been implemented, we do not need `Str` anymore.


core/src/test/java/io/parsingdata/metal/data/callback/CallbackTest.java
```diff
@@ -77,4 +76,85 @@ protected void handleFailure(Token token, Environment environment) {
...
+                final ParseItemList roots = getAllRoots(environment.order, token);
+                assertEquals(offsets[count++], roots.head.asGraph().tail.head.asValue().getOffset());
+            }
+
+            @Override
+            protected void handleFailure(Token token, Environment environment) {}
+        }));
+    }
+
+    private ParseItemList reverse(final ParseItemList oldList, final ParseItemList newList) {
```
jvdb: This method is unused, can be removed.

PR #104, opened 2016-10-12
Implement TokenRef (shorthand: token(String)) to allow references to existing tokens in recursive data structures.

### LINKED ISSUE #26: Self-referencing tokens can only be implemented using anonymous Token classes
The following 'self-referencing page' `Token` does not compile:

```java
private static final Token PAGE = 
    seq(
        def("pagePayload", 8),
        def("nextPageOffset", 4),
        sub(PAGE, ref("nextPageOffset")
    );
```

We cannot reference `PAGE` because we are defining it here.

To fix this, see the following workaround:

```java
private static final Token PAGE = new Token(null) {
    @Override
        return seq(
                def("pagePayload", 8),
                def("nextPageOffset", 4),
                sub(this, ref("nextPageOffset"))
        ).parse(scope, env, enc);
    }
};
```

Note the `sub(this, ...)` construct. This is a trick to reference back to the anonymous `PAGE` `Token` instance.

We need a way to express such a self-referencing token.

The main challenge in implementing such a token is that `Token`s are not aware of their 'parent' `Token`, so we can not locate it.

jvdb:
Resolves #26.

### DIFF
--- a/core/src/main/java/io/parsingdata/metal/Shorthand.java
+++ b/core/src/main/java/io/parsingdata/metal/Shorthand.java
@@ -19,15 +19,56 @@ package io.parsingdata.metal;
 import io.parsingdata.metal.encoding.Encoding;
 import io.parsingdata.metal.expression.Expression;
 import io.parsingdata.metal.expression.True;
-import io.parsingdata.metal.expression.comparison.*;
-import io.parsingdata.metal.expression.logical.*;
-import io.parsingdata.metal.expression.value.*;
-import io.parsingdata.metal.expression.value.arithmetic.*;
+import io.parsingdata.metal.expression.comparison.ComparisonExpression;
+import io.parsingdata.metal.expression.comparison.Eq;
+import io.parsingdata.metal.expression.comparison.EqNum;
+import io.parsingdata.metal.expression.comparison.EqStr;
+import io.parsingdata.metal.expression.comparison.GtNum;
+import io.parsingdata.metal.expression.comparison.LtNum;
+import io.parsingdata.metal.expression.logical.And;
+import io.parsingdata.metal.expression.logical.BinaryLogicalExpression;
+import io.parsingdata.metal.expression.logical.Not;
+import io.parsingdata.metal.expression.logical.Or;
+import io.parsingdata.metal.expression.logical.UnaryLogicalExpression;
+import io.parsingdata.metal.expression.value.BinaryValueExpression;
+import io.parsingdata.metal.expression.value.Cat;
+import io.parsingdata.metal.expression.value.Const;
+import io.parsingdata.metal.expression.value.ConstantFactory;
+import io.parsingdata.metal.expression.value.Elvis;
+import io.parsingdata.metal.expression.value.FoldLeft;
+import io.parsingdata.metal.expression.value.FoldRight;
+import io.parsingdata.metal.expression.value.Reducer;
+import io.parsingdata.metal.expression.value.UnaryValueExpression;
+import io.parsingdata.metal.expression.value.Value;
+import io.parsingdata.metal.expression.value.ValueExpression;
+import io.parsingdata.metal.expression.value.arithmetic.Add;
+import io.parsingdata.metal.expression.value.arithmetic.Div;
+import io.parsingdata.metal.expression.value.arithmetic.Mod;
+import io.parsingdata.metal.expression.value.arithmetic.Mul;
+import io.parsingdata.metal.expression.value.arithmetic.Neg;
 import io.parsingdata.metal.expression.value.arithmetic.Sub;
 import io.parsingdata.metal.expression.value.bitwise.ShiftLeft;
 import io.parsingdata.metal.expression.value.bitwise.ShiftRight;
-import io.parsingdata.metal.expression.value.reference.*;
-import io.parsingdata.metal.token.*;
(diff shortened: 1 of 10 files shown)


core/src/main/java/io/parsingdata/metal/Shorthand.java
```diff
@@ -99,7 +140,7 @@ private Shorthand() {}
     public static final ValueExpression self = new Self();
     public static ValueExpression len(final ValueExpression operand) { return new Len(operand); }
     public static ValueExpression ref(final String name) { return new NameRef(name); }
-    public static ValueExpression ref(final Token definition) { return new TokenRef(definition); }
+    public static ValueExpression ref(final Token definition) { return new io.parsingdata.metal.expression.value.reference.TokenRef(definition); }
```
rdvdijk: Ah, yes 😁

core/src/main/java/io/parsingdata/metal/token/Token.java
```diff
@@ -68,6 +68,8 @@ public boolean isLocal() {
         return true;
     }
 
+    public Token getDefinition(final Environment env) { return this; }
```
rdvdijk: Not sold yet on the name of this method. Until now, the term "definition" only had meaning inside a `ParseGraph`, where each item is *defined* by a `Token`. This might confuse us in the future.

Not sure what to call this one, though... For lack of a better name, I guess this will have to do.

core/src/main/java/io/parsingdata/metal/token/TokenRef.java
```diff
@@ -0,0 +1,60 @@
...
+    @Override
+    protected ParseResult parseImpl(String scope, Environment env, Encoding enc) throws IOException {
+        final Token referenced = findToken(env.order, refName);
+        if (referenced == null) { return failure(env); }
+        return referenced.parse(scope, env, enc);
+    }
+
+    private Token findToken(final ParseItem item, final String refName) {
+        if (item.getDefinition().name.equals(refName)) { return item.getDefinition(); }
+        if (item.isGraph() && !item.asGraph().isEmpty()) {
```
rdvdijk: How about we invert this `if` and keep the main logic at the end of the method?

```java
    private Token findToken(final ParseItem item, final String refName) {
        if (item.getDefinition().name.equals(refName)) { return item.getDefinition(); }
        if (!item.isGraph() || item.asGraph().isEmpty()) { return null; } // * ugh... see below
        final Token headResult = findToken(item.asGraph().head, refName);
        if (headResult != null) { return headResult; }
        return findToken(item.asGraph().tail, refName);
    }
```

Although that second `if` might be a bit easier to understand like this?

```java
if (!(item.isGraph() && !item.asGraph().isEmpty())) { return null; }
```

core/src/test/java/io/parsingdata/metal/TreeTest.java
```diff
@@ -38,35 +37,26 @@
...
-                    ).parse(scope, env.addBranch(this), enc);
-                if (result.succeeded) { return success(result.environment.closeBranch()); }
-                return failure(env);
-            }
-        };
+        seq("tree",
+            def("head", con(1), eq(con(HEAD))),
+            def("nr", con(1)),
+            def("left", con(1)),
+            pre(sub(token("tree"), last(ref("left"))), not(eq(last(ref("left")), con(0)))),
```
rdvdijk: Much, *much* cleaner.

core/src/test/java/io/parsingdata/metal/token/TokenRefTest.java
```diff
@@ -0,0 +1,71 @@
...
+        // Clearly reference the second:
+        assertTrue(createNamedTokens("out", "in", "in").parse(stream(21, 42), enc()).succeeded);
+        assertTrue(createNamedTokens("out", "in", "in").parse(stream(21, 21, 42), enc()).succeeded);
+        assertTrue(createNamedTokens("out", "in", "in").parse(stream(21, 21, 21, 42), enc()).succeeded);
+        // Clearly reference the first:
+        assertTrue(createNamedTokens("in", "out", "in").parse(stream(21, 42), enc()).succeeded);
+        assertTrue(createNamedTokens("in", "out", "in").parse(stream(21, 42, 21, 42), enc()).succeeded);
+        // Reference the first:
+        assertTrue(createNamedTokens("in", "in", "in").parse(stream(21, 42), enc()).succeeded);
+        assertTrue(createNamedTokens("in", "in", "in").parse(stream(21, 42, 21, 42), enc()).succeeded);
```
rdvdijk: Why does this reference the first? The code takes the `env.order` and looks at its `head`/`tail` for the first definition that matches the name. Wouldn't the `secondSeq` be the first token that would be found, since we search "inside out"?

core/src/main/java/io/parsingdata/metal/token/TokenRef.java
jvdb: Double negation? Please no :-)

jvdb: Added `FAILED_LOOKUP` static `Token` that always fails. Inverted the if to return it directly (guard clause).

core/src/test/java/io/parsingdata/metal/token/TokenRefTest.java
jvdb: We've discussed this.

PR #105, opened 2016-10-13
Nth value expression

### LINKED ISSUE #103: Nth() ValueExpression
Example: Suppose `ref("a")` returns a list with values `[1, 2, 3]` and `ref("x")` returns `[0, 2]` then:

`nth(ref("a"), ref("x"))`

should return `[1, 3]`, which are the 0th and 2nd values of the first list.

ccreeten:
Resolves #103.


jvdb:
There is an awful amount of code duplication in `NthExpressionTest`. Maybe you can make a method that takes an `Environment` and an `integer` (size) as arguments, runs `parse()`, checks for success (is now missing) and returns the `order` for further checking.

jvdb:
There is one branch missing [according to code coverage](https://codecov.io/gh/parsingdata/metal/src/34122cac2094d631cf720fd235623c5c104b5bc6/core/src/main/java/io/parsingdata/metal/expression/value/reference/Nth.java).

ccreeten:
I'll take a look at the test code. For the branch miss, if I'm mot mistaken, this is because the test for index < valueCount && index >= 0 being both false can never happen, since the minimum valueCount is 0.

core/src/main/java/io/parsingdata/metal/expression/value/reference/Nth.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    private OptionalValueList eval(final OptionalValueList values, final OptionalValueList indices) {
+        if (indices.isEmpty()) {
+            return OptionalValueList.EMPTY;
+        }
+        if (indices.head.isPresent()) {
+            final BigInteger index = indices.head.get().asNumeric();
+            final BigInteger valueCount = BigInteger.valueOf(values.size);
+            if (index.compareTo(valueCount) < 0 && index.compareTo(BigInteger.ZERO) >= 0) {
```
rdvdijk: This looks great. A bit more complex than I expected, but I hadn't considered the need to handle lists of values.

Because I always have to say *something* in a review, get ready for some nitpicks: maybe statically import `ZERO` and `ONE`.

core/src/main/java/io/parsingdata/metal/expression/value/reference/Nth.java
```diff
@@ -0,0 +1,81 @@
...
+ * <p>
+ * Should an index be invalid, an empty value is added instead. The resulting list is
+ * therefore always of the same size as the list of indices.
+ */
+public class Nth implements ValueExpression {
+
+    private final ValueExpression values;
+    private final ValueExpression indices;
+
+    public Nth(final ValueExpression values, final ValueExpression indices) {
```
rdvdijk: Can you add `Nth` to `ArgumentsTest`? And while you're at it, I now see that we've forgotten to add `Last` there, so add that one while you're at it :wink:.

core/src/main/java/io/parsingdata/metal/expression/value/reference/Nth.java
```diff
@@ -0,0 +1,81 @@
...
+        return eval(values, indices.tail).add(OptionalValue.empty());
+    }
+
+    private OptionalValue nth(final OptionalValueList values, final BigInteger index) {
+        if (index.equals(BigInteger.ZERO)) {
+            return values.head;
+        }
+        return nth(values.tail, index.subtract(BigInteger.ONE));
+    }
+
```
rdvdijk: Add a `toString()` (and add `nth` somewhere in `StringTest`).

ccreeten:
@jvdb was right about the branch missing, as an and operation in Java is short-circuited. The problem was that the default encoding is unsigned, so the negative number was in reality not negative, which meant it wasn't tested for this.

core/src/main/java/io/parsingdata/metal/expression/value/reference/Nth.java
ccreeten: Done.

core/src/main/java/io/parsingdata/metal/expression/value/reference/Nth.java
ccreeten: Done.

core/src/main/java/io/parsingdata/metal/expression/value/reference/Nth.java
ccreeten: Done.

PR #106, opened 2016-10-14
Add long size support to Shorthand.nod

gertjanal:


jvdb:
Looks good!

gertjanal:
Any idea where the `will increase coverage by <.01%` comes from?

jvdb:
Well, your new code has 100% coverage, so that nudges the total coverage slightly up :)

PR #107, opened 2016-10-14
Make gpg-plugin user friendly, only run on profile 'release'

gertjanal:
`mvn clean verify -P release` calls gpg, disabled by default.

Note that the gpg.skip can also be moved to the `<properties>` section so the `activeByDefault` profile can be removed, but I think this way it is more clear


jvdb:
Looks good!

PR #114, opened 2016-10-26
Style improvements

### LINKED ISSUE #7: Apply code style
Naming and other issues have been resolved (see #82, #58, #59, #76, etc.), and now the only style issue left is indenting/whitespace. We've arrived at the following decision (apart from other style issues, which are essentially all standard Java things), with regard to one non-standard practice in Metal, namely lines of the following form:

`conditional(...) { statement; }` and
`methodSignature(...) { return statement; }`

are only allowed if `statement` is one of the following:

1. A return statement returning a simple value (e.g., variable, field, literal), such as `return 5;` or `return value;`.
2. A return statement doing a trivial boolean compare, such as `return size != 0;` or `return !x;`.
3. A throw statement, such as `throw new UnsupportedOperationException();`

In all other cases, the lines should be broken up according to standard formatting rules, such as:

```Java
conditional() {
    statement;
}
```

jvdb:
Resolves #7: Unifies use of single line blocks/methods. Extracted constant for empty `Token` names.

### DIFF
--- a/core/src/main/java/io/parsingdata/metal/data/ParseValueList.java
+++ b/core/src/main/java/io/parsingdata/metal/data/ParseValueList.java
@@ -55,9 +55,8 @@ public class ParseValueList {
 
     public ParseValue get(final String name) {
         if (isEmpty()) { return null; }
-        if (head.matches(name)) {
-            return head;
-        } else {
+        if (head.matches(name)) { return head; }
+        else {
             return tail.get(name);
         }
     }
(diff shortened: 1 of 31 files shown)


core/src/main/java/io/parsingdata/metal/data/ParseValueList.java
```diff
@@ -55,18 +55,18 @@ public ParseValueList add(final ParseValueList list) {
...
     }
 
     public ParseValueList getAll(final String name) {
         if (isEmpty()) { return this; }
         final ParseValueList tailList = tail.getAll(name);
-        if (head.matches(name)) { return tailList.add(head); }
-        else { return tailList; }
+        if (head.matches(name)) {
+            return tailList.add(head);
+        } else { return tailList; }
```
rdvdijk: This looks so counterintuitive, to have a multi-line `if` and a single line `else`. But I guess it does follow our rule 😉

PR #115, opened 2016-10-27
Add generic callbacks

### LINKED ISSUE #28: Generic error callbacks
Issue #89 added callbacks for `Token`s. We need a way to assign callbacks to any error, unrelated to a specific `Token` instance.

rdvdijk:
Resolves #28


rdvdijk:
I wonder if we should improve the `Callbacks` API before merging. Are you OK with these three constructors? We could also do something like:

```java
Callbacks callbacks = Callbacks.add(generic).add(token1, tokenCallback).add(token2, tokenCallback);
```

That would hide the `TokenCallbackList` class from the end-user.

PR #116, opened 2016-10-27
Improve conversion tests and refactor Callbacks interface

### LINKED ISSUE #83: Improve value conversion code tests

jvdb:
Resolves #83, improves #28.


core/src/test/java/io/parsingdata/metal/data/callback/CallbackTest.java
```diff
@@ -66,20 +67,20 @@ public void testHandleCallback() throws IOException {
...
         final Token cho = cho("cho", ONE, TWO);
         final Token sequence = seq("seq", cho, ONE);
-        final TokenCallbackList callbacks = TokenCallbackList
-            .create(new TokenCallback(ONE, countingCallback))
-            .add(new TokenCallback(TWO, countingCallback))
-            .add(new TokenCallback(cho, countingCallback))
-            .add(new TokenCallback(sequence, countingCallback));
-        final Environment environment = new Environment(new InMemoryByteStream(new byte[] { 2, 1 }), new Callbacks(callbacks));
+        final Callbacks callbacks =
+            NONE.add(ONE, countingCallback)
```
rdvdijk: I don't really like this `NONE.add` notation. How about we just use a default constructor?

```java
Callbacks callbacks = new Callbacks().add(ONE, countingCallback);
```

Or a factory method:

```java
Callbacks callbacks = Callbacks.create().add(ONE, countingCallback);
```

I'm fine that we use things like `EMPTY` lists internally, but this is client facing API. `NONE.add` just looks 'too' weird.

jvdb: I've added a factory method `create()` to `Callbacks` that returns `NONE`. I've kept `NONE` public, because in cases where you want to state that you want no callbacks, passing `NONE` actually looks better than passing `Callbacks.create()` IMHO.

PR #118, opened 2016-10-27
Added divide ceil() and divide floor() ValueExpressions

gertjanal:
```java
assertEquals(
    Math.ceil(8.0 / 3.0),
    divCeil(ref("left"), ref("right"))
        .eval(result.environment, ENC).head.get().asNumeric().intValue(), 0);

assertEquals(
    Math.floor(8.0 / 3.0),
    divFloor(ref("left"), ref("right"))
        .eval(result.environment, ENC).head.get().asNumeric().intValue(), 0);
```

### DIFF
--- a/core/src/main/java/io/parsingdata/metal/Shorthand.java
+++ b/core/src/main/java/io/parsingdata/metal/Shorthand.java
@@ -42,6 +42,8 @@ import io.parsingdata.metal.expression.value.UnaryValueExpression;
 import io.parsingdata.metal.expression.value.Value;
 import io.parsingdata.metal.expression.value.ValueExpression;
 import io.parsingdata.metal.expression.value.arithmetic.Add;
+import io.parsingdata.metal.expression.value.arithmetic.DivCeil;
+import io.parsingdata.metal.expression.value.arithmetic.DivFloor;
 import io.parsingdata.metal.expression.value.arithmetic.Div;
 import io.parsingdata.metal.expression.value.arithmetic.Mod;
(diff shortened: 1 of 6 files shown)


core/src/main/java/io/parsingdata/metal/Shorthand.java
```diff
@@ -124,6 +126,8 @@ private Shorthand() {}
 
     public static BinaryValueExpression add(final ValueExpression left, final ValueExpression right) { return new Add(left, right); }
     public static BinaryValueExpression div(final ValueExpression left, final ValueExpression right) { return new Div(left, right); }
+    public static BinaryValueExpression divCeil(final ValueExpression left, final ValueExpression right) { return new DivCeil(left, right); }
+    public static BinaryValueExpression divFloor(final ValueExpression left, final ValueExpression right) { return new DivFloor(left, right); }
```
rdvdijk: Why `divCeil` and `divFloor`, and not just `ceil` and `floor`? That way we can just combine `ceil` and `floor` with other `BinaryValueExpression`s (e.g. `ceil(mul(...))` and `ceil(add(...))`).

ccreeten: Probably because current implementations all use BigInteger, which means integer arithmetic, so no use for floor/ceil. 

Not sure if this is the way to go though...

jvdb: Agreed, we should come up with a good model/approach to handle fractions and as a result of that, think about stuff like ceil and floor.

PR #120, opened 2016-10-28
Added pitest to the poms. Reports are generated as well.

### LINKED ISSUE #119: Add automated mutation testing with pitest

jvdb:
Resolves #119.


jvdb:
Yep. I think I'll make an issue first to handle them all in general. Any difficult problems that remain can then receive their own issue.

PR #123, opened 2016-10-31
Reduce recursion

gertjanal:
Take a look at this idea in `ByOffset`

````java
return items.tailStream()
            .filter(list -> {
                return !list.isEmpty();
            })
            .map(ParseItemList::getHead)
            .filter(head -> {
                if (head.isValue() && head.asValue().getOffset() == offset) {
                    return true;
                }
                else if (head.isGraph()) {
                    final ParseValue value = getLowestOffsetValue(head.asGraph(), null);
                    return value != null && value.getOffset() == offset;
                }
                return false;
            }).findFirst().orElse(null);
````

TODO: make `public Stream<ParseItemList> tailStream()` lazy


gertjanal:
This code is too incomplete for now. Brainfart has some flaws

PR #126, opened 2016-11-06
Handle PITest results to improve tests

### LINKED ISSUE #121: Improve code based on mutation testing results
This issue to handle all low-hanging fruit based on starting to use mutation testing. Problems that remain after this first improvement cycle can then be reported as separate issues.

Starting results are 93% line coverage and 94% mutation coverage.

jvdb:
Resolves #121. As discussed in #121, PITest appears to have issues finding all related tests, which makes its output unreliable. That said, some significant fixes have resulted from its use:

1. Generic `Callback` tests were improved to check whether they had executed.
2. `Util.tokensToString()` is now properly tested.
3. Zero-size `Def`s are now explicitly allowed and checked in a test.
4. `Mod` checked whether its right operand was `< 0`, but check should be `<= 0`.
5. Added a specific `toString()` implementation check.

And last but not least, it pointed us towards refactoring the various immutable lists into `ImmutableList<T>`, which is quite valuable.

End result for now is increase from 94% to 96% in mutation coverage, although some of the increase is related to workarounds.


core/src/main/java/io/parsingdata/metal/expression/value/arithmetic/Mod.java
```diff
@@ -34,7 +34,7 @@ public Mod(final ValueExpression left, final ValueExpression right) {
 
     @Override
     public OptionalValue eval(final Value left, final Value right, final Environment environment, final Encoding encoding) {
-        if (right.asNumeric().compareTo(BigInteger.ZERO) < 0) {
+        if (right.asNumeric().compareTo(BigInteger.ZERO) <= 0) {
```
rdvdijk: Nice.

Also: I had forgotten we had actually added `Mod`, I thought it was missing.

core/src/test/java/io/parsingdata/metal/PITestCases.java
```diff
@@ -0,0 +1,71 @@
...
+    public static final Token MUL = seq(any("a"), any("b"), def("c", 1, eqNum(mul(ref("a"), ref("b")))));
+    public static final Token MOD = seq(any("a"), any("b"), def("c", 1, eqNum(mod(ref("a"), ref("b")))));
+    public static final Token DIV = seq(any("a"), any("b"), def("c", 1, eqNum(div(ref("a"), ref("b")))));
+    public static final Token NOT = seq(any("a"), def("not(a)", con(1), eq(not(ref("a")))));
+    public static final Token OR = seq(any("a"), any("b"), def("or(a,b)", con(1), eq(or(ref("a"), ref("b")))));
+    public static final Token SHL = seq(any("a"), def("shl(a,1)", con(1), eq(shl(ref("a"), con(1)))));
+
+    @Test
+    public void handlePITestIssue() throws IOException {
+        // Taken from ArithmeticValueExpressionSemanticsTest:
```
rdvdijk: Why not move these tests into those files?

jvdb: Because those tests are parameterized. So adding them there would cause them to be executed over and over. Also, they are there already -- but PITest doesn't seem to properly consider them. So I chose to extract them to keep the duplication contained.

PR #128, opened 2016-12-01
Create traceability for nested tokens (i.e., when using the Tie token)

### LINKED ISSUE #127: Replace offset in ParseValue with more flexible location mechanism to support fragments
Mainly related to ending up with real traceability around #27.

jvdb:
Resolves #127.


jvdb:
Still need to repair code coverage. Will do that before merging to the token-in-expression branch.

jvdb:
Still requires some tests to be finished:

1. Tie.isLocal() override
2. One branch case in Tie
3. toString() in Source/Slice
4. Some error cases in Source/Slice

Except for 1 they all impact coverage. So focus should be 1 and then getting coverage back to 100%.

jvdb:
1. is probably not an issue, so that we can remove the `isLocal()` override. What it prevents for `Sub` is false positives due to overlapping offsets. But for `Tie`, `offset` is always zero and `Source` a specific object. So I don't think it will lead to false positives there.

jvdb:
Cycle detection for `Tie` will need an extensive rethink. The example found:

````Java
    seq("s",
        def("a", con(1)),
        tie(token("s"), ref("a")));
````

Issue #130, opened 2016-12-10
Implement equality checks for data structures


jvdb:
Currently we do not implement custom equality checks for our data structures (including all core data structures such as `ParseItem`, `Expression`, `ValueExpression` and `Token` (including all implementations) as well as interface data structures such as `Environment`, `Source`, etc.)

A major problem with this is that we do a `==` check on `Source` in cycle detection. This is simply wrong, because the check should determine whether two sources lead to the same state -- which means that two different instances of the same underlying objects such as `environment` or `dataExpression` should be determined as equal.

jvdb:
Some things left to do:
- Get code coverage back to 100%.
- Consider removing `CurrentOffset` or handling the empty case.
- Determine how to handle classes without fields that now return a constant in `hashCode()`.

PR #131, opened 2016-12-11
Parse the result(s) of an evaluated ValueExpression with a Token.

### LINKED ISSUE #27: Token-in-Expression parsing
Providing some `Value` objects and using it as the input stream to parse by a `Token`.

### LINKED ISSUE #127: Replace offset in ParseValue with more flexible location mechanism to support fragments

jvdb:
Resolves #27: Support for the `Tie` token type to parse tokens inside the result(s) of an evaluated `ValueExpression`.
Resolves #127: Converted basic `ByteStream` support to a `Source` and `Slice` system that supports underlying external streams (`ByteStreamSource`), result(s) of evaluated `ValueExpressions` (`DataExpressionSource`) and bootstrapping of constant values (`ConstantSlice` and `ConstantSource`).


core/src/main/java/io/parsingdata/metal/data/selection/ByOffset.java
```diff
@@ -23,32 +23,38 @@
...
             final ParseValue value = getLowestOffsetValue(head.asGraph(), null);
-            if (value != null && value.getOffset() == offset) { return head; }
+            if (value != null && matchesLocation(value, offset, source)) { return head; }
         }
-        return findItemAtOffset(items.tail, offset);
+        return findItemAtOffset(items.tail, offset, source);
+    }
+
+    private static boolean matchesLocation(final ParseValue value, final long offset, final Source source) {
+        return value.slice.offset == offset && value.slice.source == source;
```
rdvdijk: Can we safely use object reference equality here? (`source == source`)

core/src/main/java/io/parsingdata/metal/data/Source.java
```diff
@@ -0,0 +1,30 @@
...
+ * limitations under the License.
+ */
+
+package io.parsingdata.metal.data;
+
+import java.io.IOException;
+
+public abstract class Source {
+
+    public Slice create(final long offset, final int size) throws IOException {
```
rdvdijk: Naming: call this `slice` instead of `create`?

core/src/main/java/io/parsingdata/metal/data/DataExpressionSource.java
```diff
@@ -0,0 +1,62 @@
...
+        if (results.size <= index) { throw new IllegalStateException("ValueExpression dataExpression yields incorrect amount of results."); }
+        final byte[] inputData = getValueAtIndex(results, index, 0).get().getValue();
+        if (offset >= inputData.length) { return new byte[0]; }
+        final int toCopy = (int)offset + size > inputData.length ? inputData.length - (int)offset : size;
+        final byte[] outputData = new byte[toCopy];
+        System.arraycopy(inputData, (int)offset, outputData, 0, toCopy);
+        return outputData;
+    }
+
+    private OptionalValue getValueAtIndex(ImmutableList<OptionalValue> results, int index, int current) {
```
rdvdijk: (Saw some more missing `final`s in this code, you might want to check your IDE settings to catch those for you.)

core/src/test/java/io/parsingdata/metal/data/DataExpressionSourceTest.java
```diff
@@ -0,0 +1,57 @@
...
+        assertEquals(2, value.slice.source.create(2, 4).size);
+        assertEquals(0, value.slice.source.create(4, 4).size);
+    }
+
+    @Test
+    public void indexOutOfBounds() throws IOException {
+        thrown.expect(IllegalStateException.class);
+        thrown.expectMessage("ValueExpression dataExpression yields incorrect amount of results.");
+        final ParseResult result = setupResult();
+        final DataExpressionSource source = new DataExpressionSource(ref("a"), 1, result.environment, enc());
```
jvdb: Correct!

core/src/test/java/io/parsingdata/metal/ToStringTest.java
```diff
@@ -147,28 +151,39 @@ public void encoding() {
...
+        final ParseValue pv2 = new ParseValue("two", NONE, bytesToSlice(new byte[]{3, 4}), enc());
         final String pv2String = "two(0x0304)";
         assertEquals(">" + pv2String + ">" + pv1String, ImmutableList.create(pv1).add(pv2).toString());
         assertEquals(">" + pv2String + ">" + pv1String, ImmutableList.create(pv1).add(pv2).toString());
     }
 
     @Test
+    public void slice() throws IOException {
+        final ParseValue pv1 = new ParseValue("name", NONE, bytesToSlice(new byte[]{1, 2}), enc());
+        assertEquals("0102@0:2", pv1.slice.toString());
```
jvdb: Agreed. Let's put an issue on the roadmap to come up with a good general rule for `toString()` and implement it everywhere :)

See #132.

core/src/main/java/io/parsingdata/metal/data/selection/ByOffset.java
jvdb: Nope, see #130. But it's like this now because it's too big a change for 5.0.0 I think and doesn't directly harm anything. But long-term it's a serious issue.

core/src/main/java/io/parsingdata/metal/data/Source.java
jvdb: I've also renamed `ConstantSlice.bytesToSlice()` to `create()` so that:

`Slice.create()` -> `Slice`
`Source.slice()` -> `Slice`

core/src/main/java/io/parsingdata/metal/data/ByteStreamSource.java
```diff
@@ -0,0 +1,46 @@
...
+        final int readSize = input.read(offset, data);
+        if (readSize == size) { return data; }
+        final byte[] resizedData = new byte[readSize];
+        System.arraycopy(data, 0, resizedData, 0, readSize);
+        return resizedData;
+    }
+
+    @Override
+    public String toString() {
+        return input.toString();
```
jvdb: It generally gets prefixed with "source: " by `Environment`. I try to keep the size down. But we do need to think about it. See #132.

core/src/test/java/io/parsingdata/metal/token/TieTest.java
```diff
@@ -0,0 +1,278 @@
...
+        final byte[] output = new byte[data.length + 4];
+        output[0] = (byte)((data.length & 0xff000000) >> 24);
+        output[1] = (byte)((data.length & 0xff0000) >> 16);
+        output[2] = (byte)((data.length & 0xff00) >> 8);
+        output[3] = (byte) (data.length & 0xff);
+        System.arraycopy(data, 0, output, 4, data.length);
+        return output;
+    }
+
+    public static ValueExpression inflate(final ValueExpression target) {
```
jvdb: I've moved it to `Util` in metal-core instead, so that it can be used in both core/tests and formats/main. Cleanest solution I could come up with :)

core/src/main/java/io/parsingdata/metal/data/DataExpressionSource.java
```diff
@@ -0,0 +1,62 @@
...
+        this.index = index;
+        this.environment = checkNotNull(environment, "environment");
+        this.encoding = checkNotNull(encoding, "encoding");
+    }
+
+    @Override
+    protected byte[] getData(final long offset, final int size) throws IOException {
+        ImmutableList<OptionalValue> results = dataExpression.eval(environment, encoding);
+        if (results.size <= index) { throw new IllegalStateException("ValueExpression dataExpression yields incorrect amount of results."); }
+        final byte[] inputData = getValueAtIndex(results, index, 0).get().getValue();
```
jvdb: `Tie` guards all this -- if any result is `Empty`, or the amount of results that actually get turned into `DataExpressionSource` instances. It's possible to create all the error scenarios in tests and check them, but what's the point?

Still, it's quite inconsistent to check the `index` validity and not the emptiness. So touché :)

core/src/main/java/io/parsingdata/metal/token/Tie.java
```diff
@@ -0,0 +1,75 @@
...
+        super(name, encoding);
+        this.token = checkNotNull(token, "token");
+        this.dataExpression = checkNotNull(dataExpression, "dataExpression");
+    }
+
+    @Override
+    protected ParseResult parseImpl(final String scope, final Environment environment, final Encoding encoding) throws IOException {
+        final ImmutableList<OptionalValue> dataResult = dataExpression.eval(environment, encoding);
+        if (dataResult.isEmpty()) {
+            return failure(environment);
```
jvdb: Same as with `Sub`. The idea is that if you write down such a token, it should actually do something. If that depends on something, enclose it in a `Pre`. If it's OK to fail, enclose it in an `Opt`.

(I'm open to discussion, but this was the current thinking around `Sub`, which I've copied over to `Tie`.)

core/src/main/java/io/parsingdata/metal/token/Tie.java
```diff
@@ -0,0 +1,75 @@
...
+        }
+        final ParseResult result = iterate(scope, dataResult, 0, environment.addBranch(this), encoding);
+        if (result.succeeded) {
+            return success(new Environment(result.environment.closeBranch().order, environment.source, environment.offset, environment.callbacks));
```
jvdb: That doesn't reset `Environment.offset` and `Environment.callbacks`. You can make the case that the latter doesn't change, but the former is a problem I think.

PR #134, opened 2017-01-03
Gardening and documenting for the 5.0.0 release

### LINKED ISSUE #9: Document all public items
- [x] document `Token`s
- [ ] document `Shorthand`s
- [x] document `ComparisonExpression`s
- [x] document `LogicalExpression`s
- [x] document `ValueExpression`s
...

jvdb:
Resolves #9.


core/src/main/java/io/parsingdata/metal/token/Nod.java
```diff
@@ -29,6 +29,17 @@
 import io.parsingdata.metal.expression.value.OptionalValue;
 import io.parsingdata.metal.expression.value.ValueExpression;
 
+/**
+ * A {@link Token} that specifies an amount of data to skip in the input.
+ * <p>
+ * This tokens specifies a <code>size</code> (a {@link ValueExpression}) just
+ * like {@link Def} and evaluates this in the same way (failing when not a
+ * single value results). Unlike with Def, the data is not parsed but skipped
```
rdvdijk: Consider: `the data is not parsed` > `the data is not read`

core/src/main/java/io/parsingdata/metal/token/Opt.java
```diff
@@ -25,6 +25,13 @@
 import io.parsingdata.metal.data.ParseResult;
 import io.parsingdata.metal.encoding.Encoding;
 
+/**
+ * A {@link Token} that specifies an optional token.
+ *<p>
```
rdvdijk: Extreme nitpick: space before `<p>`.

core/src/main/java/io/parsingdata/metal/token/Rep.java
```diff
@@ -25,6 +25,14 @@
 import io.parsingdata.metal.data.ParseResult;
 import io.parsingdata.metal.encoding.Encoding;
 
+/**
+ * A {@link Token} that specifies a possible repetition of a token.
+ * <p>
+ * A Rep consists of a single <code>token</code>. A parse is attempted as long
+ * as parsing succeeds. As soon as a parse fails, the Rep succeeds.
```
rdvdijk: We can explicitly explain here that therefore a Rep will always succeed.

core/src/main/java/io/parsingdata/metal/token/Sub.java
```diff
@@ -32,6 +32,20 @@
 import io.parsingdata.metal.expression.value.OptionalValue;
 import io.parsingdata.metal.expression.value.ValueExpression;
 
+/**
+ * A {@link Token} that specifies a token to be parsed at a specific location
+ * in the input.
```
rdvdijk: `the input` could also be something like `the current source`, or something similar.

core/src/main/java/io/parsingdata/metal/token/Token.java
```diff
@@ -24,6 +24,25 @@
 import io.parsingdata.metal.data.ParseResult;
 import io.parsingdata.metal.encoding.Encoding;
 
+/**
+ * Base class for all Token implementations.
```
rdvdijk: What is a token?

core/src/main/java/io/parsingdata/metal/token/TokenRef.java
```diff
@@ -26,6 +26,14 @@
 import io.parsingdata.metal.data.ParseResult;
 import io.parsingdata.metal.encoding.Encoding;
 
+/**
+ * A {@link Token} that references a previously parsed token.
+ * <p>
+ * TokenRef consists of a <code>referenceName</code> (a String). In order to
+ * allow the construction of recursive tokens, this token can be used to
+ * reference an enclosing token and use it at the current location. An example
+ * for use of this token is to recursively define a linked list.
```
rdvdijk: Also explain how this token will be located.

core/pom.xml
```diff
@@ -2,7 +2,7 @@
   <modelVersion>4.0.0</modelVersion>
   <parent>
     <groupId>io.parsingdata</groupId>
-    <artifactId>metal-parent</artifactId>
+    <artifactId>metal</artifactId>
```
rdvdijk: Sneaky :wink: I think we won't have any troubles releasing this next version, but it's something to keep in mind if we do.

core/src/main/java/io/parsingdata/metal/expression/value/Value.java
```diff
@@ -55,10 +54,6 @@ public BitSet asBitSet() {
         return BitSet.valueOf(encoding.byteOrder == ByteOrder.BIG_ENDIAN ? ByteOrder.LITTLE_ENDIAN.apply(getValue()) : getValue());
     }
 
-    public OptionalValue operation(final ValueOperation operation) {
-        return operation.execute(this);
-    }
-
```
jvdb: This mechanism was needed to implement some operation on the data of a value. The `execute()` method would then handle evaluating, etc. and then pass the resulting `Value`. But that's what `ValueExpression` base classes like `UnaryValueExpression` and `BinaryValueExpression` also do, so no point in having this separate mechanism.

core/src/main/java/io/parsingdata/metal/token/Seq.java
```diff
@@ -27,6 +27,13 @@
 import io.parsingdata.metal.data.ParseResult;
 import io.parsingdata.metal.encoding.Encoding;
 
+/**
+ * A {@link Token} that specifies a dependency between a list of tokens.
```
jvdb: Agreed but it's a bit theoretical. I mean sure it's technically an AND, but most users won't care I think...

core/src/main/java/io/parsingdata/metal/expression/value/reference/CurrentOffset.java
```diff
@@ -25,7 +25,8 @@
 import io.parsingdata.metal.expression.value.ValueExpression;
 
 /**
- * Retrieve the current environment offset.
+ * A {@link ValueExpression} that represents the current offset in the parse
+ * state.
```
jvdb: Well, this is because I'm wondering whether we should explicitly discuss the names and shape of the `ParseGraph` and `Environment` APIs, or maybe let it be more conceptual at this level (and the documentation more durable).

core/src/main/java/io/parsingdata/metal/token/Sub.java
jvdb: Agreed. I prefer to call it input, but at some point we'll need to clarify somewhere that it can refer to different things...

core/pom.xml
jvdb: I especially want to make this change because when looking at Maven Central or searching for our projects, it is now named _metal-parent_ which is weird. It implies the "parent" part says something about what it does...

PR #143, opened 2017-01-20
Removed OptionalValue. Replaced all uses with Optional<Value>.

### LINKED ISSUE #124: Use Optional<T> instead of OptionalValue
Investigate whether this could be `Optional<T extends Value>` for cases where `T` is always `ParseValue`.

jvdb:
Resolves #124.

There are some cases where `ImmutableList<ParseValue>` is now `<ImmutableList<Value>`. We could investigate resolving this with something like `ImmutableList<T extends Value>` but for now I think it's not a problem, so let's only introduce extra complexity if there is a clear use case for the benefits.


jvdb:
I've turned `OptionalValue.wrap()` into `ImmutableList.wrap()` and made it generic (it can now wrap any `ImmutableList<T>` into an `ImmutableList<Optional<T>>`. It seems Better Code Hub feels that this means there are now too many calls going into `ImmutableList`.

core/src/main/java/io/parsingdata/metal/data/selection/ByName.java
```diff
@@ -52,15 +53,15 @@ public static ParseValue getValue(final ParseGraph graph, final String name) {
      * @param name Name of the value
      * @return All values with the provided name in this graph
      */
-    public static ImmutableList<ParseValue> getAllValues(final ParseGraph graph, final String name) {
+    public static ImmutableList<Value> getAllValues(final ParseGraph graph, final String name) {
```
rdvdijk: Why did this change from `ParseValue` to `Value`?

rdvdijk:
Somehow I cannot see the Better Code Hub results:

> Only the repository's owners and collaborators can view its Better Code Hub results.

jvdb:
I can't find where I can fix that. It's probably a bug since you're a repository owner :)

core/src/main/java/io/parsingdata/metal/data/selection/ByName.java
jvdb: There is an issue with the current definition of `ImmutableList.wrap()` that doesn't allow it. We could investigate an alternative declaration for either the `wrap()` method or `ValueExpression.eval()`.

PR #144, opened 2017-01-21
Use lambda expressions and method references

### LINKED ISSUE #31: Use lambda expressions
- Convert `Callback`.
- Convert `Reducer` on `Fold`s.

### LINKED ISSUE #138: Use Stream API
Investigate and try some potential low-hanging fruit around using Stream API.

### LINKED ISSUE #139: Use method references
Specifically for constructing default `Reducer` instances. Relates to #31.

jvdb:
Resolves #31. Resolves #139. Resolves #138.

The first two commits replace small interfaces with lambda expressions. The next two commits successively replace the *_REDUCER fields in `Shorthand` with method references and then inline them completely.

There is probably more possible in this area, but I think this covers the initial _conversion_ stage of introducing lambda expressions and default methods. We can do additional things as part of the regular development cycle.

The next commit converts one method to use the Stream API. The final commit updates the version to 6.0.0-SNAPSHOT because some of the changes are breaking.


core/src/main/java/io/parsingdata/metal/Util.java
```diff
@@ -51,13 +52,7 @@ private Util() {}
...
     public static String tokensToString(final Token[] tokens) {
         checkNotNull(tokens, "tokens");
-        if (tokens.length == 0) { return ""; }
-        final StringBuilder outString = new StringBuilder();
-        for (int i = 0; i < tokens.length - 1; i++) {
-            outString.append(tokens[i].toString());
-            outString.append(", ");
-        }
-        return outString.append(tokens[tokens.length - 1]).toString();
+        return Arrays.stream(tokens).map(t -> t.toString()).reduce((l, r) -> l + ", " + r).orElse("");
```
rdvdijk: Just wondering, is Java 8 "smart enough" to use a `StringBuilder` internally? I think in certain situations the compiler does optimize string concatenation like this.

jvdb: I wondered about that too... But this code is not in any inner loop so we can worry about it later.

PR #145, opened 2017-01-22
Migrate to Java 8

### LINKED ISSUE #31: Use lambda expressions

### LINKED ISSUE #124: Use Optional<T> instead of OptionalValue

### LINKED ISSUE #138: Use Stream API

### LINKED ISSUE #139: Use method references

### LINKED ISSUE #140: Use default methods
Specifically for the `ParseItem` methods such as `isGraph()`, etc.

jvdb:
Resolves #140. Resolves #124. Resolves #31. Resolves #139. Resolves #138.

This should conclude work on migrating Metal to Java 8. This does not mean that this exhausts all opportunities for using Java 8-related constructs in Metal, but this simply removes all workarounds (e.g., `OptionalValue` to `Optional<Value>`) and applies new approaches where they are a clear one-to-one replacement (e.g., `Callback` to `BiConsumer<Token, ParseResult>`).

Things that remain can be included in the regular continuing development cycle.


jvdb:
Also, note to @rdvdijk: the only addition to this PR with regard to earlier PRs is the final commit, which is based on your suggestion :)

jvdb:
One additional commit added: moved `ImmutableList.reverse()` to (static) `Reversal.reverse()` and `ImmutableList.wrap()` to `Wrapping.wrap()`.

PR #146, opened 2017-01-25
Removed all uses of instanceof.

### LINKED ISSUE #63: Remove uses of instanceof

jvdb:
The only remaining `instanceof` was in `Offset`, where a `Value` was checked to be a `ParseValue` before retrieving its `offset`. But since 5.0.0, every `Value` has a `.slice.offset`, so this check (and the accompanying cast can be removed).

Resolves #63.

### DIFF
--- a/core/src/main/java/io/parsingdata/metal/expression/value/reference/Offset.java
+++ b/core/src/main/java/io/parsingdata/metal/expression/value/reference/Offset.java
@@ -40,10 +40,7 @@ public class Offset extends UnaryValueExpression {
 
     @Override
     public Optional<Value> eval(final Value value, final Environment environment, final Encoding encoding) {
-        if (value instanceof ParseValue) {
-            return Optional.of(ConstantFactory.createFromNumeric(((ParseValue) value).slice.offset, value.encoding));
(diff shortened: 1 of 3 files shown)


core/src/main/java/io/parsingdata/metal/expression/value/reference/Offset.java
```diff
@@ -40,10 +40,7 @@
 
     @Override
     public Optional<Value> eval(final Value value, final Environment environment, final Encoding encoding) {
-        if (value instanceof ParseValue) {
-            return Optional.of(ConstantFactory.createFromNumeric(((ParseValue) value).slice.offset, value.encoding));
-        }
-        return Optional.empty();
+        return Optional.of(ConstantFactory.createFromNumeric(value.slice.offset, value.encoding));
```
rdvdijk: Whoa. :open_mouth:

jvdb: I had the same reaction! So we lose the `instanceof`, the cast _and_ the `empty()`! :)

PR #149, opened 2017-02-09
Implement custom equality for Source comparison

### LINKED ISSUE #130: Implement equality checks for data structures

### LINKED ISSUE #147: Sub uses incorrect Source for cycle detection
Instead of using the `Source` from the `Environment`, it takes it from the evaluated address expression. These tend to be the same (especially when using `Ref` which is the base case for `Sub`), but not if you `Sub` to a fixed address, because then the source will be a `ConstantSource` created by evaluating the `Con`.

### LINKED ISSUE #148: Change interface of ValueExpression to depend on ParseGraph instead of Environment
Custom equality checks (#130) also touch `Environment` through `DataExpression` which is undesirable because that includes fields (`offset`, `Callbacks` and the default `Source`) that should not affect equality when comparing `Token`, `ValueExpression` or `Expression` instances.

jvdb:
Resolves #130: Correct cycle detection for `Sub` tokens depends on `Source` equality. Implementing `equals()` and `hashCode()` correctly for `Source` means also implementing it for all `Token`s, `ValueExpression`s, `Expression`s, `Encoding`, `ParseItem`s and all concrete `Source`s. This PR completes this operation.

Resolves #148: As part of this, the interface to `ValueExpression` has changed to no longer require an `Environment` as input to its `eval()` method, but just the `ParseGraph`. This to make sure objects not related to the parse state are not inadvertently used to determine equality (e.g., `Callbacks`).

Resolves #147: A bug was also fixed where `Sub` used the `Source` of its address `ValueExpression` result as input instead of the input from the `Environment`.

Additionally, `CurrentOffset` has been removed, as it depended on a field of the `Environment` being passed (`offset`). It has been replaced by a oneliner in `Shorthand` that calculates it based on the `ParseGraph`. The new implementation is not strictly the same but should be equivalent in most cases.

`ConstantSlice` has also been removed. Its inner class `ConstantSource` has been promoted to a top-level class and the regular `Slice` implementation can be used to create a slice of a `ConstantSource`.

Testing of both `equals()` and `hashCode()` is done using two new test classes. `AutoEqualityTest` is an automated, combinatorial approach for most classes. Custom testing of some specific cases is in `EqualityTest`, along with a specific cycle detection case for `Sub`.


rdvdijk:
I still cannot see the Better Code Hub results, what is it complaining about?

> This code needs to be refactored

core/src/main/java/io/parsingdata/metal/expression/value/reference/Last.java
```diff
@@ -49,4 +50,16 @@ public String toString() {
...
+    @Override
+    public boolean equals(Object obj) {
+        return obj != null
+            && getClass() == obj.getClass()
+            && Objects.equals(operand, ((Last)obj).operand);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(operand);
```
rdvdijk: A few of these `hashCode()` implementations are exactly the same. For example: `First` and `Last`. Could that pose a problem?

core/src/main/java/io/parsingdata/metal/expression/value/reference/Self.java
```diff
@@ -32,13 +32,24 @@
...
 
+    @Override
+    public boolean equals(Object obj) {
+        return obj != null
+            && getClass() == obj.getClass();
+    }
+
+    @Override
+    public int hashCode() {
+        return 2;
```
rdvdijk: :two: ?

Just wondering, why 2? 😉

core/src/test/java/io/parsingdata/metal/AutoEqualityTest.java
```diff
@@ -0,0 +1,283 @@
...
+            instances[1] = cons.newInstance(argLists.get(0).toArray());
+            List<Object> otherInstances = new ArrayList<>();
+            for (List<Object> argList : argLists.subList(1, argLists.size())) {
+                otherInstances.add(cons.newInstance(argList.toArray()));
+            }
+            instances[2] = otherInstances.toArray();
+        } catch (Exception e) {
+            e.printStackTrace();
+        }
+        return instances;
```
rdvdijk: And as a side note, you could use the streaming API to replace some of the "old school" for loops in this code.

(But not the one right here, since `newInstance` throws an `Exception`.)

core/src/main/java/io/parsingdata/metal/data/ConstantSource.java
```diff
@@ -0,0 +1,59 @@
...
+import static io.parsingdata.metal.Util.checkNotNull;
+
+import java.io.IOException;
+import java.util.Arrays;
+
+public class ConstantSource extends Source {
+
+    private final byte[] data; // Private because array content is mutable.
+
+    public ConstantSource(byte[] data) {
```
rdvdijk: Quite a few missing `final` statements throughout this new code (also in all new `equals` methods).

*Hint*: you can configure the IntelliJ "inspections" to mark missing `final`s as **error**. That way you will never miss one again.

core/src/main/java/io/parsingdata/metal/expression/value/reference/Self.java
ccreeten: The only even prime? :smile:

core/src/main/java/io/parsingdata/metal/expression/value/ConstantFactory.java
```diff
@@ -29,8 +29,12 @@
 
     private ConstantFactory() {}
 
+    public static Slice makeConstantSlice(final byte[] data) {
```
jvdb: Good idea, I've renamed it to `createFromBytes()`. Additionally, I figured that `ConstantFactory` actually exclusively produces `Value` instances, so I moved it to `Util` instead.

core/src/test/java/io/parsingdata/metal/AutoEqualityTest.java
```diff
@@ -0,0 +1,283 @@
...
+import io.parsingdata.metal.token.Pre;
+import io.parsingdata.metal.token.Rep;
+import io.parsingdata.metal.token.RepN;
+import io.parsingdata.metal.token.Seq;
+import io.parsingdata.metal.token.Sub;
+import io.parsingdata.metal.token.Tie;
+import io.parsingdata.metal.token.Token;
+import io.parsingdata.metal.token.TokenRef;
+import io.parsingdata.metal.token.While;
+import io.parsingdata.metal.util.InMemoryByteStream;
```
jvdb: With this types of classes I would suggest not to use imports, but the lines are already long as it is...

core/src/test/java/io/parsingdata/metal/AutoEqualityTest.java
```diff
@@ -0,0 +1,283 @@
...
+        try {
+            instances[0] = cons.newInstance(argLists.get(0).toArray());
+            instances[1] = cons.newInstance(argLists.get(0).toArray());
+            List<Object> otherInstances = new ArrayList<>();
+            for (List<Object> argList : argLists.subList(1, argLists.size())) {
+                otherInstances.add(cons.newInstance(argList.toArray()));
+            }
+            instances[2] = otherInstances.toArray();
+        } catch (Exception e) {
+            e.printStackTrace();
```
jvdb: I didn't because of your following comment, as I used some Stream API stuff above it before... Fixed now!

jvdb: I didn't because of your following comment, as I used some Stream API stuff above it before... Fixed now!

core/src/test/java/io/parsingdata/metal/AutoEqualityTest.java
jvdb: Used your code fully. Only removed the extra newline :-)

jvdb: (RE: Stream API, absolutely, but no hurry there.)

core/src/test/java/io/parsingdata/metal/AutoEqualityTest.java
```diff
@@ -0,0 +1,283 @@
...
+        return result;
+    }
+
+    public static Object OTHER_TYPE = new Object() {};
+
+    private final Object object;
+    private final Object same;
+    private final Object[] other;
+
+    public AutoEqualityTest(final Object object, final Object same, final Object[] other) throws NoSuchMethodException {
```
jvdb: Wow... This has lead me to also remove a huge amount of constructors from all `ParameterizedParse` subclasses :)

core/src/main/java/io/parsingdata/metal/expression/value/arithmetic/Mul.java
```diff
@@ -35,7 +35,7 @@ public Mul(final ValueExpression left, final ValueExpression right) {
     }
 
     @Override
-    public Optional<Value> eval(final Value left, final Value right, final Environment environment, final Encoding encoding) {
+    public Optional<Value> eval(final Value left, final Value right, final ParseGraph graph, final Encoding encoding) {
         return Optional.of(ConstantFactory.createFromNumeric(left.asNumeric().multiply(right.asNumeric()), encoding));
     }
```
jvdb: Correct! I would like to discuss this a bit further. There is some low-hanging fruit, but if we go this path we need to go all-in and I want to make sure we understand what that entails.

core/src/main/java/io/parsingdata/metal/data/ByteStreamSource.java
```diff
@@ -43,4 +44,16 @@ public String toString() {
         return input.toString();
     }
 
+    @Override
+    public boolean equals(Object obj) {
+        return obj != null
+            && getClass() == obj.getClass()
```
jvdb: This was actually the main complaint of BetterCodeHub :)

core/src/main/java/io/parsingdata/metal/data/ParseGraph.java
```diff
@@ -118,4 +119,20 @@ public String toString() {
...
     }
 
+    @Override
+    public boolean equals(Object obj) {
+        return obj != null
+            && getClass() == obj.getClass()
+            && Objects.equals(head, ((ParseGraph)obj).head)
+            && Objects.equals(tail, ((ParseGraph)obj).tail)
+            && Objects.equals(branched, ((ParseGraph)obj).branched)
+            && Objects.equals(definition, ((ParseGraph)obj).definition);
```
jvdb: I see your point but I'm not convinced. I propose we keep this for now and see if we want to change it later.

core/src/main/java/io/parsingdata/metal/expression/True.java
```diff
@@ -26,11 +28,22 @@
...
 
+    @Override
+    public boolean equals(Object obj) {
+        return obj != null
+            && getClass() == obj.getClass();
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(true);
```
jvdb: Yeah, but semantically that makes sense right? I mean if it has to collide with _something_...

core/src/main/java/io/parsingdata/metal/expression/value/reference/Self.java
jvdb: I actually think `CurrentOffset` had 1, but since it was removed, now all that remains is 2.

(Remember, always two there are...)

jvdb:
Massive improvements! Thanks for the great feedback @rdvdijk!

core/src/main/java/io/parsingdata/metal/expression/value/reference/Self.java
jvdb: Anyway, 2 seems to be a fine number!

core/src/main/java/io/parsingdata/metal/expression/value/reference/Last.java
rdvdijk: How expensive is `getClass().hashCode()`? That's one we could add.

core/src/main/java/io/parsingdata/metal/expression/True.java
rdvdijk: Yeah, seems fine.

rdvdijk:
Approved, pending the Travis build :computer:

core/src/main/java/io/parsingdata/metal/expression/value/reference/Last.java
jvdb: It seems that `getClass()` is cheap, so that might be a good idea. We could even cache it statically. I've made an issue: #150.

PR #152, opened 2017-02-10
Removed ParseResult, replaced it with Optional<Environment>

### LINKED ISSUE #142: Use Optional<Environment> instead of ParseResult
The assumption here is that when the result is false, that the Environment can be omitted. This needs to be investigated before we decide either way.

jvdb:
Resolves #142.

Fairly mechanical global replacement:
- `ParseResult` -> `Optional<Environment>`
- `.succeeded` -> `.isPresent()`
- `.environment` -> `.get()`

Additionally, since `success()` and `failure()` were in `ParseResult`, they were moved to `Util`. It's a point of discussion whether they should exist after this change, since they are essentially calls to `Optional.of()` and `Optional.empty()`. I personally think that reading `success()` or `failure()` is more readable, so I prefer to keep them for the time being.

Biggest functional change is in `BaseCallback`, where `handleFailure()` doesn't get an `Environment` passed in anymore, since it's not available on failure. This is actually better since it was ambiguous _which_ `Environment` you would receive in the old scenario.

Finally, when converting the callback handling, I found an `==` compare on `Token` instances in `Callbacks.handleCallbacks()` which is now replaced with a call to `equals()`! (#130)


jvdb:
Bettercodehub is complaining again, but a fair point again: `Util` is slowly turning into a place that too many things depend on.

rdvdijk:
> Biggest functional change is in `BaseCallback`, where `handleFailure()` doesn't get an `Environment` passed in anymore, since it's not available on failure. This is actually better since it was ambiguous which `Environment` you would receive in the old scenario.

I don't really like this change in functionality.

A common complaint that drove us to add callbacks is: "Why did my parse fail?".

We added the callbacks to handle not only the "happy case" of successful parses of tokens, but also to analyse the case of parsing that fails. If we do not pass an `Environment` into `handleFailure()`, there is no way of knowing why something went wrong. I think it's very useful to pass the `Environment` that was used when the parsing failed; for example a developer can evaluate an expression on it, to see why it failed (for example, why did a `ref` fail?).

jvdb:
But then you oppose the entire change? Because there's nothing to pass if we make this change to the API...

rdvdijk:
> But then you oppose the entire change?

I haven't rejected the PR, right? 😁 

Haven't taken the time yet to look at the entire change to see if I oppose *all* of it 😉 Overall I like the idea of removing `ParseResult`. I just wanted to reply to that detail of your description.

rdvdijk:
`Token` has now changed to:

```java
    public Optional<Environment> parse(final String scope, final Environment environment, final Encoding encoding) throws IOException {
        final Encoding activeEncoding = this.encoding != null ? this.encoding : encoding;
        final Optional<Environment> result = parseImpl(makeScope(scope), environment, activeEncoding);
        environment.callbacks.handle(this, result);
        return result;
    }
```

If we do want to pass an `Environment` to callbacks in case of a failed parse, we could just pass in the `environment` variable, i.e. the environment that we had *before* the failed parse.

That does mean that we cannot let the `Callbacks` determine when a parse failed or succeeded, but we have to do it here. So, our code would change to something like:

```java
        ...
        final Optional<Environment> result = parseImpl(makeScope(scope), environment, activeEncoding);
        if (result.isPresent()) {
            environment.callbacks.success(this, result.get());
        } else {
            environment.callbacks.failure(this, environment);
        }

```

Not too bad, right? Especially if we move this code into a separate method.

jvdb:
Let's see what I can make of it :-)

jvdb:
Note by the way that `BaseCallback` is just an optional base class. The callback mechanism itself doesn't distinguish between succeeded/failed parses and just passes along the return value. But essentially that has the same result as removing it from the `BaseCallback`'s failure handler: no access.

I agree that passing the original `Environment` may be a solution, especially since that is less ambiguous than the original approach, where each `Token` could determine which `Environment` to return upon failure.

We could also change the overall callback interface to always pass through the before _and_ after `Environment`. I mean if we want to make debugging easy...

jvdb:
Ok so the `BiConsumer<Token, Optional<Environment>` is gone and the `Callback` is now back. Which is actually nice because it makes way more sense to implement the latter than the former. Along with the better method naming (`handle()` instead of `accept()`).

The functional interface is then: `Callback.handle(Token token, Environment before, Optional<Environment> after)` and `BaseCallback` turns that into: `handleSuccess(Token token, Environment before, Environment after)` and `handleFailure(Token token, Environment before)`.

Very nice! This also allowed me to reinstate a check in the unit tests (`CallbackTest` line 224).

core/src/main/java/io/parsingdata/metal/token/Tie.java
```diff
@@ -57,28 +56,28 @@ public Tie(final String name, final Token token, final ValueExpression dataExpre
...
         if (dataResult.isEmpty()) {
-            return failure(environment);
+            return failure();
         }
-        final ParseResult result = iterate(scope, dataResult, 0, environment.addBranch(this), encoding);
-        if (result.succeeded) {
-            return success(new Environment(result.environment.closeBranch().order, environment.source, environment.offset, environment.callbacks));
+        final Optional<Environment> result = iterate(scope, dataResult, 0, environment.addBranch(this), encoding);
+        if (result.isPresent()) {
+            return success(new Environment(result.get().closeBranch().order, environment.source, environment.offset, environment.callbacks));
```
rdvdijk: ~~Can we just pass `result.get().closeBranch()`, here instead of reconstructing an entire `new Environment`?~~

Nevermind. But could we do `result.get().closeBranch().source(environment.source).seek(environment.offset)`?

core/src/main/java/io/parsingdata/metal/data/callback/BaseCallback.java
```diff
@@ -16,24 +16,23 @@
...
         } else {
-            handleFailure(token, result.environment);
+            handleFailure(token, before);
         }
     }
 
-    protected abstract void handleSuccess(Token token, Environment environment);
-    protected abstract void handleFailure(Token token, Environment environment);
+    protected abstract void handleSuccess(Token token, Environment before, Environment after);
+    protected abstract void handleFailure(Token token, Environment before);
```
rdvdijk: Use default (empty) methods here? I noticed that in our tests we have a case where we don't implement `handleFailure`.

PR #155, opened 2017-02-14
Converted Seq/Cho tokens to use ImmutableList<Token> instead of Token[]

### LINKED ISSUE #153: Convert Seq/Cho tokens to use ImmutableList<Token> instead of Token[]
Since we're iterating recursively, it actually works nicely to use a recursive, immutable data structure. This will also have good synergy with the trampoline approach to improve recursive iteration.

Additionally, the `token` field can then be made public instead of hidden behind a getter (and `clone()`d upon retrieval).

This does however mean that `ImmutableList<T>` enters visibility from the perspective of `Source` equality, so it will have to implements `equals()` and `hashCode()`.

jvdb:
Resolves #153.

This allows the field to be public, like (almost) all other fields in Metal. It works well with the recursive implementation and will hopefully work equally well with trampolines.

Added `equals()` and `hashCode()` methods to `ImmutableList<T>` because adding it to tokens means it's now visible from the perspective of `Source` equality (used by `Sub` to detect cycles).


core/src/main/java/io/parsingdata/metal/data/transformation/Array.java
```diff
@@ -0,0 +1,36 @@
...
+
+import static io.parsingdata.metal.Util.checkNotNull;
+
+import io.parsingdata.metal.data.ImmutableList;
+
+public final class Array {
+
+    private Array() {}
+
+    public static <T> ImmutableList<T> toList(final T[] array) {
```
rdvdijk: Make this part of `ImmutableList`? (`ImmutableList.fromArray(...)`?)

jvdb: Or maybe even `create(final T[] array)`?

core/src/test/java/io/parsingdata/metal/EqualityTest.java
```diff
@@ -136,4 +137,17 @@ public void multiConstructorTypes() {
         }
     }
 
+    @Test
+    public void immutableList() {
+        final ImmutableList<String> object = ImmutableList.create("a");
```
jvdb: I use the terms object, same and other in the other test code around equality, so here I'm striving for consistency, but in another dimension :)

PR #156, opened 2017-02-16
Adds Shorthands for map function (mapLeft/mapRight) using Expand ValueExpression

### LINKED ISSUE #113: Implement map method to apply some operation to all members of a list
For example, when dealing with a list of offsets that all need to be corrected with some base value, the list of addresses passed to `Sub` needs to be modified by a single `ValueExpression`.

jvdb:
Resolves #113.

Adds `mapLeft()` and `mapRight()` to `Shorthand` using the new `Expand` `ValueExpression` to allow _map_-functionality for `BinaryValueExpression`s.

Part of this PR should also be commit 645f9b40fc4884b6101c1932adbd240c73bb4ae6, which was accidentally committed directly to the master.

### DIFF
--- a/core/src/main/java/io/parsingdata/metal/Shorthand.java
+++ b/core/src/main/java/io/parsingdata/metal/Shorthand.java
@@ -171,6 +171,8 @@ public final class Shorthand {
     public static ValueExpression fold(final ValueExpression values, final BinaryOperator<ValueExpression> reducer, final ValueExpression initial) { return foldRight(values, reducer, initial); }
     public static ValueExpression rev(final ValueExpression values) { return new Reverse(values); }
     public static ValueExpression exp(final ValueExpression base, final ValueExpression count) { return new Expand(base, count); }
+    public static BinaryValueExpression mapLeft(final BiFunction<ValueExpression, ValueExpression, BinaryValueExpression> func, final ValueExpression left, final ValueExpression rightExpand) { return func.apply(left, exp(rightExpand, count(left))); }
+    public static BinaryValueExpression mapRight(final BiFunction<ValueExpression, ValueExpression, BinaryValueExpression> func, final ValueExpression leftExpand, final ValueExpression right) { return func.apply(exp(leftExpand, count(right)), right); }
 
     public static BinaryLogicalExpression and(final Expression left, final Expression right) { return new And(left, right); }
     public static BinaryLogicalExpression or(final Expression left, final Expression right) { return new Or(left, right); }
--- a/core/src/test/java/io/parsingdata/metal/ShorthandsTest.java
+++ b/core/src/test/java/io/parsingdata/metal/ShorthandsTest.java
@@ -30,6 +30,8 @@ import static io.parsingdata.metal.Shorthand.eqStr;
 import static io.parsingdata.metal.Shorthand.expTrue;
 import static io.parsingdata.metal.Shorthand.gtNum;
 import static io.parsingdata.metal.Shorthand.last;
+import static io.parsingdata.metal.Shorthand.mapLeft;
+import static io.parsingdata.metal.Shorthand.mapRight;
 import static io.parsingdata.metal.Shorthand.opt;
 import static io.parsingdata.metal.Shorthand.pre;
 import static io.parsingdata.metal.Shorthand.ref;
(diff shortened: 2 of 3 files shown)


core/src/test/java/io/parsingdata/metal/expression/value/ExpandTest.java
```diff
@@ -99,7 +99,7 @@ public void expandList() {
         }
     }
 
-    private static ParseValue createParseValue(String name, int value) {
+    public static ParseValue createParseValue(String name, int value) {
```
ccreeten: Can this perhaps be moved, as it is used across multiple classes, and not really specifically belongs to this class?

core/src/test/java/io/parsingdata/metal/ShorthandsTest.java
```diff
@@ -214,4 +218,28 @@ public void checkSeqTokens() {
         assertEquals(DEFB, seq.tokens.tail.head);
     }
 
+    final ParseGraph PARSEGRAPH = ParseGraph.EMPTY.add(createParseValue("a", 126)).add(createParseValue("a", 84)).add(createParseValue("a", 42));
+
+    @Test
+    public void mapLeftWithSub() {
```
ccreeten: Nice! Question: What if the operand to expand has more than one value?

Say I would add two ParseValues both name ```b``` with value 1 and 2 to the graph. If I would then call:
```java
mapLeft(Shorthand::sub, ref("a"), ref("b"))
```
The expand would then expand the right operand to a list of size 6, with alternating values of 1 and 2. The result of sub would then be a list of size 6, with three empties at the end. Would performing the operation on the cartesian product instead make any sense/be useful?

jvdb: Possibly, but we decided to not go for carthesian product in general `BinaryValueExpression` handling because the result would be too difficult to navigate declaratively. I think that will be the case here as well.

But I do agree that expanding a multiple value list breaks the concept of map (because it doesn't map everything in tjat situation...)

core/src/main/java/io/parsingdata/metal/Shorthand.java
```diff
@@ -171,6 +171,8 @@ private Shorthand() {}
     public static ValueExpression fold(final ValueExpression values, final BinaryOperator<ValueExpression> reducer, final ValueExpression initial) { return foldRight(values, reducer, initial); }
     public static ValueExpression rev(final ValueExpression values) { return new Reverse(values); }
     public static ValueExpression exp(final ValueExpression base, final ValueExpression count) { return new Expand(base, count); }
+    public static BinaryValueExpression mapLeft(final BiFunction<ValueExpression, ValueExpression, BinaryValueExpression> func, final ValueExpression left, final ValueExpression rightExpand) { return func.apply(left, exp(rightExpand, count(left))); }
+    public static BinaryValueExpression mapRight(final BiFunction<ValueExpression, ValueExpression, BinaryValueExpression> func, final ValueExpression leftExpand, final ValueExpression right) { return func.apply(exp(leftExpand, count(right)), right); }
```
rdvdijk: We need to think about how to document 'composed' shorthands like this. Most other shorthands are documented in the `Token`s they return, but we cannot do that here.

core/src/main/java/io/parsingdata/metal/Shorthand.java
```diff
@@ -171,6 +171,8 @@ private Shorthand() {}
     public static ValueExpression fold(final ValueExpression values, final BinaryOperator<ValueExpression> reducer, final ValueExpression initial) { return foldRight(values, reducer, initial); }
     public static ValueExpression rev(final ValueExpression values) { return new Reverse(values); }
     public static ValueExpression exp(final ValueExpression base, final ValueExpression count) { return new Expand(base, count); }
```
ccreeten: It can be seen in the commit mentioned in the pr description.

rdvdijk: It was joke 😜

ccreeten: Urgh, apparently if I open a link in my mail app, it opens inside that app, in which these emoticons are not shown. :cry:

core/src/test/java/io/parsingdata/metal/expression/value/ExpandTest.java
jvdb: I tend to not pay much attention to these things in test code, but you have a point. I'll make an issue so collect all these _dummy_ values in a static `TestData` class or something like that.

core/src/main/java/io/parsingdata/metal/Shorthand.java
jvdb: Agreed. In a way they are "just" macros, but even so they represent some chunked knowledge on a higher level of abstraction than their parts. Maybe in the guide that is still planned?

Issue #157, opened 2017-02-23
Generalize pre and postcondition specification


jvdb:
Currently, Metal supports specifying preconditions using the `Pre` `Token`, which acts more like an _if_, because a `false` precondition there does not lead to failure (only a `true` precondition and a subsequently failing parse does that).

Postconditions are only specified on `Def` `Token`s, which often leads to difficult placement of postconditions on larger blocks (e.g., this leads to "put this as a postcondition on the first `Def` after that block".)

A more general approach would be:
- [x] Turn `Pre` into a real precondition, so that its predicate must be `true` _and_ have a successful parse of the associated `Token` in order to succeed itself. This may lead to some issues around current behaviour which must be investigated.
- [x] Add a `Post` `Token` which does the same, but first executes the parse on its associated `Token`.
- [x] Remove the predicate argument/field from `Def`, since that can be expressed nesting a `Def` in a `Post`.
- [x] Change `Def` behaviour around zero-size values. They are allowed, but should not lead to a `ParseValue` being added to the `ParseGraph`.
- [x] Introduce `Shorthand` for `empty`: a zero-size `Def` without a postcondition. Doesn't parse anything but always succeeds.
- [x] Consider removing `Opt`. Can be replaced in `Shorthand` by `cho(argument, empty)`.

PR #160, opened 2017-03-02
Generalize pre- and postcondition checking and adapt Def

jvdb:
See the description of issue #157 for a detailed overview: each of the six commits performs one of the items detailed there.


core/src/test/java/io/parsingdata/metal/token/PostTest.java
```diff
@@ -0,0 +1,94 @@
...
+    public void postconditionParseFails() throws IOException {
+        final Optional<Environment> result = SEQUENCE.parse(stream(1, 2), enc());
+
+        // parse fails
+        assertThat(result.isPresent(), is(false));
+    }
+
+    @Test
+    public void postconditionNull() throws IOException {
+        final Token noPostcondition = post(def("value", 1, eq(con(0))), null);
```
jvdb: I agree! It made sense to have the predicate be optional when it was still part of `Def`, but for both `Pre` and `Post` they should be mandatory (so I will change it to a hard NotNull-check).

core/src/test/java/io/parsingdata/metal/TreeTest.java
```diff
@@ -57,10 +56,16 @@
         seq("tree",
             def("head", con(1), eq(con(HEAD))),
             def("nr", con(1)),
-            def("left", con(1)),
-            pre(sub(token("tree"), last(ref("left"))), not(eq(last(ref("left")), con(0)))),
-            def("right", con(1)),
-            pre(sub(token("tree"), last(ref("right"))), not(eq(last(ref("right")), con(0))))
+            cho(def("left_terminator", con(1), eq(con(0))),
```
rdvdijk: Just wondering: why did you rewrite this token? (I do not disagree with this new one, by the way.)

core/src/test/java/io/parsingdata/metal/data/callback/CallbackTest.java
```diff
@@ -154,9 +159,9 @@ protected void handleFailure(Token token, Environment before) {}
 
     @Test
     public void genericCallback() throws IOException {
-        final Deque<Token> expectedSuccessDefinitions = new ArrayDeque<>(Arrays.asList(ONE, TWO, ONE, TWO, FOUR, SEQ124, CHOICE));
-        final Deque<Long> expectedSuccessOffsets = new ArrayDeque<>(Arrays.asList(1L, 2L, 1L, 2L, 3L, 3L, 3L));
-        final Deque<Token> expectedFailureDefinitions = new ArrayDeque<>(Arrays.asList(THREE, SEQ123));
+        final Deque<Token> expectedSuccessDefinitions = new ArrayDeque<>(Arrays.asList(DEF_ONE, POST_ONE, DEF_TWO, POST_TWO, DEF_THREE, DEF_ONE, POST_ONE, DEF_TWO, POST_TWO, DEF_FOUR, POST_FOUR, SEQ124, CHOICE));
+        final Deque<Long> expectedSuccessOffsets = new ArrayDeque<>(Arrays.asList(1L, 1L, 2L, 2L, 3L, 1L, 1L, 2L, 2L, 3L, 3L, 3L, 3L));
+        final Deque<Token> expectedFailureDefinitions = new ArrayDeque<>(Arrays.asList(POST_THREE, SEQ123));
```
rdvdijk: These lists of expected success and failure definitions are nice, it illustrates what happens during parsing.

We *could* make it extremely explicit and create a list/chain of expectations that combines both success/failure and the offset where those occur, to show when during the parse run.

Something like:

```java
expectSuccess(DEF_ONE, 1), // token, offset
expectSuccess(POST_ONE, 1), 
expectSuccess(DEF_TWO, 2), 
expectSuccess(POST_TWO, 2), 
expectSuccess(DEF_THREE, 3), 
expectFailure(POST_THREE, 3), 
expectFailure(SEQ123, 3),
expectSuccess(DEF_ONE, 1), 
expectSuccess(POST_ONE, 1), 
expectSuccess(DEF_TWO, 2), 
expectSuccess(POST_TWO, 2), 
expectSuccess(DEF_FOUR, 3), 
expectSuccess(POST_FOUR, 3), 
expectSuccess(SEQ124, 3), 
expectSuccess(CHOICE, 3);
```

Food for thought.

core/src/test/java/io/parsingdata/metal/token/PostTest.java
```diff
@@ -0,0 +1,79 @@
...
+    private static final Token SEQUENCE =
+        post(seq(any("header"),
+                 def("value", 1, eq(con(1)))),
+                 eq(ref("header"), con(1)));
+
+    @Test
+    public void postconditionTrue() throws IOException {
+        final Optional<Environment> result = SEQUENCE.parse(stream(1, 1), enc());
+
+        // token parses and postcondition is true
```
rdvdijk: (I was about to comment: "Hey, Dutchie! In English it's 'post condition' (two words)." But then I Googled and was proven wrong.)

core/src/test/java/io/parsingdata/metal/token/WhileTest.java
```diff
@@ -77,7 +77,7 @@ public void whileWithoutExpression() throws IOException {
     @Test
     public void testToString() {
         final Token simpleWhile = whl(def("value", con(1)), eq(con(1)));
-        final String simpleWhileString = "While(Def(value,Const(0x01),True), Eq(Const(0x01)))";
+        final String simpleWhileString = "While(Def(value,Const(0x01)), Eq(Const(0x01)))";
```
rdvdijk: Subtle change, but *much* better.

core/src/test/java/io/parsingdata/metal/SubStructTest.java
```diff
@@ -133,15 +133,15 @@ public void linkedListWithCycleToLowerOffset() throws IOException {
     }
 
     private void checkBranch(final ParseGraph graph, final int graphOffset, final int nextOffset) {
-        checkValue(graph.head, 1, graphOffset + 2); // footer
+        checkValue(graph.head.asGraph().head, 1, graphOffset + 2); // footer
         checkValue(graph.tail.tail.head, nextOffset, graphOffset + 1); // next
-        checkValue(graph.tail.tail.tail.head, 0, graphOffset); // header
+        checkValue(graph.tail.tail.tail.head.asGraph().head, 0, graphOffset); // header
```
rdvdijk: These `graph.tail.tail.tail.head` tests are very hard to review :stuck_out_tongue_closed_eyes:

core/src/test/java/io/parsingdata/metal/TreeTest.java
rdvdijk: The reason I'm asking is that in practice I would be more likely to write the original style (`def+pre(sub)`), that is: "read pointer, sub elsewhere based on the pointer's value".

core/src/main/java/io/parsingdata/metal/Shorthand.java
```diff
@@ -81,14 +81,15 @@
...
-    public static Token def(final String name, final ValueExpression size, final Encoding encoding) { return def(name, size, null, encoding); }
+    public static Token def(final String name, final ValueExpression size, final Encoding encoding) { return new Def(name, size, encoding); }
     public static Token def(final String name, final ValueExpression size) { return def(name, size, (Encoding)null); }
     public static Token def(final String name, final long size, final Expression predicate, final Encoding encoding) { return def(name, con(size), predicate, encoding); }
     public static Token def(final String name, final long size, final Expression predicate) { return def(name, size, predicate, null); }
-    public static Token def(final String name, final long size, final Encoding encoding) { return def(name, size, null, encoding); }
+    public static Token def(final String name, final long size, final Encoding encoding) { return def(name, con(size), encoding); }
     public static Token def(final String name, final long size) { return def(name, size, (Encoding)null); }
+    public static final Token empty = def(NO_NAME, 0L);
```
jvdb: Good points. I will make a separate issue for this: #163.

core/src/main/java/io/parsingdata/metal/token/Def.java
```diff
@@ -29,35 +29,27 @@
...
- * A Def consists of a <code>size</code> (a {@link ValueExpression}, in bytes)
- * and a <code>predicate</code> (an {@link Expression}). The
- * <code>predicate</code> may be <code>null</code>.
+ * A Def consists of a <code>size</code> (a {@link ValueExpression}.
  * <p>
- * Parsing will succeed if <code>size</code> evaluates to a single value, if
- * that many bytes are available in the input and if <code>predicate</code>
- * (if present) evaluates to <code>true</code>.
+ * Parsing will succeed if <code>size</code> evaluates to a single value and if
+ * that many bytes are available in the input.
```
jvdb: There are _always_ 0 bytes available right? But I agree it needs to be mentioned explicitly.

core/src/main/java/io/parsingdata/metal/token/Post.java
```diff
@@ -25,46 +26,55 @@
...
         this.token = checkNotNull(token, "token");
+        this.predicate = checkNotNull(predicate, "predicate");
     }
 
     @Override
     protected Optional<Environment> parseImpl(final String scope, final Environment environment, final Encoding encoding) throws IOException {
         final Optional<Environment> result = token.parse(scope, environment.addBranch(this), encoding);
         if (result.isPresent()) {
-            return success(result.get().closeBranch());
+            final Environment newEnvironment = result.get().closeBranch();
```
jvdb: It shouldn't make a difference, but I agree we should be consistent.

core/src/test/java/io/parsingdata/metal/TreeTest.java
jvdb: With the new `Pre` behaviour, `Opt` would be needed to prevent parse failures on the terminator.

core/src/test/java/io/parsingdata/metal/data/callback/CallbackTest.java
jvdb: Great point. It would be really nice to much more explicitly define these steps. Also then to adapt those `head.tail.tail.tail`-style tests (that you mention later on as well...)

PR #168, opened 2017-04-21
Force argument check at compile time for the seq and cho shorthands.

mvanaken:
Moved the exception tests to the ArgumentsTest.


core/src/main/java/io/parsingdata/metal/Shorthand.java
```diff
@@ -203,4 +204,8 @@ private Shorthand() {}
         return outBytes;
     }
 
+    private static Token[] concatVarArgTokens(final Token token1, final Token token2, final Token... tokens) {
+        return Stream.concat(Stream.of(token1), Stream.concat(Stream.of(token2), Stream.of(tokens))).toArray(Token[]::new);
```
rdvdijk: Would this work? 

```java
Stream.concat(Stream.of(token1, token2), Stream.of(tokens)).toArray(Token[]::new);
```

Also: I assume that the appropriate `null`-checking is done in the actual `Token`s?

rdvdijk:
@jvdb I see "An error occured during analysis" for the Better Code Hub analysis. Can't really see why, can you?

core/src/main/java/io/parsingdata/metal/Shorthand.java
rdvdijk: Wow. You changed it before I even committed by review? Nice.

mvanaken: :)

Yes, your assumption is right. The constructors of Seq and Cho check for `null` for all tokens.

core/src/main/java/io/parsingdata/metal/Shorthand.java
```diff
@@ -90,10 +91,10 @@ private Shorthand() {}
     public static Token def(final String name, final long size, final Encoding encoding) { return def(name, con(size), encoding); }
     public static Token def(final String name, final long size) { return def(name, size, (Encoding)null); }
     public static final Token empty = def(NO_NAME, 0L);
-    public static Token cho(final String name, final Encoding encoding, final Token... tokens) { return new Cho(name, encoding, tokens); }
-    public static Token cho(final String name, final Token... tokens) { return cho(name, null, tokens); }
-    public static Token cho(final Encoding encoding, final Token... tokens) { return cho(NO_NAME, encoding, tokens); }
-    public static Token cho(final Token... tokens) { return cho((Encoding)null, tokens); }
+    public static Token cho(final String name, final Encoding encoding, final Token token1, final Token token2, final Token... tokens) { return new Cho(name, encoding, concatVarArgTokens(token1, token2, tokens)); }
```
ccreeten: I personally would opt to enforce these on the tokens as well, as discussed in the issue.

jvdb: Agreed! There doesn't seem to be a valid reason to have them be different... Plus it requires code in the `Shorthand`, which we try to avoid.

rdvdijk:
This looks great.

And: another contributor! :tada:

Issue #173, opened 2017-06-12
Reimplement ParseGraph methods with trampoline


jvdb:
As discussed (and decoupled from) #162, this refers to `ParseGraph`'s `add(ParseValue)`, `add(ParseReference)`, `addBranch()` and `closeBranch()` methods.

Issue #172, opened 2017-06-04
Abstract/parameterize helper methods with lambdas


jvdb:
`ByName.getAllValues()` and `ByToken.getAllValues()` only differ in their comparison code to determine a match. These methods can be merged and parameterized with a lambda that implements the comparison.

Additionally, inspect other helper code for similar cases and resolve those too.

Issue #166, opened 2017-04-20
Move and/or adapt all helper methods not in the parsing scope


jvdb:
Issue #162 handles reimplementation of all code (including helper methods) to prevent stack overflows by using trampolines. Since that issue explicitly aims to first convert everything in the parsing scope, this has provided an analysis of the code that is non-essential (i.e., not in the parsing scope) in metal-core:

 - [ ] `ByName.getValue()`
 - [ ] `ByName.get()`
 - [ ] `ByName.getAll()`
 - [ ] `ByToken.get()`
 - [ ] `ByToken.getAll()`
 - [ ] `ByType.getReferences()`

`ByItem.getGraphAfter()` and `Reversal.reverse(ParseGraph)` were also on this list, but were removed entirely, see #64.

While a lot of test code tends to depend on it, the above methods can be safely removed from metal-core. Perhaps they should be moved to a separate component such as metal-tools or metal-api.

If the methods are not moved, then the recursive ones should be reimplemented using trampolines.

This list or any decision around it may change based on progress on #169 and #172.

ccreeten:
Funny, my Eclipse could not compile lambdas referring to the inheritance trampoline interfaces (it didn't see the safe completed trampoline as a functional interface). Only in the most recent version it appears to be solved.

On that note, perhaps you could annotate all functional interfaces with @FunctionalInterface, so it becomes clear which are the case, and added with that, static checking to verify it.

Issue #161, opened 2017-03-03
All methods that can return a null should return an Optional instead


jvdb:
- [x] `ParseGraph.current()`
- [x] `ByOffset.findItemAtOffset()`

Some items were removed from this list since they will be removed from the core library as part of #166.

PR #174, opened 2017-06-12
Reimplement recursive code to use trampolines to prevent stack overflows

### LINKED ISSUE #162: Prevent stack overflows during parsing caused by recursive implementation by using trampolines
Because Metal is implemented with as little mutable state as possible, many functionality is implemented using recursion. Since the JVM does not currently implement tail call elimination, this regularly leads to stack overflows.

The goal is to prevent these stack overflows by letting the recursive functions use trampolines. This is a mechanism that encodes the next recursive step into a method (typically using a lambda expression) that can be popped off the stack and executed. As a result, this allows a recursive style of programming without the stack overflow issue.

Recursion is used extensively in Metal. The first step (and goal of this issue) is to reimplement all recursive code using trampolines that can get executed during parsing. So this includes:

`Token`s:
 - [x] `Rep`
 - [x] `RepN`
 - [x] `Seq`
 - [x] `Cho`
 - [x] `While`
 - [x] `Sub`, which delegates to `ByOffset.hasRootAtOffset()`
 - [x] `Tie`
 - [x] `TokenRef` (tree processing, see below)

`ValueExpression`s:
 - [x] `NameRef`, which delegates to `ByName.getAllValues()` and `Wrapping.wrap()`
 - [x] `TokenRef`, which delegates to `ByToken.getAllValues()` and `Wrapping.wrap()`
 - [x] `First`
 - [x] `Nth`, which uses Reversal.reverse()
 - [x] `Reverse`, which delegates to `Reversal.reverse()`
 - [x] `Fold`
 - [x] `Expand`, which delegates to `ImmutableList.add(ImmutableList)`
 - [x] `BinaryValueExpression`
 - [x] `UnaryValueExpression`
 - [x] `Expand`
 - [x] `Elvis`

`Expression`s:
 - [x] `ComparisonExpression`

Data structures:
 - [x] `ImmutableList<T>` (`createFromArray()` method and `add(ImmutableList)` method, latter does tree processing, see below)
 - [x] `DataExpressionSource` (`getValueAtIndex()` method)
 - [x] `Callbacks` (`handleCallbacks()` method, note: it's `void`)

The remaining recursive code should also use trampolines (or be moved). This is addressed in #166.

Also to consider is recursion in four methods that are used a lot and a not prone to recursing extensively:  `ParseGraph`'s `add(ParseValue)`, `add(ParseReference)`, `addBranch()` and `closeBranch()`. This is discussed in #173.

One issue to consider is whether the `Trampoline.next()` method should throw an `IOException`. This is required by all `Token` implementations (except `TokenRef`) because `Token.parse()` throws it. It is unnecessary for the others. Possible solutions appear to be either wrapping the `IOException` (ugly) or creating two trampoline interfaces (duplication). Currently the latter is used: there are now `Trampoline` and `SafeTrampoline`, where the former's `next()` and `computeResult()` methods may throw an `IOException`.

Some methods are difficult to implement using only tail call recursion because they process a tree and handle both head and tail cases recursively, of which then only one is in the tail call. Converting these to trampolines as is can still result in stack overflows in some cases, so they need to be rewritten to tail call recursion. Below is a list of these cases as encountered during conversion:

 - [x] `ByOffset.getLowestOffsetValue()`
 - [x] `ByName.getAllValues()`
 - [x] `ByToken.getAllValues()`
 - [x] `ImmutableList.add(ImmutableList)`
 - [x] `TokenRef.lookup()`
 - [x] `ParseGraph.current()`
 - [x] `ByToken.getAllRoots()`

jvdb:
See #162 for details on all considerations and affected code. Additionally, note that some code is not reimplemented. Not converted code should either be discussed in #166 or #173.

Some conversions may seem complex, but this is because trampolines work with tail recursion, while some recursive code in Metal was not tail recursive. So this code had to be rewritten to both be tail recursive _and_ utilize trampolines. For examples, see the list at the bottom of #162.

There is quite a lot of use of the `Reversal.reverse(ImmutableList)` method in the new implementations. While this may seem to be slow, the alternative (where every step of the recursion changes the order) is actually much slower. So my suggestion is that we accept the final reverse step at the end of many of these computations.

Additionally, note that there are two trampoline implementations: `Trampoline` and `SafeTrampoline`. They duplicate almost all of their code. The only difference is that on `Trampoline`, the `next()` and `computeResult()` methods may throw can `IOException` as required by `Token.parse()`. My proposal here is to keep both implementations as there may be some changes into whether we keep the `IOException`, so doing some complex engineering around this may be unnecessary for now.

Also, some methods may look very similar due to the overlap in scaffolding in the trampolines. Note that there is also a separate issue to address this: #172. In order to keep the size of this already far too large PR down, I propose we keep the duplication for now.

Resolves #162.


core/src/main/java/io/parsingdata/metal/token/Cho.java
```diff
@@ -43,27 +46,23 @@
...
-    private Optional<Environment> iterate(final String scope, final Environment environment, final Encoding encoding, final ImmutableList<Token> list) throws IOException {
+    private Trampoline<Optional<Environment>> iterate(final String scope, final Optional<Environment> environment, final Encoding encoding, final ImmutableList<Token> list) throws IOException {
         if (list.isEmpty()) {
-            return failure();
+            return complete(Util::failure);
         }
-        final Optional<Environment> result = list.head.parse(scope, environment, encoding);
-        if (result.isPresent()) { return result; }
-        return iterate(scope, environment, encoding, list.tail);
+        return list.head.parse(scope, environment.get(), encoding)
```
rdvdijk: Is this `environment.get()` safe?

core/src/main/java/io/parsingdata/metal/token/TokenRef.java
```diff
@@ -60,22 +64,20 @@ public TokenRef(final String name, final String referenceName, final Encoding en
...
-        }
-        if (!item.isGraph() || item.asGraph().isEmpty()) { return LOOKUP_FAILED; }
-        final Token headResult = lookup(item.asGraph().head, referenceName);
-        if (headResult != LOOKUP_FAILED) { return headResult; }
-        return lookup(item.asGraph().tail, referenceName);
+    private SafeTrampoline<Token> lookup(final ImmutableList<ParseItem> items, final String referenceName) {
+        if (items.isEmpty()) { return complete(() -> LOOKUP_FAILED); }
+        final ParseItem item = items.head;
+        if (item.getDefinition().name.equals(referenceName)) { return complete(item::getDefinition); }
+        if (item.isGraph() && !item.asGraph().isEmpty()) { return intermediate(() -> lookup(items.tail.add(item.asGraph().tail).add(item.asGraph().head), referenceName)); }
```
rdvdijk: Move `if` bodies to new lines.

core/src/test/java/io/parsingdata/metal/TrampolineTest.java
```diff
@@ -0,0 +1,76 @@
...
+    public void nextOnFinalSafeTrampoline() {
+        thrown.expect(UnsupportedOperationException.class);
+        thrown.expectMessage("A CompletedTrampoline does not have a next computation.");
+        ((SafeTrampoline.CompletedTrampoline<Integer>) () -> 42).next();
+    }
+
+    @Test
+    public void noStackOverflow() {
+        // The 100000th Fibonacci number has 20899 digits
+        Assert.assertEquals(20899, fibonacci(BigInteger.ZERO, BigInteger.ONE, 100000).computeResult().toString().length());
```
rdvdijk: Neat :+1:

Nitpick: `static import` `assertEquals`.

core/src/main/java/io/parsingdata/metal/token/Pre.java
```diff
@@ -57,7 +58,7 @@ public Pre(final String name, final Token token, final Expression predicate, fin
             return failure();
         }
         final Optional<Environment> result = token.parse(scope, environment.addBranch(this), encoding);
-        return result.isPresent() ? success(result.get().closeBranch()) : failure();
+        return result.map(resultEnvironment -> success(resultEnvironment.closeBranch())).orElseGet(Util::failure);
```
rdvdijk: Code style: In `Rep` (below), these span multiple lines, let's do that everywhere. So:

```java
return result
    .map(resultEnvironment -> success(resultEnvironment.closeBranch()))
    .orElseGet(Util::failure);
```

core/src/main/java/io/parsingdata/metal/SafeTrampoline.java
```diff
@@ -0,0 +1,51 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.parsingdata.metal;
+
+public interface SafeTrampoline<T> {
```
ccreeten: But you will have to watch out not to use Foo as static type where Bar is used, as then users will have to handle IOException anyway.

core/src/test/java/io/parsingdata/metal/TrampolineTest.java
ccreeten: I wonder what the average stack size has to be in order for a standard recursive method to complete without overflow :smile:

core/src/main/java/io/parsingdata/metal/data/DataExpressionSource.java
```diff
@@ -45,17 +48,17 @@ public DataExpressionSource(final ValueExpression dataExpression, final int inde
     protected byte[] getData(final long offset, final int size) throws IOException {
         final ImmutableList<Optional<Value>> results = dataExpression.eval(graph, encoding);
         if (results.size <= index) { throw new IllegalStateException("ValueExpression dataExpression yields " + results.size + " result(s) (expected at least " + (index + 1) + ")."); }
-        final byte[] inputData = getValueAtIndex(results, index, 0).get().getValue();
+        final byte[] inputData = getValueAtIndex(results, index, 0).computeResult().get().getValue();
```
ccreeten: This is a personal thing, but I prefer to create a method with ''logical/expected" parameters, which then passes extra necessary parameters to the computing method, e.g. in this case, you only call ```getValueAtIndex(results, index)```, even for private methods. (also, this method assumes that the given index is valid with respect to the list size?)

ccreeten:
I know performance is not a thing as for now, but have you tested it on some real world cases to see what kind effect these refactorings have? Probably isn't too bad, but I think it is interesting to know.

core/src/main/java/io/parsingdata/metal/SafeTrampoline.java
jvdb: @ccreeten Good point. But I think we can take this risk as trampolines are never exposed in the public API, so it's purely an internal thing for us to deal with...

core/src/main/java/io/parsingdata/metal/Trampoline.java
```diff
@@ -0,0 +1,53 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.parsingdata.metal;
+
+import java.io.IOException;
+
+public interface Trampoline<T> {
```
jvdb: I will add some documentation. JavaDoc is not necessary I think since we do not expose any trampolines in the public API. So it's only relevant for those maintaining/extending Metal.

core/src/main/java/io/parsingdata/metal/data/ImmutableList.java
```diff
@@ -59,9 +63,13 @@ private ImmutableList(final T head, final ImmutableList<T> tail) {
 
     public ImmutableList<T> add(final ImmutableList<T> list) {
         checkNotNull(list, "list");
-        if (list.isEmpty()) { return this; }
```
jvdb: It's a duplicate check, especially since it cannot be removed from the `addRecursive()` method (as that is peeling off the `list` and needs the empty check to end recursion eventually). So if we keep it there we have duplicate code as well as a redundant duplicate check on all added non-empty lists.

core/src/main/java/io/parsingdata/metal/data/ParseGraph.java
```diff
@@ -97,15 +100,15 @@ ParseGraph closeBranch() {
...
     public ParseValue current() {
-        if (isEmpty()) { return null; }
-        if (head.isValue()) {
-            return head.asValue();
-        }
-        if (head.isGraph()) {
-            final ParseValue value = head.asGraph().current();
-            if (value != null) { return value; }
-        }
-        return tail.current(); // Ignore current if it's a reference (or an empty graph)
```
jvdb: I removed it on purpose because it doesn't add much to the understanding of the code but could turn stale if we ever add another type of `ParseItem`.

core/src/main/java/io/parsingdata/metal/data/ParseGraph.java
```diff
@@ -97,15 +100,15 @@ ParseGraph closeBranch() {
...
+        return current(ImmutableList.create(this)).computeResult();
+    }
+
+    private SafeTrampoline<ParseValue> current(ImmutableList<ParseItem> items) {
+        if (items.isEmpty()) { return complete(() -> null); }
```
jvdb: See #161. There are some situations (especially with `ValueExpression`s, where the default return value is already an `Optional`) where it won't be straightforward to remove the use of `null`. But I think this one can actually quite easily be fixed :)

core/src/main/java/io/parsingdata/metal/data/ParseGraph.java
```diff
@@ -97,15 +100,15 @@ ParseGraph closeBranch() {
...
+        final ParseItem item = items.head;
+        if (item.isValue()) { return complete(item::asValue); }
+        if (item.isGraph() && !item.asGraph().isEmpty()) { return intermediate(() -> current(items.tail.add(item.asGraph().tail).add(item.asGraph().head))); }
```
jvdb: Turned it into _four_ lines...

core/src/main/java/io/parsingdata/metal/data/selection/ByName.java
```diff
@@ -56,20 +60,21 @@ public static ParseValue getValue(final ParseGraph graph, final String name) {
     public static ImmutableList<Value> getAllValues(final ParseGraph graph, final String name) {
         checkNotNull(graph, "graph");
         checkNotNull(name, "name");
-        return getAllValuesRecursive(graph, name);
+        return reverse(getAllValuesRecursive(ImmutableList.create(graph), new ImmutableList<>(), name).computeResult());
```
jvdb: Everywhere where a `reverse()` was added where it didn't occur originally is because of the way the data structure is constructed: in the non-tail recursive implementations, the following pattern is used when traversing the data structure:

Are we currently at a match?
* No -> Return recurse.
* Yes -> X = Recurse. Add current result to X. Return X.

This ensures that the constructed data structure mirrors that of the original in that the _deepest_ match is put in the deepest part of the returned structure.

When converting to tail recursion, we cannot add something to the result of a recursive call, because we _must_ end with the recursive call. So instead we pass along an intermediate result that we add our local result to. But that means that the _first_ match ends up in the _deepest_ part of the resulting data structure.

So we end up calling `reverse()` on the final result to mirror the original order.

core/src/main/java/io/parsingdata/metal/data/ImmutableList.java
rdvdijk: Yes, I saw that it was added in the recursive step. The one reason for my comment is that if we leave the check here, there is no need for a trampoline at all. A _very_ minor optimization.

core/src/main/java/io/parsingdata/metal/data/selection/ByName.java
```diff
@@ -56,20 +60,21 @@ public static ParseValue getValue(final ParseGraph graph, final String name) {
...
         if (head.isValue() && head.asValue().matches(name)) {
-            return tailResults.add(head.asValue());
+            return intermediate(() -> getAllValuesRecursive(graphList.tail.add(graph.tail), values.add(head.asValue()), name));
         }
         if (head.isGraph()) {
-            return tailResults.add(getAllValuesRecursive(head.asGraph(), name));
+            return intermediate(() -> getAllValuesRecursive(graphList.tail.add(graph.tail).add(graph.head.asGraph()), values, name));
         }
-        return tailResults;
+        return intermediate(() -> getAllValuesRecursive(graphList.tail.add(graph.tail), values, name));
```
jvdb: Wow indeed! Implemented! I named the methods `addIfGraph()` and `addIfMatchingValue()` to express what they do.

core/src/main/java/io/parsingdata/metal/data/selection/ByOffset.java
```diff
@@ -31,35 +34,39 @@
...
         if (graph.head.isValue()) {
-            return getLowestOffsetValue(graph.tail, lowest == null || lowest.slice.offset > graph.head.asValue().slice.offset ? graph.head.asValue() : lowest);
+            return intermediate(() -> getLowestOffsetValue(graphList.tail.add(graph.tail), lowest == null || lowest.slice.offset > graph.head.asValue().slice.offset ? graph.head.asValue() : lowest));
         }
         if (graph.head.isGraph()) {
-            return getLowestOffsetValue(graph.tail, getLowestOffsetValue(graph.head.asGraph(), lowest));
+            return intermediate(() -> getLowestOffsetValue(graphList.tail.add(graph.head.asGraph()).add(graph.tail), lowest));
         }
-        return getLowestOffsetValue(graph.tail, lowest);
+        return intermediate(() -> getLowestOffsetValue(graphList.tail.add(graph.tail), lowest));
```
jvdb: Agreed. I think instead of refactoring many methods to a similar pattern, we should postpone that and do it as part of #172.

core/src/main/java/io/parsingdata/metal/data/selection/ByToken.java
```diff
@@ -68,38 +72,46 @@ public static ParseItem get(final ParseGraph graph, final Token definition) {
...
-        checkNotNull(definition, "definition");
-        return getAllRootsRecursive(graph, null, definition);
+        return getAllRootsRecursive(ImmutableList.create(new Pair(checkNotNull(graph, "graph"), null)), checkNotNull(definition, "definition"), new ImmutableList<>()).computeResult();
     }
 
-    private static ImmutableList<ParseItem> getAllRootsRecursive(final ParseItem item, final ParseGraph parent, final Token definition) {
-        final ImmutableList<ParseItem> result = item.getDefinition().equals(definition) && (parent == null || !parent.getDefinition().equals(definition))
-            ? ImmutableList.create(item)
-            : new ImmutableList<>();
+    private static SafeTrampoline<ImmutableList<ParseItem>> getAllRootsRecursive(final ImmutableList<Pair> list, final Token definition, final ImmutableList<ParseItem> result) {
```
jvdb: Renamed `list` to `backlog` (because it's the list of roots that still need to be checked) and `result` to `rootList` (because it's the list of found roots).

core/src/main/java/io/parsingdata/metal/data/selection/ByToken.java
```diff
@@ -68,38 +72,46 @@ public static ParseItem get(final ParseGraph graph, final Token definition) {
...
         if (item.isGraph() && !item.asGraph().isEmpty()) {
-            return result
-                .add(getAllRootsRecursive(item.asGraph().tail, item.asGraph(), definition))
-                .add(getAllRootsRecursive(item.asGraph().head, item.asGraph(), definition));
+            return intermediate(() -> getAllRootsRecursive(list.tail.add(new Pair(item.asGraph().head, item.asGraph())).add(new Pair(item.asGraph().tail, item.asGraph())), definition, nextResult));
+        }
+        return intermediate(() -> getAllRootsRecursive(list.tail, definition, nextResult));
+    }
+
+    static class Pair {
```
jvdb: Makes sense, but I don't want to clutter it too much with helper method definitions.

core/src/main/java/io/parsingdata/metal/expression/value/Fold.java
```diff
@@ -58,23 +61,24 @@ public Fold(final ValueExpression values, final BinaryOperator<ValueExpression>
...
         if (initial.size > 1) { return new ImmutableList<>(); }
         final ImmutableList<Optional<Value>> values = prepareValues(this.values.eval(graph, encoding));
-        if (values.isEmpty() || containsEmpty(values)) { return initial; }
+        if (values.isEmpty() || containsEmpty(values).computeResult()) { return initial; }
         if (!initial.isEmpty()) {
-            return ImmutableList.create(fold(graph, encoding, reducer, initial.head, values));
+            return ImmutableList.create(fold(graph, encoding, reducer, initial.head, values).computeResult());
         }
-        return ImmutableList.create(fold(graph, encoding, reducer, values.head, values.tail));
+        return ImmutableList.create(fold(graph, encoding, reducer, values.head, values.tail).computeResult());
```
jvdb: Not possible, because the passed in pair are in different cases:

1. `initial.head` + `values`
2. `values.head` + `values.tail`
3. `reducedValue.head` + `tail.tail`

core/src/main/java/io/parsingdata/metal/token/Cho.java
jvdb: Yes and no. There doesn't appear to be a use for an `Optional<Environment>` here at all, so I just changed it to `Environment`, which eliminates the `get()` outright.

But I did notice that there is no checking for `null` on both the `scope` and `environment` arguments (`encoding` is checked because it's explicitly allowed to be `null`). But before I make a big change like this while resolving some unrelated PR, let's investigate further: #175.

core/src/main/java/io/parsingdata/metal/token/TokenRef.java
jvdb: The rule was the put the return of a value on the same line, but actual code on a new line. So I agree on the bottom one but the others are just values.

core/src/test/java/io/parsingdata/metal/TrampolineTest.java
jvdb: I think the non-trampolined version stack overflowed somewhere between 1k and 10k (with my default stack size).

core/src/main/java/io/parsingdata/metal/data/DataExpressionSource.java
jvdb: With regard to methods: that's a good style. With private methods I tend to just immediately call the lowest level version :-)

With regard to final question: I think that's checked in the line above it?

jvdb:
@ccreeten: Excellent point. I haven't noticed significant slowdowns, but a somewhat comparative test is a good idea!

core/src/main/java/io/parsingdata/metal/data/ImmutableList.java
jvdb: Oh good point. The duplication can prevent the creation of an unnecessary trampoline :) In this case I will go for maintainability (= prevent duplication).

ccreeten: But what if the added check is more expensive? CPU don't like branching :laughing:

core/src/main/java/io/parsingdata/metal/data/DataExpressionSource.java
ccreeten: Yes, it is a bit tied to the style mentioned above. The method itself does not check it's parameters. But again, although I'd personally do it, for private methods I think it can be left as is.

core/src/main/java/io/parsingdata/metal/expression/value/Fold.java
rdvdijk: Ah, missed that! Got it.

core/src/main/java/io/parsingdata/metal/token/TokenRef.java
```diff
@@ -60,22 +64,22 @@ public TokenRef(final String name, final String referenceName, final Encoding en
...
+        return lookup(ImmutableList.create(environment.order), referenceName).computeResult().parse(scope, environment, encoding);
     }
 
-    private Token lookup(final ParseItem item, final String referenceName) {
-        if (item.getDefinition().name.equals(referenceName)) {
-            return item.getDefinition();
+    private SafeTrampoline<Token> lookup(final ImmutableList<ParseItem> items, final String referenceName) {
+        if (items.isEmpty()) { return complete(() -> LOOKUP_FAILED); }
+        final ParseItem item = items.head;
+        if (item.getDefinition().name.equals(referenceName)) { return complete(item::getDefinition); }
```
rdvdijk: One final thing: We need to do a bit of code style cleanup.

Why is this all one line:

```java
if (item.getDefinition().name.equals(referenceName)) { return complete(item::getDefinition); }
```

But this isn't? (From `While.java`, below.)

```java
        if (!environment.isPresent()) {
            return complete(Util::failure);
        }
```

As you might expect, the second style has my preference. Code style used to be a bit more consistent before this PR :relaxed:

core/src/main/java/io/parsingdata/metal/data/selection/ByName.java
```diff
@@ -56,20 +60,26 @@ public static ParseValue getValue(final ParseGraph graph, final String name) {
...
-        }
-        return tailResults;
+    private static SafeTrampoline<ImmutableList<Value>> getAllValuesRecursive(final ImmutableList<ParseGraph> graphList, final ImmutableList<Value> valueList, final String name) {
+        if (graphList.isEmpty()) { return complete(() -> valueList); }
+        final ParseGraph graph = graphList.head;
+        if (graph.isEmpty()) { return intermediate(() -> getAllValuesRecursive(graphList.tail, valueList, name)); }
+        return intermediate(() -> getAllValuesRecursive(addIfGraph(graphList.tail.add(graph.tail), graph.head),
+                                                        addIfMatchingValue(valueList, graph.head, name),
+                                                        name));
+    }
```
rdvdijk: that's my boy

core/src/main/java/io/parsingdata/metal/token/TokenRef.java
rdvdijk: And to add to that.

Previously 'simple' return statements have now become 'not so simple'. E.g.:

This was previously accepted by our own code style rules:

```java
if (something) { return head; }
```

In _many_ cases, this has now turned into:

```java
if (something) { return complete(() -> head); }
```

No so 'simple' anymore... Food for thought.

jvdb: We will have to do another round of cleanup before release.

core/src/test/java/io/parsingdata/metal/TrampolineTest.java
```diff
@@ -0,0 +1,78 @@
...
+        ((SafeTrampoline.CompletedTrampoline<Integer>) () -> 42).next();
+    }
+
+    @Test
+    public void noStackOverflow() {
+        // The 100000th Fibonacci number has 20899 digits
+        assertEquals(20899, fibonacci(BigInteger.ZERO, BigInteger.ONE, 100000).computeResult().toString().length());
+    }
+
+    private SafeTrampoline<BigInteger> fibonacci(final BigInteger l, final BigInteger r, final long count) {
```
mvanaken: This is awesome! 😎

core/src/main/java/io/parsingdata/metal/SafeTrampoline.java
```diff
@@ -0,0 +1,52 @@
...
+    @Override
+    default T computeResult() {
+        SafeTrampoline<T> current = this;
+        while (current.hasNext()) {
+            current = current.next();
+        }
+        return current.result();
+    }
+
+    static <T> SafeTrampoline<T> complete(final CompletedTrampoline<T> completedTrampoline) { return completedTrampoline; }
```
jvdb: Correct! Without these methods we could still directly use lambda's, but they would need to be type casted. So these methods prevent that by explicitly naming the types of the functional interfaces involved.

ccreeten:
Funny, my Eclipse could not compile lambdas referring to the inheritance trampoline interfaces (it didn't see the safe completed trampoline as a functional interface). Only in the most recent version it appears to be solved.

On that note, perhaps you could annotate all functional interfaces with @FunctionalInterface, so it becomes clear which are the case, and added with that, static checking to verify it.

PR #176, opened 2017-06-13
Null checks on Token.parse() for the scope and environment arguments.

### LINKED ISSUE #175: Add checks for null to scope and environment arguments to Token.parse()

jvdb:
Resolves #175.


core/src/test/java/io/parsingdata/metal/token/TokenTest.java
```diff
@@ -0,0 +1,40 @@
...
+public class TokenTest {
+
+    @Rule
+    public ExpectedException thrown = ExpectedException.none();
+    private final Token token = new Token("", null) {
+        @Override
+        protected Optional<Environment> parseImpl(String scope, Environment environment, Encoding encoding) throws IOException {
+            return null;
+        }
+    };;
```
rdvdijk: Double `;`.

PR #177, opened 2017-06-15
Make all ParseGraph constructors private

### LINKED ISSUE #64: Check private constructor issue in ParseGraph
The only private constructor in `ParseGraph` is used only by two helper methods (`ByItem.getGraphAfter()` and `Reversal.reverse(ParseGraph)`) which in turn are only used in test code and not in any client code we know of. So the solution is: make the constructor private, remove the helper methods and repair the test code to not depend on them.

jvdb:
There was only one `ParseGraph` constructor not private. Making it private was a problem for two helper methods that used it: `ByItem.getGraphAfter()` and `Reversal.reverse(ParseGraph)`. Both were removed (along with their tests) since they were only used in test code and not in any client code we know of. Since `getGraphAfter()` was the only public method of `ByItem`, the utility class was removed entirely.

As a result of these changes, `ParseGraph` was removed from `AutoEqualityTest` (which requires a public constructor). To bring the code coverage back to 100% a hand-written test was added to `EqualityTest`.

Finally, `TreeTest` relied heavily on `Reversal.reverse(ParseGraph)` so parts of the test code there was rewritten.

Resolves #64.


core/src/test/java/io/parsingdata/metal/TreeTest.java
```diff
@@ -89,67 +87,65 @@ public TreeTest() throws IOException {
...
+            assertTrue(subStruct.tail.head.asValue().matches(name));
+            final int pointer = subStruct.tail.head.asValue().asNumeric().intValue();
+            assertTrue(subStruct.head.isGraph()); // Sub
+            if (subStruct.head.asGraph().head.isReference()) { // If the Sub contains a Reference, it's a cycle
+                checkResolve(subStruct.head.asGraph().head.asReference(), root, pointer);
+            } else {
+                checkStructure(subStruct.head.asGraph().head.asGraph(), root, pointer); // Recurse
+            }
+        }
+    }
```
rdvdijk: Kudos. This test is actually much easier to grok with these comments!

Issue #169, opened 2017-05-10
Optimize last(ref(x)) expressions by parameterizing result counts


jvdb:
Sometimes an expression such as `last(ref("x"))` can be extremely slow, for example if the `ParseGraph` is extremely large and/or if there are many `x`es. This would not be bad if all `x`es were needed, but as the `last(...)` around the `ref("x")` reveals, only the first one is needed.

There are basic rewriting techniques, but these are ugly because Java doesn't have pattern matching, so would boil down to `instanceof` code that replaces `last(ref(...))` for example with an invocation of a custom `ValueExpression` such as `lastRef(...)` (which would also need to be implemented).

A better approach would be to pass a `limit` argument to `ValueExpression.eval()` to specify the maximum amount of desired results. This way, implementations can stop and return once they collect the maximum amount.

There is a problem with this approach as well however, which is that it's not possible to limit every type of expression in the same way. Suppose the following expression: `last(reverse(ref("a")))`. If the `last()` would simply pass a `limit=1` argument to `reverse()` and that would in turn pass it on, the final result would be the last `a` instead of the first as probably intended.

Instead, a workaround is to add a `limit` argument to `Ref<T>` along with a specific `last(Ref<T>)` factory method in `Shorthand` that optimizes the specific case of only referencing the last item in a `Ref<T>` expression.

Issue #133, opened 2017-01-01
Update copyrights


rdvdijk:
A new year, a new copyright ending year! :wink:

- [ ] source code
- [ ] README.md

jvdb:
As discussed, let's determine what a good approach is to handling this. It seems that the year only indicates an initial claim to copyright. So this would mean we don't need to update this all the time. At the same time, if we make significant changes it might make sense to update it...

PR #179, opened 2017-06-23
Generalize helper methods

### LINKED ISSUE #172: Abstract/parameterize helper methods with lambdas

mvanaken:
Resolves #172.

### DIFF
--- a/core/src/main/java/io/parsingdata/metal/data/selection/ByPredicate.java
+++ b/core/src/main/java/io/parsingdata/metal/data/selection/ByPredicate.java
@@ -27,7 +27,9 @@ import io.parsingdata.metal.data.ParseGraph;
 import io.parsingdata.metal.data.ParseItem;
 import io.parsingdata.metal.data.ParseValue;
 
-class ByPredicate {
+public final class ByPredicate {
+
+    private ByPredicate() {}
 
     public static final int NO_LIMIT = -1;
 
--- a/core/src/main/java/io/parsingdata/metal/data/selection/ByPredicate.java
+++ b/core/src/main/java/io/parsingdata/metal/data/selection/ByPredicate.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright 2013-2016 Netherlands Forensic Institute
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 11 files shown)


core/src/main/java/io/parsingdata/metal/data/transformation/Wrapping.java
```diff
@@ -30,14 +30,14 @@
 
     private Wrapping() {}
 
-    public static <T> ImmutableList<Optional<T>> wrap(final ImmutableList<T> list) {
+    public static <T, U extends T> ImmutableList<Optional<T>> wrapToBase(final ImmutableList<U> list) {
```
rdvdijk: I would be OK if we just keep the name `wrap()`.

core/src/main/java/io/parsingdata/metal/data/selection/ByPredicate.java
```diff
@@ -0,0 +1,58 @@
+/*
+ * Copyright 2013-2016 Netherlands Forensic Institute
```
rdvdijk: It's 2017 😉

core/src/main/java/io/parsingdata/metal/data/selection/ByPredicate.java
```diff
@@ -0,0 +1,58 @@
...
+import io.parsingdata.metal.data.ParseValue;
+
+public final class ByPredicate {
+
+    private ByPredicate() {}
+
+    public static final int NO_LIMIT = -1;
+
+    static SafeTrampoline<ImmutableList<ParseValue>> getAllValues(final ImmutableList<ParseGraph> graphList, final ImmutableList<ParseValue> valueList, final Predicate<ParseValue> predicate, final int limit) {
+        if (graphList.isEmpty() || valueList.size == limit) { return complete(() -> valueList); }
```
rdvdijk: I think we have tests for "get all" and "get last", but no explicit test for something like "get the last 3". Maybe add that?

Also: this is not (yet?) exposed to the `ref()` and `token()` expressions, right? Is there an issue yet to add that?

core/src/main/java/io/parsingdata/metal/data/selection/ByPredicate.java
jvdb: Agreed. There's a separate issue however, because it's best if it's one big commit to fix this: #133.

core/src/main/java/io/parsingdata/metal/data/selection/ByPredicate.java
jvdb: #169. And good point about explicitly testing other values for `limit` (possibly also weird edge cases, where `limit` > available, etc.

However, I have some additional refactoring plans that may make those tests unnecessary, so my proposal is to not dive into adding many tests. But just so we don't forget: #180.

PR #182, opened 2017-06-28
Collapse NameRef and TokenRef

### LINKED ISSUE #181: Unify NameRef and TokenRef implementations
After the refactoring of #172, `NameRef` and `TokenRef` have become extremely similar. So as a prelude to #166, collapse those implementations into a single `Ref<T>` implementation.

jvdb:
Resolves #181.

`NameRef` and `TokenRef` are replaced by `Ref<T>`, which has factory methods for both `String` (name) and `Token` (definition). Internally these arguments are transformed via a fixed mapping onto a predicate. Since the generated predicate always has the same structure, only the argument is used for equality checking.

Since `Wrapping.wrap()` is now only used by `Ref<T>`, it has been moved there and the `Wrapping` class is gone.

Explicit equality tests have been added, because `Ref<T>` only has a private constructor, so it cannot be tested by `AutoEqualityTest` and `ArgumentsTest`.


core/src/main/java/io/parsingdata/metal/expression/value/reference/Ref.java
```diff
@@ -0,0 +1,90 @@
...
+ * @param <T> The type of reference to match on.
+ */
+public class Ref<T> implements ValueExpression {
+
+    public final T reference;
+    public final Predicate<ParseValue> predicate;
+
+    private Ref(final T reference, final Predicate<ParseValue> predicate) {
+        this.reference = checkNotNull(reference, "reference");
+        this.predicate = checkNotNull(predicate, "predicate");
```
rdvdijk: My OCD is very satisfied by the fact that these two words have the same amount of characters. :relieved:

core/src/main/java/io/parsingdata/metal/expression/value/reference/Nth.java
```diff
@@ -46,9 +46,6 @@
  * <code>values</code>. For every invalid index (such as
  * {@link Optional#empty()}, a negative value or an index that is out of
  * bounds) empty is returned.
- *
- * @see NameRef
- * @see TokenRef
```
rdvdijk: `@see Ref` ?

core/src/main/java/io/parsingdata/metal/data/selection/ByToken.java
```diff
@@ -70,12 +70,6 @@ public static ParseItem get(final ParseGraph graph, final Token definition) {
         return results;
     }
 
-    public static ImmutableList<ParseValue> getAllValues(final ParseGraph graph, final Token definition) {
-        checkNotNull(graph, "graph");
-        checkNotNull(definition, "definition");
-        return reverse(ByPredicate.getAllValues(ImmutableList.create(graph), new ImmutableList<>(), (value) -> value.definition.equals(definition), NO_LIMIT).computeResult());
```
rdvdijk: The static import `NO_LIMIT` should disappear from this file, right?

core/src/main/java/io/parsingdata/metal/expression/value/reference/Nth.java
jvdb: Some of those `@see`s were quite random. The only one that made sense was a reference from `token/TokenRef` because the name fully overlapped. But now that it doesn't overlap anymore, I've removed it as well.

PR #183, opened 2017-06-28
Added limit test for ByPredicate.getAllValues()

### LINKED ISSUE #180: Test limit argument on helper methods
As introduced by #172 and discussed in #169.

jvdb:
Resolves #180.

When having 5 results, tests whether a limit of 0-5 returns those amounts and whether 6 also returns 5.


jvdb:
The codecov failure seems to be related to a master build that failed earlier (which was due to some internal problem at travis and/or codecov).

PR #184, opened 2017-06-29
Optimize last(ref(x)) expressions

### LINKED ISSUE #169: Optimize last(ref(x)) expressions by parameterizing result counts

jvdb:
Resolves #169.

Turned the `limit` argument into a `ValueExpression` as well. Added accompanying `Shorthand` definitions. Added various tests.

### DIFF
--- a/core/src/main/java/io/parsingdata/metal/data/selection/ByName.java
+++ b/core/src/main/java/io/parsingdata/metal/data/selection/ByName.java
@@ -17,7 +17,6 @@
 package io.parsingdata.metal.data.selection;
 
 import static io.parsingdata.metal.Util.checkNotNull;
-import static io.parsingdata.metal.data.selection.ByPredicate.NO_LIMIT;
 import static io.parsingdata.metal.data.transformation.Reversal.reverse;
 
 import io.parsingdata.metal.data.ImmutableList;
@@ -34,7 +33,7 @@ public final class ByName {
      * @return The first value (bottom-up) with the provided name in the provided graph
      */
     public static ParseValue getValue(final ParseGraph graph, final String name) {
-        return ByPredicate.getAllValues(ImmutableList.create(graph), new ImmutableList<>(), (value) -> value.matches(name), 1).computeResult().head;
+        return ByPredicate.getAllValues(graph, (value) -> value.matches(name), 1).head;
     }
 
     /**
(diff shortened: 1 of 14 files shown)


core/src/main/java/io/parsingdata/metal/data/selection/ByName.java
```diff
@@ -34,7 +33,7 @@ private ByName() {}
      * @return The first value (bottom-up) with the provided name in the provided graph
      */
     public static ParseValue getValue(final ParseGraph graph, final String name) {
-        return ByPredicate.getAllValues(ImmutableList.create(graph), new ImmutableList<>(), (value) -> value.matches(name), 1).computeResult().head;
+        return ByPredicate.getAllValues(graph, (value) -> value.matches(name), 1).head;
```
rdvdijk: What happened here?

core/src/test/java/io/parsingdata/metal/ReferenceValueExpressionSemanticsTest.java
```diff
@@ -101,4 +108,11 @@ private static Token refMatch(final Expression pred) {
                 def("b", con(1), pred));
     }
 
+    private static Token limitedSum(final int limit) {
+        return
+            seq(rep(def("a", con(1), not(eq(con(0))))),
+                def("zero", con(1), eq(con(0))),
+                def("sum", con(1), eq(fold(new NameRef("a", con(limit)), Shorthand::add))));
```
jvdb: It's a remnant from before there was a shorthand which included the `limit` argument :) I'll fix it.

core/src/main/java/io/parsingdata/metal/data/selection/ByName.java
jvdb: Before there was a nice public and static method exposing the `limit` argument, the only way to use it was by directly calling the internal (package private) method. Essentially what I forgot to do with the `NameRef` constructor above :)

PR #185, opened 2017-06-29
Convert methods that may return a null to returning an Optional

### LINKED ISSUE #161: All methods that can return a null should return an Optional instead

jvdb:
Resolves #161.

Two methods converted: `ParseGraph.current()` and `ByOffset.findItemAtOffset()`. There are some more methods, but they are scheduled to be removed from core (see #166).

One pattern that is used a couple of times is notable: there are placed where `A extends B` and we must return an `Optional<B>` but we have an `Optional<A>`. This can be easily done by mapping to the identity function: `var.map(x -> x)` which maps the value inside the `Optional` to the desired type. This happens where we have an `Optional<ParseValue>` but we need an `Optional<Value>`.

### DIFF
--- a/core/src/main/java/io/parsingdata/metal/expression/comparison/ComparisonExpression.java
+++ b/core/src/main/java/io/parsingdata/metal/expression/comparison/ComparisonExpression.java
@@ -16,6 +16,8 @@
 
 package io.parsingdata.metal.expression.comparison;
 
+import static java.util.function.Function.identity;
+
 import static io.parsingdata.metal.SafeTrampoline.complete;
 import static io.parsingdata.metal.SafeTrampoline.intermediate;
 import static io.parsingdata.metal.Util.checkNotNull;
@@ -57,7 +59,7 @@ public abstract class ComparisonExpression implements Expression {
 
     @Override
     public boolean eval(final ParseGraph graph, final Encoding encoding) {
-        final ImmutableList<Optional<Value>> values = value == null ? ImmutableList.create(graph.current().map(parseValue -> parseValue)) : value.eval(graph, encoding);
+        final ImmutableList<Optional<Value>> values = value == null ? ImmutableList.create(graph.current().map(identity())) : value.eval(graph, encoding);
         if (values.isEmpty()) { return false; }
         final ImmutableList<Optional<Value>> predicates = predicate.eval(graph, encoding);
         if (values.size != predicates.size) { return false; }
--- a/core/src/main/java/io/parsingdata/metal/expression/comparison/ComparisonExpression.java
+++ b/core/src/main/java/io/parsingdata/metal/expression/comparison/ComparisonExpression.java
@@ -57,7 +57,7 @@ public abstract class ComparisonExpression implements Expression {
 
     @Override
     public boolean eval(final ParseGraph graph, final Encoding encoding) {
-        final ImmutableList<Optional<Value>> values = value == null ? ImmutableList.create(Optional.of(graph.current())) : value.eval(graph, encoding);
+        final ImmutableList<Optional<Value>> values = value == null ? ImmutableList.create(graph.current().map(parseValue -> parseValue)) : value.eval(graph, encoding);
         if (values.isEmpty()) { return false; }
         final ImmutableList<Optional<Value>> predicates = predicate.eval(graph, encoding);
         if (values.size != predicates.size) { return false; }
--- a/core/src/main/java/io/parsingdata/metal/expression/value/reference/Self.java
+++ b/core/src/main/java/io/parsingdata/metal/expression/value/reference/Self.java
@@ -34,7 +34,7 @@ public class Self implements ValueExpression {
 
     @Override
     public ImmutableList<Optional<Value>> eval(final ParseGraph graph, final Encoding encoding) {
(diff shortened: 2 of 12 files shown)


core/src/main/java/io/parsingdata/metal/expression/comparison/ComparisonExpression.java
```diff
@@ -57,7 +57,7 @@ public ComparisonExpression(final ValueExpression value, final ValueExpression p
 
     @Override
     public boolean eval(final ParseGraph graph, final Encoding encoding) {
-        final ImmutableList<Optional<Value>> values = value == null ? ImmutableList.create(Optional.of(graph.current())) : value.eval(graph, encoding);
+        final ImmutableList<Optional<Value>> values = value == null ? ImmutableList.create(graph.current().map(parseValue -> parseValue)) : value.eval(graph, encoding);
```
rdvdijk: `.map(parseValue -> parseValue)` looks very weird 😉 

IntelliJ is telling me we can replace `parseValue -> parseValue` with `Function.identity()`. (I propose we `static import` `identity()`?)

core/src/main/java/io/parsingdata/metal/expression/value/reference/Self.java
```diff
@@ -34,7 +34,7 @@
 
     @Override
     public ImmutableList<Optional<Value>> eval(final ParseGraph graph, final Encoding encoding) {
-        return ImmutableList.create(Optional.of(graph.current()));
+        return ImmutableList.create(graph.current().map(parseValue -> parseValue));
```
rdvdijk: As mentioned elsewhere, use `Function.identity()`.

core/src/main/java/io/parsingdata/metal/expression/comparison/ComparisonExpression.java
ccreeten: Ahhh, generics :smile: I think an other way could be to declare the type of the values variable and compare method parameter as something like ```ImmutableList<? extends Optional<? extends Value>>```.

jvdb: @rdvdijk I was actually hoping something like `Function.identity()` would exist but couldn't find it late last night :) I agree that would be much better.

@ccreeten I agree that could make the code nicer by removing the need for such a conversion. But at the same time it's also good to have easy-to-understand signatures.

core/src/main/java/io/parsingdata/metal/expression/comparison/ComparisonExpression.java
```diff
@@ -57,7 +59,7 @@ public ComparisonExpression(final ValueExpression value, final ValueExpression p
 
     @Override
     public boolean eval(final ParseGraph graph, final Encoding encoding) {
-        final ImmutableList<Optional<Value>> values = value == null ? ImmutableList.create(Optional.of(graph.current())) : value.eval(graph, encoding);
+        final ImmutableList<Optional<Value>> values = value == null ? ImmutableList.create(graph.current().map(identity())) : value.eval(graph, encoding);
```
mvanaken: I think I get it now... This identity is required to convert the `Optional<ParseValue>` to `Optional<Value>`, and the identity method works, even thought it is T -> T, because the variable is defined as `ImmutableList<Optional<Value>>`.

Seriously, it is mind blowing!

mvanaken: The null check for value makes me think: wouldn't it be nice if we can remove all the null values in the shorthands? I think we could have an `Empty` ValueExpression (or something) and overload the constructor with only a predicate and initialize value with the Empty ValueExpression. We then let this empty value expression return `ImmutableList.create(graph.current().map(identity()))` when `value.eval(...)` and it is sufficient to simply call for `value.eval(graph, encoding)` here. This can be adjusted for all the null usages in the shorthands.

Just a thought! I'll stop typing now...

[Edit: Oh hee, there already is the `Self` ValueExpression which contains `ImmutableList.create(graph.current().map(identity()))` in its `eval` so we could use that? Ok, I really stop typing now!]

`<MUTE OFF>`
[Edit 2:  Oh! Or... simple lambda expressions! (I see different implementation for when value == null). Would be nice to have a nullless metal! ]
`<MUTE ON>`

jvdb: Indeed :)

jvdb: I was thinking along similar lines when editing this code. I don't think there is a clean default such ad `self` here because sometimes the `null` really means _nothing_.

But your suggestion is quite interesting: we could perhaps create an _empty_ expression that throws an exception if you `eval()` it, but exists solely to signal that it formally doesn't exist. Still, the end result would be similar to a `null` (exception thrown when not checking). Worth exploring!

ccreeten: Because of the target type of the map expression :smile:

PR #189, opened 2017-07-12
Added Until token for variable size values

### LINKED ISSUE #109: Add Token to express a value with a terminator
The most basic version of such a `Token` would be (in `Shorthand` notation):

`until("data", post(def("terminator", con(1)), eq(con(0))))`

which would be equivalent to the following:
```
seq(
    def("data", con(X)),
    post(def("terminator", con(1)), eq(con(0))))
```
Where `X` would be the variable that is iterated over.

Additionally, this entire approach can be generalized as:

```
range("x",
    seq(
        def("data", last(ref("x"))),
        post(def("terminator", con(1)), eq(con(0)))))
```
Where "x" is the name of a variable that is bound to a value in the scope of the `range` token. This value
would then be iterated over until the nested `Token` parses.

Arguments to consider (both to a basic `until` and/or a generalized `range` token):

1. Names. In the `until` case for both the nested `seq` as well as the variable size `def`.
2. Range values, both a `minSize` as well as a `maxSize`, which would both be a `ValueExpression` to define the start and end values of the iteration.
3. Interval size, a `stepSize` `ValueExpression` to set the value to grow with.

Example for 2 and 3: `minSize`=3, `maxSize`=9, `stepSize`=2 would attempt the values 3, 5, 7 and 9.

A fully generalized and maximally configurable `range` `Token` would then have the following signature:

`Token range(String varName, ValueExpression minSize, ValueExpression maxSize, ValueExpression stepSize, Token token)`

A thing to consider is whether a direct implementation of `until` would have major runtime performance advantages over having it as a `Shorthand` for common `range` cases.

rdvdijk:
Resolves #109.

This token can be used to parse zero-terminated strings and *varints*.

See `UntilTest` for examples of inclusive (`untilInclusive()`) and exclusive (`threeNewLines()`) parsing.

### DIFF
--- a/core/src/main/java/io/parsingdata/metal/token/Until.java
+++ b/core/src/main/java/io/parsingdata/metal/token/Until.java
@@ -31,10 +31,36 @@ import io.parsingdata.metal.Util;
 import io.parsingdata.metal.data.Environment;
 import io.parsingdata.metal.data.ImmutableList;
 import io.parsingdata.metal.data.ParseValue;
+import io.parsingdata.metal.data.Slice;
 import io.parsingdata.metal.encoding.Encoding;
 import io.parsingdata.metal.expression.value.Value;
 import io.parsingdata.metal.expression.value.ValueExpression;
 
+/**
+ * A {@link Token} that specifies a value to parse in the input until
+ * another token is parsed.
+ * <p>
+ * An Until consists of an <code>initialSize</code>, a <code>stepSize</code>
+ * and a <code>maxSize</code> (all {@link ValueExpression}s) and a
+ * <code>terminator</code>(a {@link Token}). First <code>initialSize</code>,
+ * <code>stepSize</code>, and <code>maxSize</code> are evaluated. Using this
+ * token's name, a value of length <code>initialSize</code> is added to the
+ * <code>Environment</code> and an attempt is made to parse the
+ * <code>terminator</code>. If it succeeds, the resulting environment is
+ * returned. Otherwise, <code>stepSize</code> is added to the
+ * <code>initialSize</code> and if the resulting value is below
+ * <code>maxSize</code>, a value with the resulting size is added to the
+ * original <code>Environment</code> and a new attempt to parse the
+ * <code>terminator</code> is made. Parsing fails if no combination of any size
+ * is found where the <code>terminator</code> parses successfully.
+ * <p>
+ * If the <code>ValueExpressions</code> evaluate to lists, they are treated
+ * as sets of values to attempt. If <code>stepSize</code> is negative,
+ * <code>maxSize</code> is allowed to be smaller than <code>initialSize</code>.
+ * A <code>stepSize</code> of zero is not allowed.
+ *
+ * @see ValueExpression
+ */
 public class Until extends Token {
 
     public static final ValueExpression DEFAULT_INITIAL = con(0);
@@ -68,7 +94,9 @@ public class Until extends Token {
 
     private Trampoline<Optional<Environment>> iterate(final String scope, final Environment environment, final int currentSize, final int stepSize, final int maxSize, final Encoding encoding) throws IOException {
         if (stepSize == 0 || stepSize > 0 && currentSize > maxSize || stepSize < 0 && currentSize < maxSize) { return complete(Util::failure); }
-        return terminator.parse(scope, currentSize == 0 ? environment : environment.add(new ParseValue(name, this, environment.slice(currentSize), encoding)).seek(environment.offset + currentSize), encoding)
+        final Slice slice = environment.slice(currentSize);
+        if (slice.size != currentSize) { return complete(Util::failure); }
+        return terminator.parse(scope, currentSize == 0 ? environment : environment.add(new ParseValue(name, this, slice, encoding)).seek(environment.offset + currentSize), encoding)
             .map(nextEnvironment -> complete(() -> success(nextEnvironment)))
             .orElseGet(() -> intermediate(() -> iterate(scope, environment, currentSize + stepSize, stepSize, maxSize, encoding)));
     }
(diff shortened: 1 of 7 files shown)


core/src/main/java/io/parsingdata/metal/token/Until.java
```diff
@@ -0,0 +1,131 @@
...
+ * returned. Otherwise, <code>stepSize</code> is added to the
+ * <code>initialSize</code> and if the resulting value is below
+ * <code>maxSize</code>, a value with the resulting size is added to the
+ * original <code>Environment</code> and a new attempt to parse the
+ * <code>terminator</code> is made. Parsing fails if no combination of any size
+ * is found where the <code>terminator</code> parses successfully.
+ * <p>
+ * If the <code>ValueExpressions</code> evaluate to lists, they are treated
+ * as sets of values to attempt. If <code>stepSize</code> is negative,
+ * <code>maxSize</code> is allowed to be smaller than <code>initialSize</code>.
```
ccreeten: Must be rather than allowed?

core/src/main/java/io/parsingdata/metal/token/Until.java
```diff
@@ -0,0 +1,131 @@
...
+ * A <code>stepSize</code> of zero is not allowed.
```
ccreeten: It is, but it simply makes the parse fail? (same with above actually)

core/src/main/java/io/parsingdata/metal/token/Until.java
```diff
@@ -0,0 +1,131 @@
...
+    private Trampoline<Optional<Environment>> iterate(final String scope, final Environment environment, final int currentSize, final int stepSize, final int maxSize, final Encoding encoding) throws IOException {
+        if (stepSize == 0 || stepSize > 0 && currentSize > maxSize || stepSize < 0 && currentSize < maxSize) { return complete(Util::failure); }
+        final Slice slice = environment.slice(currentSize);
+        if (slice.size != currentSize) { return complete(Util::failure); }
+        return terminator.parse(scope, currentSize == 0 ? environment : environment.add(new ParseValue(name, this, slice, encoding)).seek(environment.offset + currentSize), encoding)
+            .map(nextEnvironment -> complete(() -> success(nextEnvironment)))
+            .orElseGet(() -> intermediate(() -> iterate(scope, environment, currentSize + stepSize, stepSize, maxSize, encoding)));
+    }
+
+    private boolean checkNotValidList(ImmutableList<Optional<Value>> list) {
```
ccreeten: Nagging, but the tail is allowed to contain empty values?

ccreeten:
Cool stuff by the way!

core/src/main/java/io/parsingdata/metal/token/Until.java
```diff
@@ -0,0 +1,131 @@
...
+
+    @Override
+    public String toString() {
+        return getClass().getSimpleName() + "(" + makeNameFragment() + initialSize + "," + stepSize + "," + maxSize + "," + terminator + ")";
+    }
+
+    @Override
+    public boolean equals(final Object obj) {
+        return super.equals(obj)
+            && Objects.equals(initialSize, ((Until)obj).initialSize)
```
ccreeten: Is this safe? What if obj is an instance of an other subclass but equal to super? Then these casts will fail?

core/src/main/java/io/parsingdata/metal/token/Until.java
ccreeten: And I guess negative initial and max aren't either :smile:

core/src/test/java/io/parsingdata/metal/token/ParameterizedUntilTest.java
```diff
@@ -0,0 +1,87 @@
...
+    @Parameters(name = "{0} ({4})")
+    public static Collection<Object[]> data() {
+        return Arrays.asList(new Object[][] {
+            { "[a,b,c,a,b,c] i=0,s=1,m=6 ab",     untilToken(0, 1, 6, con('c'), con("ab")),     stream("abcabc", US_ASCII), enc(), true },
+            { "[a,b,c,a,b,c] i=3,s=1,m=6 abcab",  untilToken(3, 1, 6, con('c'), con("abcab")),  stream("abcabc", US_ASCII), enc(), true },
+            { "[a,b,c,a,b,c] i=0,s=2,m=6 ab",     untilToken(0, 2, 6, con('c'), con("ab")),     stream("abcabc", US_ASCII), enc(), true },
+            { "[a,b,c,a,b,c] i=1,s=2,m=6 abcab",  untilToken(1, 2, 6, con('c'), con("abcab")),  stream("abcabc", US_ASCII), enc(), true },
+            { "[a,b,c,a,b,c] i=0,s=3,m=6",        untilToken(0, 3, 6, con('c'), con("")),       stream("abcabc", US_ASCII), enc(), false },
+            { "[a,b,c,a,b,c] i=0,s=0,m=6",        untilToken(0, 0, 6, con(""), con("")),        stream("abcabc", US_ASCII), enc(), false },
+            { "[a,b,c,a,b,c] i=6,s=-1,m=0 abcab", untilToken(6, -1, 0, con('c'), con("abcab")), stream("abcabc", US_ASCII), enc(), true },
```
ccreeten: Cool!

core/src/test/java/io/parsingdata/metal/token/UntilTest.java
```diff
@@ -0,0 +1,86 @@
...
+        assertEquals(values.tail.tail.head.asString(), INPUT_1);
+    }
+
+    @Test
+    public void untilInclusive() throws IOException {
+        final Optional<Environment> environment = createToken(con(1), post(empty, eq(mod(last(ref("line")), con(256)), con('\n')))).parse(stream(INPUT, US_ASCII), enc());
+        assertTrue(environment.isPresent());
+        ImmutableList<ParseValue> values = getAllValues(environment.get().order, "line");
+        assertEquals(3, values.size);
+        assertEquals(values.head.asString(), INPUT_3 + '\n');
```
ccreeten: Perhaps use some matchers for this.

PR #190, opened 2017-07-12
Def and Until tokens should not allow empty name

rdvdijk:
This is because a `ParseValue` with an empty name leads to `matches()` anomalies.

```java
    public boolean matches(final String name) {
        return this.name.equals(name) || this.name.endsWith(Token.SEPARATOR + name);
    }
```


jvdb:
Why did codecov comment twice?

rdvdijk:
@jvdb Maybe because I made a 2nd comment (typo fix)?

jvdb:
@rdvdijk Normally it edits its own comments to update it... Its second comment here was also edited. Probably just lost track of the first one somehow :-)

rdvdijk:
I deleted the first one.

PR #195, opened 2017-07-24
Improved hashCode() implementations to prevent collisions

### LINKED ISSUE #150: Improve hashCode() implementations for similar classes
Especially expression classes such as subclasses of `Expression` (e.g., `And` and `Or`) or `ValueExpression` (e.g., `Last` and `First`) often same the same set of fields, which means that the current basic `hashCode()` implementations generates collisions when their operands are the same.

A solution is to add some class-specific value to the hash calculation. Possible candidates are `getClass().hashCode()` and a class-specific constant.

jvdb:
Resolves #150.

Because we have many classes with the same set of fields (e.g., `Cho` <-> `Seq`, `And` <-> `Or`, `First` <-> `Last`, etc.), basing the value returned by `hashCode()` solely on an object's fields will cause many collisions. To remedy this, the value of `getClass().hashCode()` has been added to all overridden `hashCode()` calculations.

For some classes this may be unnecessary (e.g., `DataExpressionSource`, `ParseGraph`, etc.), but to prevent us from having to reexamine `hashCode()` implementations whenever some subclass is added, the call has been added to all instances.

### DIFF
--- a/core/src/main/java/io/parsingdata/metal/expression/True.java
+++ b/core/src/main/java/io/parsingdata/metal/expression/True.java
@@ -43,7 +43,7 @@ public class True implements Expression {
 
     @Override
     public int hashCode() {
-        return Objects.hash(true);
+        return Objects.hash(getClass().hashCode(), true);
     }
 
 }
--- a/core/src/main/java/io/parsingdata/metal/expression/True.java
+++ b/core/src/main/java/io/parsingdata/metal/expression/True.java
@@ -16,8 +16,6 @@
 
 package io.parsingdata.metal.expression;
 
-import java.util.Objects;
-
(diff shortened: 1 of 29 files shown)


core/src/main/java/io/parsingdata/metal/expression/True.java
```diff
@@ -43,7 +43,7 @@ public boolean equals(final Object obj) {
 
     @Override
     public int hashCode() {
-        return Objects.hash(true);
+        return Objects.hash(getClass().hashCode(), true);
```
rdvdijk: As discussed offline: only use `getClass().hashCode()`, omit `true`.

PR #198, opened 2017-07-25
Improved and unified toString() implementations, added some missing ones

### LINKED ISSUE #132: Review and improve toString() implementations
`toString()` implementations often don't contain all relevant information (such as their own name) but sometimes do run the risk of being unwieldy with large amounts of data (such as printing their entire contents in hex...)

jvdb:
Resolves #132.

- Added missing `toString()` implementations for `Fold` (and by extension `FoldLeft` and `FoldRight`) and `Reverse`.
- Added class name to `toString()` implementations where missing (`ByteStreamSource`, `ConstantSource`, `DataExpressionSource`, `Environment` and `Slice`), with some small exceptions:
  - `ImmutableList` is such as basic data structure that it doesn't mention its class name at all.
  - `ParseGraph`, `ParseReference` and `ParseValue` are so common that they are abbreviated to `pg`, `pref` and `pval` respectively (actually `ParseReference` is quite rare, but it's nice to have a uniform naming scheme).
- Removed spaces from `toString()` implementations to make the strings more compact.
- Unified printing of hex strings:
  - Prefix everywhere with "0x".
  - Maximum amount of printed bytes in a hex string is 11. When an array is bigger, only the first 10 bytes are printed followed by an ellipsis.

### DIFF
--- a/core/src/main/java/io/parsingdata/metal/Util.java
+++ b/core/src/main/java/io/parsingdata/metal/Util.java
@@ -64,17 +64,13 @@ public final class Util {
     public static String bytesToShortHexString(final byte[] bytes) {
         checkNotNull(bytes, "bytes");
         if (bytes.length > (MAX_VALUES_IN_HEX_STRING + 1)) {
-            final char[] hexChars = new char[(MAX_VALUES_IN_HEX_STRING * 2) + 3];
-            hexChars[hexChars.length-1] = '.';
-            hexChars[hexChars.length-2] = '.';
-            hexChars[hexChars.length-3] = '.';
-            return fillHexDigits(bytes, hexChars, MAX_VALUES_IN_HEX_STRING);
-        } else {
-            return fillHexDigits(bytes, new char[bytes.length * 2], bytes.length);
+            return fillHexDigits(bytes, MAX_VALUES_IN_HEX_STRING) + "...";
         }
+        return fillHexDigits(bytes, bytes.length);
     }
 
-    private static String fillHexDigits(final byte[] bytes, final char[] hexChars, final int count) {
+    private static String fillHexDigits(final byte[] bytes, final int count) {
+        char[] hexChars = new char[count * 2];
         for (int j = 0; j < count; j++) {
             final int v = bytes[j] & 0xFF;
             hexChars[j * 2] = HEX_ARRAY[v >>> 4];
--- a/core/src/main/java/io/parsingdata/metal/Util.java
+++ b/core/src/main/java/io/parsingdata/metal/Util.java
@@ -34,6 +34,8 @@ public final class Util {
 
     private Util() {}
 
+    final public static int MAX_VALUES_IN_HEX_STRING = 10;
+
     final private static char[] HEX_ARRAY = "0123456789ABCDEF".toCharArray(); // Private because array content is mutable.
 
     public static <T>T checkNotNull(final T argument, final String name) {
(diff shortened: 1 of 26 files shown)


core/src/main/java/io/parsingdata/metal/Util.java
```diff
@@ -59,10 +61,21 @@ public static boolean notNullAndSameClass(final Object object, final Object othe
...
     }
 
-    public static String bytesToHexString(final byte[] bytes) {
+    public static String bytesToShortHexString(final byte[] bytes) {
         checkNotNull(bytes, "bytes");
-        final char[] hexChars = new char[bytes.length * 2];
-        for (int j = 0; j < bytes.length; j++) {
+        if (bytes.length > (MAX_VALUES_IN_HEX_STRING + 1)) {
+            final char[] hexChars = new char[(MAX_VALUES_IN_HEX_STRING * 2) + 3];
+            hexChars[hexChars.length-1] = '.';
```
mvanaken: I would include the additional dots with `+ "..."` after the return, instead of these single sets in the array. This method can be simplified further, since you would not need a hexChar array for the fillHexDigits, only the count:

````
if (bytes.length > (MAX_VALUES_IN_HEX_STRING + 1)) {
    return fillHexDigits(bytes, MAX_VALUES_IN_HEX_STRING) + "...";
}
return fillHexDigits(bytes, bytes.length);
````

File: core/src/main/java/io/parsingdata/metal/Util.java [hunk 2]
```diff
@@ -59,10 +61,21 @@ public static boolean notNullAndSameClass(final Object object, final Object othe
             && object.getClass() == other.getClass();
     }
 
-    public static String bytesToHexString(final byte[] bytes) {
+    public static String bytesToShortHexString(final byte[] bytes) {
```
mvanaken: Maybe add some javadoc, to explain the ellipsis principle? ;)

core/src/main/java/io/parsingdata/metal/data/DataExpressionSource.java
```diff
@@ -63,7 +63,7 @@ public DataExpressionSource(final ValueExpression dataExpression, final int inde
 
     @Override
     public String toString() {
-        return dataExpression.toString() + "[" + index + "](" + graph + "," + encoding + ")";
+        return getClass().getSimpleName() + "(" + dataExpression.toString() + "[" + index + "](" + graph + "," + encoding + "))";
```
mvanaken: With large string concatenations (> 5 * `+`) I tend to use String.formats, such that you see the overal idea of the string:
````java
String.format("%s(%s[%d](%s,%s))", getClass.getSimpleName(), dataExpression.toString(), index, graph, encoding);
````
But that is very personal, because I also understand that it is harder to see which `%s` matches with which string.

File: core/src/main/java/io/parsingdata/metal/data/ParseGraph.java [hunk 4]
```diff
@@ -119,11 +119,11 @@ ParseGraph closeBranch() {
 
     @Override
     public String toString() {
-        if (this == EMPTY) { return "graph(EMPTY)"; }
+        if (this == EMPTY) { return "pg(EMPTY)"; }
```
mvanaken: Is the `p` really needed? Could be shorter, so instead `pg`, `pref`, `pval`, maybe,  `gr`, `ref`, `val`?

core/src/main/java/io/parsingdata/metal/Util.java
jvdb: With the suggested changes, the code is now trivial! :)

core/src/main/java/io/parsingdata/metal/data/DataExpressionSource.java
jvdb: I agree, also because of performance concerns around this plain string concatenation. But I do propose we do that some other time, as it's not an important change (doesn't affect runtime performance). And about legibility... I don't know.

core/src/main/java/io/parsingdata/metal/data/ParseGraph.java
jvdb: I personally always explicitly think about *parse*graph, *parse*value, etc. so I think the _p_ makes it stand out, but that may be personal.

PR #213, opened 2017-10-03
Improved Slice creation interface

### LINKED ISSUE #210: Change slice()/isAvailable() on Environment
Based on original comment on #208 by @mvanaken: https://github.com/parsingdata/metal/pull/208#discussion_r140179314.

Instead of having `isAvailable()` on `Environment`, let `slice()` return an `Optional`.

jvdb:
Resolves #210.

Instead of having `isAvailable()` also on `Environment`, the 'slice()` method now returns an `Optional<Slice>` instead of just a `Slice`. Furthermore the location and visibility of methods has been improved. More can be done but I believe this is a good step.

### DIFF
--- a/core/src/main/java/io/parsingdata/metal/data/ByteStreamSource.java
+++ b/core/src/main/java/io/parsingdata/metal/data/ByteStreamSource.java
@@ -19,12 +19,13 @@ package io.parsingdata.metal.data;
 import static io.parsingdata.metal.Util.checkNotNull;
 
 import java.io.IOException;
+import java.io.UncheckedIOException;
 import java.math.BigInteger;
 import java.util.Objects;
 
 import io.parsingdata.metal.Util;
 
-public class ByteStreamSource extends Source {
+public class ByteStreamSource implements Source {
 
     public final ByteStream input;
 
@@ -33,9 +34,13 @@ public class ByteStreamSource extends Source {
     }
 
     @Override
-    protected byte[] getData(final long offset, final BigInteger length) throws IOException {
-        if (!isAvailable(offset, length)) { throw new IOException("Data to read is not available ([offset=" + offset + ";length=" + length + ";source=" + this + ")."); }
-        return input.read(offset, length.intValue());
+    public byte[] getData(final long offset, final BigInteger length) {
+        if (!isAvailable(offset, length)) { throw new RuntimeException("Data to read is not available ([offset=" + offset + ";length=" + length + ";source=" + this + ")."); }
+        try {
+            return input.read(offset, length.intValue());
+        } catch (final IOException exception) {
+            throw new UncheckedIOException(exception);
+        }
     }
 
     @Override
--- a/core/src/main/java/io/parsingdata/metal/data/ByteStreamSource.java
+++ b/core/src/main/java/io/parsingdata/metal/data/ByteStreamSource.java
@@ -25,7 +25,7 @@ import java.util.Objects;
 
 import io.parsingdata.metal.Util;
 
-public class ByteStreamSource implements Source {
+public class ByteStreamSource extends Source {
 
     public final ByteStream input;
 
@@ -34,7 +34,7 @@ public class ByteStreamSource implements Source {
     }
 
     @Override
-    public byte[] getData(final long offset, final BigInteger length) {
+    protected byte[] getData(final long offset, final BigInteger length) {
         if (!isAvailable(offset, length)) { throw new RuntimeException("Data to read is not available ([offset=" + offset + ";length=" + length + ";source=" + this + ")."); }
         try {
             return input.read(offset, length.intValue());
@@ -44,7 +44,7 @@ public class ByteStreamSource implements Source {
     }
 
     @Override
-    public boolean isAvailable(final long offset, final BigInteger length) {
+    protected boolean isAvailable(final long offset, final BigInteger length) {
         return input.isAvailable(offset, length.intValue());
     }
 
--- a/core/src/main/java/io/parsingdata/metal/data/ByteStreamSource.java
+++ b/core/src/main/java/io/parsingdata/metal/data/ByteStreamSource.java
@@ -35,7 +35,7 @@ public class ByteStreamSource extends Source {
 
     @Override
     protected byte[] getData(final long offset, final BigInteger length) {
-        if (!isAvailable(offset, length)) { throw new RuntimeException("Data to read is not available ([offset=" + offset + ";length=" + length + ";source=" + this + ")."); }
+        if (!isAvailable(offset, length)) { throw new IllegalStateException("Data to read is not available ([offset=" + offset + ";length=" + length + ";source=" + this + ")."); }
         try {
             return input.read(offset, length.intValue());
         } catch (final IOException exception) {
--- a/core/src/main/java/io/parsingdata/metal/data/ByteStreamSource.java
+++ b/core/src/main/java/io/parsingdata/metal/data/ByteStreamSource.java
@@ -29,7 +29,7 @@ public class ByteStreamSource extends Source {
 
     public final ByteStream input;
 
-    public ByteStreamSource(final ByteStream input) {
+    ByteStreamSource(final ByteStream input) {
         this.input = checkNotNull(input, "input");
     }
 
--- a/core/src/main/java/io/parsingdata/metal/data/ConstantSource.java
+++ b/core/src/main/java/io/parsingdata/metal/data/ConstantSource.java
@@ -26,7 +26,7 @@ import java.util.Objects;
 
 import io.parsingdata.metal.Util;
 
-public class ConstantSource extends Source {
+public class ConstantSource implements Source {
 
     private final byte[] data; // Private because array content is mutable.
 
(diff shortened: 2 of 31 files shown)


core/src/main/java/io/parsingdata/metal/data/Source.java
```diff
@@ -16,16 +16,11 @@
 
 package io.parsingdata.metal.data;
 
-import java.io.IOException;
 import java.math.BigInteger;
 
 public abstract class Source {
```
rdvdijk: I'm still a bit conflicted about this `abstract class`. The only reason that it is `abstract`, and not an `interface`, is that we can make the methods `protected`, correct?

I'd still prefer to use `interface` for something that is clearly an `interface`.

core/src/main/java/io/parsingdata/metal/token/Def.java
```diff
@@ -64,10 +64,10 @@ public Def(final String name, final ValueExpression size, final Encoding encodin
...
             return success(environment);
         }
-        if (!environment.isAvailable(dataSize)) {
-            return failure();
-        }
-        return success(environment.add(new ParseValue(scope, this, environment.slice(dataSize), encoding)).seek(dataSize.add(BigInteger.valueOf(environment.offset)).longValue()));
+        return environment
+            .slice(dataSize)
+            .map(slice -> success(environment.add(new ParseValue(scope, this, slice, encoding)).seek(dataSize.add(BigInteger.valueOf(environment.offset)).longValue())))
+            .orElse(failure());
```
rdvdijk: Brilliant.

core/src/main/java/io/parsingdata/metal/token/Until.java
```diff
@@ -99,9 +99,15 @@ public Until(final String name, final ValueExpression initialSize, final ValueEx
...
-        if (!environment.isAvailable(currentSize)) { return complete(Util::failure); }
-        final Slice slice = environment.slice(currentSize);
-        return terminator.parse(scope, currentSize.compareTo(ZERO) == 0 ? environment : environment.add(new ParseValue(name, this, slice, encoding)).seek(environment.offset + currentSize.longValue()), encoding)
+        return environment
+            .slice(currentSize)
+            .map(slice -> parseSlice(scope, environment, currentSize, stepSize, maxSize, encoding, slice))
+            .orElse(complete(Util::failure));
+    }
+
+    private Trampoline<Optional<Environment>> parseSlice(String scope, Environment environment, BigInteger currentSize, BigInteger stepSize, BigInteger maxSize, Encoding encoding, Slice slice) {
```
rdvdijk: 7 arguments... :stuck_out_tongue_winking_eye:

core/src/main/java/io/parsingdata/metal/data/ByteStreamSource.java
```diff
@@ -28,18 +29,22 @@
...
+    ByteStreamSource(final ByteStream input) {
         this.input = checkNotNull(input, "input");
     }
 
     @Override
-    protected byte[] getData(final long offset, final BigInteger length) throws IOException {
-        if (!isAvailable(offset, length)) { throw new IOException("Data to read is not available ([offset=" + offset + ";length=" + length + ";source=" + this + ")."); }
-        return input.read(offset, length.intValue());
+    protected byte[] getData(final long offset, final BigInteger length) {
+        if (!isAvailable(offset, length)) { throw new RuntimeException("Data to read is not available ([offset=" + offset + ";length=" + length + ";source=" + this + ")."); }
```
rdvdijk: I'd rather not use plain `RuntimeException`s. In this case `IllegalStateException` is appropriate, I think?

core/src/main/java/io/parsingdata/metal/data/ConstantSource.java
```diff
@@ -35,15 +35,15 @@ public ConstantSource(final byte[] data) {
     }
 
     @Override
-    protected byte[] getData(final long offset, final BigInteger length) throws IOException {
-        if (!isAvailable(offset, length)) { throw new IOException("Data to read is not available ([offset=" + offset + ";length=" + length + ";source=" + this + ")."); }
+    protected byte[] getData(final long offset, final BigInteger length) {
+        if (!isAvailable(offset, length)) { throw new RuntimeException("Data to read is not available ([offset=" + offset + ";length=" + length + ";source=" + this + ")."); }
```
rdvdijk: (See my other comment about using `RuntimeException`.)

core/src/main/java/io/parsingdata/metal/data/Slice.java
```diff
@@ -31,24 +34,29 @@
...
         this.offset = offset;
         this.length = checkNotNull(length, "length");
     }
 
+    public static Optional<Slice> createFromSource(final Source source, final long offset, final BigInteger length) {
+        if (length.compareTo(ZERO) < 0 || !source.isAvailable(offset, length)) { return Optional.empty(); }
+        return Optional.of(new Slice(source, offset, length));
+    }
+
+    public static Slice createFromBytes(final byte[] data) {
```
rdvdijk: Add a `null` check?

core/src/main/java/io/parsingdata/metal/data/Slice.java
```diff
@@ -31,24 +34,29 @@
...
     public final BigInteger length;
 
-    public Slice(final Source source, final long offset, final BigInteger length) {
+    private Slice(final Source source, final long offset, final BigInteger length) {
         this.source = checkNotNull(source, "source");
         this.offset = offset;
         this.length = checkNotNull(length, "length");
     }
 
+    public static Optional<Slice> createFromSource(final Source source, final long offset, final BigInteger length) {
```
rdvdijk: Does this factory method need argument checking?

core/src/main/java/io/parsingdata/metal/data/Slice.java
```diff
@@ -31,24 +34,29 @@
...
+        if (length.compareTo(ZERO) < 0 || !source.isAvailable(offset, length)) { return Optional.empty(); }
+        return Optional.of(new Slice(source, offset, length));
+    }
+
+    public static Slice createFromBytes(final byte[] data) {
+        return new Slice(new ConstantSource(data), 0, BigInteger.valueOf(data.length));
```
rdvdijk: You *could* call `createFromSource` here...

core/src/main/java/io/parsingdata/metal/data/Source.java
```diff
@@ -16,16 +16,11 @@
...
 
-    Slice slice(final long offset, final BigInteger length) {
-        return new Slice(this, offset, length);
-    }
```
rdvdijk: (For future reference:)

To me it seems intuitive to think that `Source`s create their own slices. Can you explain why it is not the responsibility of `Source` anymore to create `Slice`s?

core/src/main/java/io/parsingdata/metal/data/Slice.java
jvdb: Agreed, but I don't want to have this dirty automatic `.get()` added to it at several places.

core/src/main/java/io/parsingdata/metal/data/Source.java
jvdb: I would say that if a `Source` could be split into partial sources, then I agree completely. But in this case, we are _creating_ a `Slice`, so that is a more natural fit as a public factory method on `Slice` itself.

You could argue that given their relationship, perhaps `Slice` should be nested inside `Source`.

core/src/main/java/io/parsingdata/metal/data/Source.java
jvdb: I have an intuitive preference for interface as well. I think that if we end up not using the visibility effects for anything of real practical value, it will be better to convert it to an interface.

core/src/main/java/io/parsingdata/metal/data/ByteStreamSource.java
```diff
@@ -28,18 +29,22 @@
...
-    public ByteStreamSource(final ByteStream input) {
+    ByteStreamSource(final ByteStream input) {
         this.input = checkNotNull(input, "input");
     }
 
     @Override
-    protected byte[] getData(final long offset, final BigInteger length) throws IOException {
-        if (!isAvailable(offset, length)) { throw new IOException("Data to read is not available ([offset=" + offset + ";length=" + length + ";source=" + this + ")."); }
-        return input.read(offset, length.intValue());
+    protected byte[] getData(final long offset, final BigInteger length) {
```
ccreeten: Just for information, as you want this Metal implementation to be functional/immutable and strive for correctness, should we then not check these arguments as well, in order to more easily spot possible mistake (e.g., if ever a negative length is passed here?)

core/src/main/java/io/parsingdata/metal/data/Slice.java
ccreeten: Is a negative offset valid? I am a bit confused, perhaps isAvailable handles those cases (I'd prefer to fail early), but you do  specifically check for the length here.

core/src/test/java/io/parsingdata/metal/data/ConstantSliceTest.java
```diff
@@ -45,7 +45,7 @@ public void nullInput() {
     public void readBeyondSourceSize() throws IOException {
         final byte[] input = { 1, 2, 3, 4 };
         final Slice slice = createFromBytes(input);
-        thrown.expect(IOException.class);
+        thrown.expect(RuntimeException.class);
```
ccreeten: You expect any runtime exception?

core/src/test/java/io/parsingdata/metal/data/SourceAndSliceTest.java
```diff
@@ -89,26 +88,24 @@ private boolean compareDataSlices(byte[] data, int offset) {
 
     @Test
     public void readBeyondEndOfSource() throws IOException {
-        thrown.expect(IOException.class);
+        thrown.expect(RuntimeException.class);
```
ccreeten: Same here and down below, this tests for a whole slew of exceptions which hopefully will never be thrown.

PR #214, opened 2017-10-06
Converted offset to BigInteger

### LINKED ISSUE #16: Support arbitrary precision integer values on all arithmetic and addressing operations

jvdb:
Resolves #16.

The only places where we still use `long` or `int` are the following:

1. The `size` field of `ImmutableList` and `ParseGraph`.
2. When dealing with `byte[]`, which means there is a hard limit on what fits in an `int` anyway. For example: the `length` argument on methods of the `ByteStream` interface. Since `getData()` returns a `byte[]` anyway, there's no point in changing this now.
3. Many internal values that will remain within `long` scope I think (such as calculated result of `RepN`, result of `Count`, etc.) These values can be `BigInteger` and they are converted when necessary, but I don't think it actually makes sense to convert them now.

### DIFF
--- a/core/src/main/java/io/parsingdata/metal/data/ConstantSource.java
+++ b/core/src/main/java/io/parsingdata/metal/data/ConstantSource.java
@@ -34,16 +34,16 @@ public class ConstantSource extends Source {
     }
 
     @Override
-    protected byte[] getData(final long offset, final BigInteger length) {
+    protected byte[] getData(final BigInteger offset, final BigInteger length) {
         if (!isAvailable(offset, length)) { throw new IllegalStateException("Data to read is not available ([offset=" + offset + ";length=" + length + ";source=" + this + ")."); }
         final byte[] outputData = new byte[length.intValue()];
-        System.arraycopy(data, (int)offset, outputData, 0, outputData.length);
+        System.arraycopy(data, offset.intValue(), outputData, 0, outputData.length);
         return outputData;
     }
 
     @Override
-    protected boolean isAvailable(final long offset, final BigInteger length) {
-        return length.intValue() + offset <= data.length;
+    protected boolean isAvailable(final BigInteger offset, final BigInteger length) {
+        return length.add(offset).compareTo(BigInteger.valueOf(data.length)) <= 0;
     }
 
     @Override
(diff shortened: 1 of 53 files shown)


core/src/main/java/io/parsingdata/metal/data/ConstantSource.java
```diff
@@ -34,16 +34,16 @@ public ConstantSource(final byte[] data) {
     }
 
     @Override
-    protected byte[] getData(final long offset, final BigInteger length) {
+    protected byte[] getData(final BigInteger offset, final BigInteger length) {
         if (!isAvailable(offset, length)) { throw new IllegalStateException("Data to read is not available ([offset=" + offset + ";length=" + length + ";source=" + this + ")."); }
         final byte[] outputData = new byte[length.intValue()];
-        System.arraycopy(data, (int)offset, outputData, 0, outputData.length);
+        System.arraycopy(data, offset.intValue(), outputData, 0, outputData.length);
```
rdvdijk: This isn't a problem here, but for future reference, a quote from the `BigInteger#intValue()` JavaDoc:

> if this BigInteger is too big to fit in an int, only the low-order 32 bits are returned

core/src/main/java/io/parsingdata/metal/data/Slice.java
```diff
@@ -29,22 +29,22 @@
...
 
-    public static Optional<Slice> createFromSource(final Source source, final long offset, final BigInteger length) {
+    public static Optional<Slice> createFromSource(final Source source, final BigInteger offset, final BigInteger length) {
         if (checkNotNull(length, "length").compareTo(ZERO) < 0 || !checkNotNull(source, "source").isAvailable(offset, length)) { return Optional.empty(); }
         return Optional.of(new Slice(source, offset, length));
     }
 
     public static Slice createFromBytes(final byte[] data) {
-        return new Slice(new ConstantSource(checkNotNull(data, "data")), 0, BigInteger.valueOf(data.length));
+        return new Slice(new ConstantSource(checkNotNull(data, "data")), BigInteger.ZERO, BigInteger.valueOf(data.length));
```
rdvdijk: Elsewhere you've statically imported `BigInteger.ZERO` and `BigInteger.ONE`. Can you do it everywhere? (Also in throughout the test code.)

core/src/main/java/io/parsingdata/metal/data/ConstantSource.java
jvdb: As discussed, replaced all invocations of `byteValue()`, `intValue()` and `longValue()` with their `*Exact()` counterparts.

PR #215, opened 2017-10-08
Handle error cases caused by negative offsets

### LINKED ISSUE #207: Handle negative offsets
This is relevant for the `Sub` `Token`, but also more generally for seek operations on `Slice` and `Source`.

jvdb:
Resolves #207.

This PR solves two different problems both related to the handling of negative offsets.

First is preventing the creation of an `Environment` with an `offset` field that holds a negative value. Some code reading revealed that this could occur through a call to the `.seek()` method or through invoking the top-level constructor, both by passing in a negative `offset` argument. Both cases have been fixed. This involved some tweaking of `Def`, `Sub` and `Until` to make them correctly handle error cases (instead of causing an exception to be thrown).

One issue to consider is that the `Environment` constructor now throws an exception while the `.seek()` method returns an `Optional.empty()`. I believe this is warranted because the former is a constructor. However, we can still fix this by changing the `Environment` class to have a private constructor and then handle creation of `Environment` objects through (a) factory method(s).

Second is no longer allowing a negative `offset` argument to the implementations of `Source.getData()`, as well as not allowing the creation of a `Slice` that starts at a negative offset.

### DIFF
--- a/core/src/main/java/io/parsingdata/metal/Util.java
+++ b/core/src/main/java/io/parsingdata/metal/Util.java
@@ -16,7 +16,10 @@
 
 package io.parsingdata.metal;
 
+import static java.math.BigInteger.ZERO;
+
 import java.io.ByteArrayOutputStream;
+import java.math.BigInteger;
 import java.util.Optional;
 import java.util.zip.DataFormatException;
 import java.util.zip.Inflater;
@@ -56,6 +59,11 @@ public final class Util {
     public static boolean notNullAndSameClass(final Object object, final Object other) {
         return other != null
             && object.getClass() == other.getClass();
+    }
+
+    public static BigInteger checkNotNegative(final BigInteger argument, final String name) {
+        if (checkNotNull(argument, name).compareTo(ZERO) < 0) { throw new IllegalArgumentException("Argument " + name + " may not be negative."); }
+        return argument;
     }
 
     public static String bytesToHexString(final byte[] bytes) {
--- a/core/src/main/java/io/parsingdata/metal/data/DataExpressionSource.java
+++ b/core/src/main/java/io/parsingdata/metal/data/DataExpressionSource.java
@@ -20,6 +20,7 @@ import static java.math.BigInteger.ZERO;
 
 import static io.parsingdata.metal.Trampoline.complete;
 import static io.parsingdata.metal.Trampoline.intermediate;
+import static io.parsingdata.metal.Util.checkNotNegative;
 import static io.parsingdata.metal.Util.checkNotNull;
 
 import java.math.BigInteger;
@@ -48,8 +49,9 @@ public class DataExpressionSource extends Source {
 
     @Override
     protected byte[] getData(final BigInteger offset, final BigInteger length) {
+        checkNotNegative(offset, "offset");
         final Value inputValue = getValue();
-        if (offset.compareTo(ZERO) < 0 || length.add(offset).compareTo(inputValue.slice.length) > 0) { throw new IllegalStateException("Data to read is not available ([offset=" + offset + ";length=" + length + ";source=" + this + ")."); }
+        if (length.add(offset).compareTo(inputValue.slice.length) > 0) { throw new IllegalStateException("Data to read is not available ([offset=" + offset + ";length=" + length + ";source=" + this + ")."); }
         final byte[] outputData = new byte[length.intValueExact()];
         System.arraycopy(inputValue.getValue(), offset.intValueExact(), outputData, 0, outputData.length);
         return outputData;
--- a/core/src/main/java/io/parsingdata/metal/data/DataExpressionSource.java
+++ b/core/src/main/java/io/parsingdata/metal/data/DataExpressionSource.java
@@ -16,6 +16,8 @@
 
 package io.parsingdata.metal.data;
 
+import static java.math.BigInteger.ZERO;
+
 import static io.parsingdata.metal.Trampoline.complete;
 import static io.parsingdata.metal.Trampoline.intermediate;
 import static io.parsingdata.metal.Util.checkNotNull;
@@ -47,7 +49,7 @@ public class DataExpressionSource extends Source {
     @Override
     protected byte[] getData(final BigInteger offset, final BigInteger length) {
         final Value inputValue = getValue();
-        if (length.add(offset).compareTo(inputValue.slice.length) > 0) { throw new IllegalStateException("Data to read is not available ([offset=" + offset + ";length=" + length + ";source=" + this + ")."); }
+        if (offset.compareTo(ZERO) < 0 || length.add(offset).compareTo(inputValue.slice.length) > 0) { throw new IllegalStateException("Data to read is not available ([offset=" + offset + ";length=" + length + ";source=" + this + ")."); }
         final byte[] outputData = new byte[length.intValueExact()];
         System.arraycopy(inputValue.getValue(), offset.intValueExact(), outputData, 0, outputData.length);
         return outputData;
--- a/core/src/main/java/io/parsingdata/metal/token/Def.java
+++ b/core/src/main/java/io/parsingdata/metal/token/Def.java
@@ -66,7 +66,7 @@ public class Def extends Token {
         }
         return environment
             .slice(dataSize)
-            .map(slice -> success(environment.add(new ParseValue(scope, this, slice, encoding)).seek(dataSize.add(environment.offset))))
+            .map(slice -> environment.add(new ParseValue(scope, this, slice, encoding)).seek(dataSize.add(environment.offset)))
             .orElse(failure());
     }
 
--- a/core/src/main/java/io/parsingdata/metal/token/Sub.java
+++ b/core/src/main/java/io/parsingdata/metal/token/Sub.java
@@ -68,7 +68,7 @@ public class Sub extends Token {
             return failure();
         }
         return iterate(scope, addresses, environment.addBranch(this), encoding).computeResult()
-            .flatMap(nextEnvironment -> success(nextEnvironment.seek(environment.offset)));
+            .flatMap(nextEnvironment -> nextEnvironment.seek(environment.offset));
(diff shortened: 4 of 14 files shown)


core/src/main/java/io/parsingdata/metal/Util.java
```diff
@@ -58,6 +61,11 @@ public static boolean notNullAndSameClass(final Object object, final Object othe
             && object.getClass() == other.getClass();
     }
 
+    public static BigInteger checkNotNegative(final BigInteger argument, final String name) {
+        if (checkNotNull(argument, name).compareTo(ZERO) < 0) { throw new IllegalArgumentException("Argument " + name + " may not be negative."); }
```
rdvdijk: ..`, but was: " + argument`.

core/src/main/java/io/parsingdata/metal/token/Def.java
```diff
@@ -66,7 +66,7 @@ public Def(final String name, final ValueExpression size, final Encoding encodin
         }
         return environment
             .slice(dataSize)
-            .map(slice -> success(environment.add(new ParseValue(scope, this, slice, encoding)).seek(dataSize.add(environment.offset))))
+            .map(slice -> environment.add(new ParseValue(scope, this, slice, encoding)).seek(dataSize.add(environment.offset)))
```
rdvdijk: Wait, what? How come this `success` can be safely deleted here?

core/src/main/java/io/parsingdata/metal/token/Sub.java
```diff
@@ -87,7 +87,10 @@ public Sub(final String name, final Token token, final ValueExpression address,
         if (hasRootAtOffset(environment.order, token.getCanonical(environment), offset, environment.source)) {
             return success(environment.add(new ParseReference(offset, environment.source, token.getCanonical(environment))));
         }
-        return token.parse(scope, environment.seek(offset), encoding);
+        return environment
+            .seek(offset)
+            .map(newEnvironment -> token.parse(scope, newEnvironment, encoding))
+            .orElse(failure());
```
rdvdijk: Even though this is a bit longer than before, I like it!

core/src/main/java/io/parsingdata/metal/token/Until.java
```diff
@@ -106,9 +107,10 @@ public Until(final String name, final ValueExpression initialSize, final ValueEx
     }
 
     private Trampoline<Optional<Environment>> parseSlice(String scope, Environment environment, BigInteger currentSize, BigInteger stepSize, BigInteger maxSize, Encoding encoding, Slice slice) {
-        return terminator
-            .parse(scope, currentSize.compareTo(ZERO) == 0 ? environment : environment.add(new ParseValue(name, this, slice, encoding)).seek(environment.offset.add(currentSize)), encoding)
-            .map(nextEnvironment -> complete(() -> success(nextEnvironment)))
+        return (currentSize.compareTo(ZERO) == 0 ? Optional.of(environment) : environment.add(new ParseValue(name, this, slice, encoding)).seek(environment.offset.add(currentSize)))
```
rdvdijk: I'd prefer if you just turn the zero-check into a guard clause here, instead of "cramming" it into the single statement. (I know you're a fan... 😁) The creation of the optional environment seems a bit _forced_.

core/src/test/java/io/parsingdata/metal/ErrorsTest.java
```diff
@@ -71,7 +73,17 @@ public void multiValueInRepN() throws IOException {
...
+    }
+
+    @Test
+    public void environmentWithNegativeOffset() {
+        thrown.expect(IllegalArgumentException.class);
+        thrown.expectMessage("Argument offset may not be negative.");
+        new Environment(new ByteStream() {
+            @Override public byte[] read(BigInteger offset, int length) throws IOException { return new byte[0]; }
+            @Override public boolean isAvailable(BigInteger offset, int length) { return false; }
+        }, BigInteger.valueOf(-1));
```
rdvdijk: Why do we need a new (inline) `ByteStream` implementation here? Don't we have one handy inside the test scope already?

core/src/test/java/io/parsingdata/metal/data/SourceAndSliceTest.java
```diff
@@ -111,4 +111,15 @@ public void startReadBeyondEndOfSlice() {
         assertFalse(Slice.createFromSource(source, BigInteger.valueOf(5), ZERO).isPresent());
     }
 
+    @Test
+    public void startReadAtNegativeOffsetSource() {
+        thrown.expect(IllegalStateException.class);
```
rdvdijk: `thrown.expectMessage("....")`

core/src/test/java/io/parsingdata/metal/data/SourceAndSliceTest.java
```diff
@@ -111,4 +111,15 @@ public void startReadBeyondEndOfSlice() {
...
+        source.getData(BigInteger.valueOf(-1L), ONE);
+    }
+
+    @Test
+    public void startReadAtNegativeOffsetSlice() {
+        assertFalse(Slice.createFromSource(source, BigInteger.valueOf(-1L), ONE).isPresent());
```
rdvdijk: (We really need to get you into Hamcrest one of these days... 😉)

core/src/main/java/io/parsingdata/metal/data/DataExpressionSource.java
```diff
@@ -47,7 +49,7 @@ public DataExpressionSource(final ValueExpression dataExpression, final int inde
     @Override
     protected byte[] getData(final BigInteger offset, final BigInteger length) {
         final Value inputValue = getValue();
-        if (length.add(offset).compareTo(inputValue.slice.length) > 0) { throw new IllegalStateException("Data to read is not available ([offset=" + offset + ";length=" + length + ";source=" + this + ")."); }
+        if (offset.compareTo(ZERO) < 0 || length.add(offset).compareTo(inputValue.slice.length) > 0) { throw new IllegalStateException("Data to read is not available ([offset=" + offset + ";length=" + length + ";source=" + this + ")."); }
```
rdvdijk: Call the new `checkNotNegative` here instead?

core/src/main/java/io/parsingdata/metal/Util.java
ccreeten: Perhaps you should create an ```UnsignedBigInteger``` or a ```PositiveBigInteger``` class. :laughing:

core/src/main/java/io/parsingdata/metal/data/DataExpressionSource.java
ccreeten: Same goes for the other source classes.

core/src/main/java/io/parsingdata/metal/token/Def.java
ccreeten: Because ```Environment#add``` now returns an optional environment.

ccreeten: Because ```Environment#seek``` now returns an optional environment.

core/src/main/java/io/parsingdata/metal/token/Until.java
ccreeten: Returns only, right?

core/src/main/java/io/parsingdata/metal/token/Def.java
jvdb: That would lead to a very redundant `.map(x -> x).orElse(empty())` construct... So I appreciate the sentiment but I don't really know what a good solution is...

core/src/test/java/io/parsingdata/metal/ErrorsTest.java
jvdb: Thing is, I don't want to pull in something like an `InMemoryByteStream` is all I want is a vanilla `ByteStream`. But maybe there's some other test version around...

core/src/main/java/io/parsingdata/metal/data/DataExpressionSource.java
jvdb: That would turn it into multiple guard clauses. But I agree it's probably better...

core/src/main/java/io/parsingdata/metal/token/Def.java
ccreeten: WIsn't something like the following possible:
```java
return environment
    .slice(dataSize)
    .map(slice -> environment.add(new ParseValue(scope, this, slice, encoding)))
    .map(env -> env.seek(dataSize.add(environment.offset)))
    .orElse(failure());
```
Not saying it is the better solution, but I think it is functionally possible? Perhaps a bit more clear?

core/src/main/java/io/parsingdata/metal/Util.java
jvdb: Good point but let's do this in a general review of all error messages.

core/src/main/java/io/parsingdata/metal/token/Def.java
jvdb: Sure, but then something like `Environment.add()` would have to return an `Optional` even though there's no real reason it could fail. I see the relative beauty of writing it like this, but implementing brings all kinds of trade-offs.

core/src/main/java/io/parsingdata/metal/token/Until.java
jvdb: Agreed that the `Optional.of(environment)` is dodgy. But it either explodes into a really verbose `if` (with `isPresent()`, `get()`, etc.) and even removes the first `map()` or it's like this...

core/src/main/java/io/parsingdata/metal/token/Def.java
ccreeten: Or I am overlooking something, because to me it seems like you are confusing ```map``` with ```flatMap``` :smile:

PR #217, opened 2017-10-16
Minor issues and gardening for upcoming 7.0.0 release

### LINKED ISSUE #158: Exclude incorrect FindBugs check on equals method
FindBugs complains about `BC_EQUALS_METHOD_SHOULD_WORK_FOR_ALL_OBJECTS` but this is incorrect: our `equals()` works on all objects, but the check is performed in the base class and FindBugs (understadably) doesn't detect this. So we should disable this check.

This can be done using an external xml file to control FindBugs, but it's probably nicer to just exclude the check entirely in the root pom. However, I haven't been able to find the exact name of the visitor in order to disable it.

For reference, here's an example from the [FindBugs Maven site](http://gleclaire.github.io/findbugs-maven-plugin/usage.html):

```xml
<configuration>
    <omitVisitors>FindDeadLocalStores,UnreadFields</omitVisitors>
</configuration>
```

So until this is fixed, the Maven output of building Metal reports 26 warnings, which are all caused by this issue.

### LINKED ISSUE #209: Methods on ParseGraph have inconsistent visibility
The methods `addBranch` and `closeBranch` are package-private. The methods `add(ParseValue)` and `add(ParseRef)` are `public`. Since only `Environment` ever make use of these methods, all of these methods can be made `package-private`.

The current test code that uses the `public` methods should not be affected if they are placed in the correct package.

### LINKED ISSUE #212: Add Travis build configuration for OracleJDK9

jvdb:
Resolves #212: Added OracleJDK9 to Travis build configuration for building and testing. Fixed an issue with retrieving resources on JDK9.
Resolves #158: FindBugs now excludes equals()-check that reports incorrect error.
Resolves #209: The `add(ParseValue)` and `add(ParseReference)` methods on `ParseGraph` are now package-private.
Minor refactoring to pom, updated versions of all dependencies to current releases.
Migrated use of `orElse()` to `orElseGet()`.
Improved formatting by moving all `return` statements to their own line.

### DIFF
--- a/core/src/main/java/io/parsingdata/metal/expression/comparison/Eq.java
+++ b/core/src/main/java/io/parsingdata/metal/expression/comparison/Eq.java
@@ -40,9 +40,13 @@ public class Eq extends ComparisonExpression {
     public boolean compare(final Value left, final Value right) {
         final byte[] leftBytes = left.getValue();
         final byte[] rightBytes = right.getValue();
-        if (leftBytes.length != rightBytes.length) { return false; }
+        if (leftBytes.length != rightBytes.length) {
+            return false;
+        }
         for (int i = 0; i < leftBytes.length; i++) {
-            if (leftBytes[i] != rightBytes[i]) { return false; }
+            if (leftBytes[i] != rightBytes[i]) {
+                return false;
+            }
         }
         return true;
     }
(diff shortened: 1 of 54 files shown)


core/src/main/java/io/parsingdata/metal/expression/comparison/Eq.java
```diff
@@ -40,9 +40,13 @@ public Eq(final ValueExpression value, final ValueExpression predicate) {
     public boolean compare(final Value left, final Value right) {
         final byte[] leftBytes = left.getValue();
```
ccreeten: You could  perhaps also now pre-check on slice size, so that when these do not match, no data is read?

PR #219, opened 2017-10-19
Callback handle without Optional

### LINKED ISSUE #218: Remove Optional type from argument in Callback.handle()

jvdb:
Resolves #218.
Investigated with @mvanaken ways to abstract and improve the handling of callbacks. The end result looks good and sets up for next refactoring (e.g., removing `Callbacks` from `Environment`).


core/src/main/java/io/parsingdata/metal/token/Token.java
```diff
@@ -64,7 +66,9 @@ protected Token(final String name, final Encoding encoding) {
     public Optional<Environment> parse(final String scope, final Environment environment, final Encoding encoding) {
         final Encoding activeEncoding = this.encoding != null ? this.encoding : encoding;
         final Optional<Environment> result = parseImpl(makeScope(checkNotNull(scope, "scope")), checkNotNull(environment, "environment"), activeEncoding);
-        environment.callbacks.handle(this, environment, result);
+        environment.callbacks.handle(this, result
+            .map(after -> success(this, environment, after))
```
ccreeten: Oh, sneaky formatting :D But nice way of handling!

PR #220, opened 2017-10-20
Recursion in ParseGraph addBranch() and add()

gertjanal:
Added a test to prove that ParseGraph.addBranch and ParseGraph.add() are recursive and causing problems in current implementations. Interesting sidenote: the stack overflows did not occur in Metal 5.


jvdb:
This is a known issue, see #173.

jvdb:
Thanks for the test! I've referenced this PR in the related issue, so that we can recover the test when we address it.

PR #221, opened 2017-10-26
Improve abstraction of parse state

### LINKED ISSUE #216: Unexpected current offset behaviour while working with Sub
Something popped up while @TheUnknownCylon was working on a Metal definition:
```java
static Token FORMAT =
    seq(
        sub(def("any", 1), con(0xFF)),
        def("none", CURRENT_OFFSET)
    );
```
In this example, we first parse a value at offset ```0xFF``` with a length of one.  When parsing the second ```none``` value, using the current offset as length, we are outside of the sub. So in this case we would expect the length to be 0 (and no value to be parsed), yet it is ```0x100```. 

This is because the new ```CURRENT_OFFSET``` implementation uses the last added value to the graph, and the definition of it is ```offset + length``` of the latest value added to the ```ParseGraph```, which in our case is the ```any``` value.

This seems like it should not be the case?

jvdb:
Resolves #216.

This PR tackles some abstraction issues. The resolution resolves a functional bug. The six commits show the following six changes:

1. `Callbacks` was extracted from the `Environment` class.
2. The `Environment` class was renamed to `ParseState`.
3. Of the three remaining constructors on `ParseState`, two are renamed to factory methods to better reflect their functionality.
4. Both `equals()` and `hashCode()` methods were implemented on `ParseState`.
5. The `eval()` methods of `ValueExpression` and `Expression` were modified to accept a `ParseState` instead of a `ParseGraph`.
6. The `Shorthand` approximate implementation of `CURRENT_OFFSET` is replaced with an implementation that returns the actual `offset` field of the `ParseState`.

After that, another branch was merged that recreates an `Environment` class: #222.

### DIFF
--- a/core/src/main/java/io/parsingdata/metal/Shorthand.java
+++ b/core/src/main/java/io/parsingdata/metal/Shorthand.java
@@ -61,6 +61,7 @@ import io.parsingdata.metal.expression.value.arithmetic.Sub;
 import io.parsingdata.metal.expression.value.bitwise.ShiftLeft;
 import io.parsingdata.metal.expression.value.bitwise.ShiftRight;
 import io.parsingdata.metal.expression.value.reference.Count;
+import io.parsingdata.metal.expression.value.reference.CurrentOffset;
 import io.parsingdata.metal.expression.value.reference.First;
 import io.parsingdata.metal.expression.value.reference.Last;
 import io.parsingdata.metal.expression.value.reference.Len;
@@ -86,7 +87,7 @@ public final class Shorthand {
 
     public static final Token EMPTY = def(EMPTY_NAME, 0L);
     public static final ValueExpression SELF = new Self();
-    public static final ValueExpression CURRENT_OFFSET = elvis(add(offset(SELF), len(SELF)), con(0));
+    public static final ValueExpression CURRENT_OFFSET = new CurrentOffset();
     public static final Expression TRUE = new True();
 
     private Shorthand() {}
--- a/core/src/main/java/io/parsingdata/metal/data/DataExpressionSource.java
+++ b/core/src/main/java/io/parsingdata/metal/data/DataExpressionSource.java
@@ -35,13 +35,13 @@ public class DataExpressionSource extends Source {
 
     public final ValueExpression dataExpression;
     public final int index;
-    public final ParseGraph graph;
+    public final ParseState parseState;
     public final Encoding encoding;
 
-    public DataExpressionSource(final ValueExpression dataExpression, final int index, final ParseGraph graph, final Encoding encoding) {
+    public DataExpressionSource(final ValueExpression dataExpression, final int index, final ParseState parseState, final Encoding encoding) {
         this.dataExpression = checkNotNull(dataExpression, "dataExpression");
         this.index = index;
-        this.graph = checkNotNull(graph, "graph");
+        this.parseState = checkNotNull(parseState, "parseState");
         this.encoding = checkNotNull(encoding, "encoding");
     }
 
@@ -63,7 +63,7 @@ public class DataExpressionSource extends Source {
     }
 
     private Value getValue() {
-        final ImmutableList<Optional<Value>> results = dataExpression.eval(graph, encoding);
+        final ImmutableList<Optional<Value>> results = dataExpression.eval(parseState, encoding);
         if (results.size <= index) {
             throw new IllegalStateException("ValueExpression dataExpression yields " + results.size + " result(s) (expected at least " + (index + 1) + ").");
         }
@@ -81,7 +81,7 @@ public class DataExpressionSource extends Source {
 
     @Override
     public String toString() {
-        return getClass().getSimpleName() + "(" + dataExpression.toString() + "[" + index + "](" + graph + "," + encoding + "))";
+        return getClass().getSimpleName() + "(" + dataExpression.toString() + "[" + index + "](" + parseState + "," + encoding + "))";
     }
 
     @Override
@@ -89,13 +89,13 @@ public class DataExpressionSource extends Source {
         return Util.notNullAndSameClass(this, obj)
             && Objects.equals(dataExpression, ((DataExpressionSource)obj).dataExpression)
             && Objects.equals(index, ((DataExpressionSource)obj).index)
-            && Objects.equals(graph, ((DataExpressionSource)obj).graph)
+            && Objects.equals(parseState, ((DataExpressionSource)obj).parseState)
             && Objects.equals(encoding, ((DataExpressionSource)obj).encoding);
     }
 
     @Override
     public int hashCode() {
-        return Objects.hash(getClass().hashCode(), dataExpression, index, graph, encoding);
+        return Objects.hash(getClass().hashCode(), dataExpression, index, parseState, encoding);
     }
 
 }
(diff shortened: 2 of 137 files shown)


core/src/test/java/io/parsingdata/metal/ErrorsTest.java
```diff
@@ -73,15 +73,15 @@ public void multiValueInRepN() throws IOException {
                 any("b"),
                 repn(dummy, ref("b"))
             );
-       Optional<Environment> result = multiRepN.parse(stream(2, 2, 2, 2), enc());
+       Optional<ParseState> result = multiRepN.parse(stream(2, 2, 2, 2), enc());
```
mvanaken: Minor, but result can be made final.

core/src/main/java/io/parsingdata/metal/token/Tie.java
```diff
@@ -59,24 +60,24 @@ public Tie(final String name, final Token token, final ValueExpression dataExpre
...
         }
-        return iterate(scope, dataResult, 0, environment, environment.addBranch(this), encoding).computeResult();
+        return iterate(scope, dataResult, 0, parseState, parseState.addBranch(this), callbacks, encoding).computeResult();
     }
 
-    private Trampoline<Optional<Environment>> iterate(final String scope, final ImmutableList<Optional<Value>> values, final int index, final Environment returnEnvironment, final Environment environment, final Encoding encoding) {
+    private Trampoline<Optional<ParseState>> iterate(final String scope, final ImmutableList<Optional<Value>> values, final int index, final ParseState returnParseState, final ParseState parseState, final Callbacks callbacks, final Encoding encoding) {
         if (values.isEmpty()) {
-            return complete(() -> success(new Environment(environment.closeBranch().order, returnEnvironment.source, returnEnvironment.offset, returnEnvironment.callbacks)));
+            return complete(() -> success(new ParseState(parseState.closeBranch().order, returnParseState.source, returnParseState.offset)));
```
mvanaken: Not using a factory method here?

core/src/main/java/io/parsingdata/metal/data/ParseState.java
```diff
@@ -0,0 +1,100 @@
...
+import io.parsingdata.metal.expression.value.ValueExpression;
+import io.parsingdata.metal.token.Token;
+
+public class ParseState {
+
+    public final ParseGraph order;
+    public final BigInteger offset;
+    public final Source source;
+
+    public ParseState(final ParseGraph order, final Source source, final BigInteger offset) {
```
mvanaken: If it is possible you may consider making this ParseState constructor private and always use the ParseStateFactory instead.

core/src/main/java/io/parsingdata/metal/data/ParseState.java
```diff
@@ -0,0 +1,100 @@
...
+        this.order = checkNotNull(order, "order");
+        this.source = checkNotNull(source, "source");
+        this.offset = checkNotNegative(offset, "offset");
+    }
+
+    public static ParseState createFromByteStream(final ByteStream input, final BigInteger offset) {
```
mvanaken: Shouldn't this method be part of the ParseStateFactory?

core/src/main/java/io/parsingdata/metal/data/DataExpressionSource.java
```diff
@@ -81,21 +81,21 @@ private Value getValue() {
...
             && Objects.equals(index, ((DataExpressionSource)obj).index)
-            && Objects.equals(graph, ((DataExpressionSource)obj).graph)
+            && Objects.equals(parseState, ((DataExpressionSource)obj).parseState)
             && Objects.equals(encoding, ((DataExpressionSource)obj).encoding);
     }
 
     @Override
     public int hashCode() {
-        return Objects.hash(getClass().hashCode(), dataExpression, index, graph, encoding);
+        return Objects.hash(getClass().hashCode(), dataExpression, index, parseState, encoding);
```
ccreeten: You can probably remove the calls to hashcode everywhere where using ```Objects#hash```?

core/src/main/java/io/parsingdata/metal/Shorthand.java
```diff
@@ -86,7 +87,7 @@
 
     public static final Token EMPTY = def(EMPTY_NAME, 0L);
     public static final ValueExpression SELF = new Self();
-    public static final ValueExpression CURRENT_OFFSET = elvis(add(offset(SELF), len(SELF)), con(0));
+    public static final ValueExpression CURRENT_OFFSET = new CurrentOffset();
```
ccreeten: It's back with a vengeance!

core/src/main/java/io/parsingdata/metal/data/ParseState.java
```diff
@@ -0,0 +1,100 @@
...
+import java.math.BigInteger;
+import java.util.Objects;
+import java.util.Optional;
+
+import io.parsingdata.metal.Util;
+import io.parsingdata.metal.encoding.Encoding;
+import io.parsingdata.metal.expression.value.ValueExpression;
+import io.parsingdata.metal.token.Token;
+
+public class ParseState {
```
ccreeten: Perhaps make final?

ccreeten: Perhaps add some documentation :smiling_imp:

core/src/main/java/io/parsingdata/metal/data/ParseState.java
jvdb: As discussed, `ParseStateFactory` is a test class. I think this single constructor makes sense because it only stores fields.

core/src/main/java/io/parsingdata/metal/data/ParseState.java
jvdb: See above :)

core/src/main/java/io/parsingdata/metal/token/Tie.java
jvdb: I suppose that would have to be a method comparable to `seek()` that replaces the `ParseGraph` on `returnEnvironment`. It could be done, but this seems to be a rather isolated use.

core/src/test/java/io/parsingdata/metal/token/TieTest.java
```diff
@@ -76,25 +76,25 @@
...
-        final Optional<Environment> result = parseContainer();
-        checkFullParse(INC_PREV_MOD_100, fold(ref("value"), Shorthand::cat).eval(result.get().order, enc()).head.get().getValue());
+        final Optional<ParseState> result = parseContainer();
+        checkFullParse(INC_PREV_MOD_100, fold(ref("value"), Shorthand::cat).eval(result.get(), enc()).head.get().getValue());
     }
 
-    private Optional<Environment> checkFullParse(Token token, byte[] data) throws IOException {
-        final Optional<Environment> result = token.parse(new Environment(new InMemoryByteStream(data)), enc());
+    private Optional<ParseState> checkFullParse(Token token, byte[] data) throws IOException {
+        final Optional<ParseState> result = token.parse(createFromByteStream(new InMemoryByteStream(data)), enc());
```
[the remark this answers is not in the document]
jvdb: `stream()` takes a vararg, not an array :(

core/src/main/java/io/parsingdata/metal/data/DataExpressionSource.java
jvdb: It seems to get called on it anyway and all tests are green. Good catch!

core/src/main/java/io/parsingdata/metal/data/ParseState.java
jvdb: It's probably worth considering in general which classes should be made `final`: #223.

PR #222, opened 2017-10-27
Return of the Environment

jvdb:


core/src/main/java/io/parsingdata/metal/data/Environment.java
```diff
@@ -0,0 +1,63 @@
...
+import io.parsingdata.metal.encoding.Encoding;
+
+public class Environment {
+
+    public final String scope;
+    public final ParseState parseState;
+    public final Callbacks callbacks;
+    public final Encoding encoding;
+
+    public Environment(final String scope, final ParseState parseState, final Callbacks callbacks, final Encoding encoding) {
```
mvanaken: Just a thought: I think these constructors can be package private, but then the EnvironmentFactory must be moved to a different package (io.parsingdata.metal.data). I do think that a public Environment is undesired. At least there is/should be no need for users to create there own environment. Or am I missing an  possible scenario for which environment should be public?

jvdb: As discussed, the `EnvironmentFactory` is current a test class. So I won't make this change now -- but all the same it's a good discussion to perhaps make one as part of the production classes.

core/src/main/java/io/parsingdata/metal/token/Token.java
```diff
@@ -64,28 +64,16 @@ protected Token(final String name, final Encoding encoding) {
...
 
-    public Optional<ParseState> parse(final String scope, final ParseState parseState, final Callbacks callbacks, final Encoding encoding) {
-        final Encoding activeEncoding = this.encoding != null ? this.encoding : encoding;
-        final Optional<ParseState> result = parseImpl(makeScope(checkNotNull(scope, "scope")), checkNotNull(parseState, "parseState"), callbacks, activeEncoding);
-        callbacks.handle(this, result
-            .map(after -> success(this, parseState, after))
-            .orElseGet(() -> failure(this, parseState)));
+    public Optional<ParseState> parse(final Environment environment) {
+        final Environment activeEnvironment = this.encoding != null ? environment.withEncoding(this.encoding) : environment;
+        final Optional<ParseState> result = parseImpl(activeEnvironment.withScope(makeScope(environment.scope)));
```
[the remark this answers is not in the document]
jvdb: I kept the `SEPARATOR` on `Token` for now. Seems like a good place to specify what the separator should be, regardless of where it is used.

PR #228, opened 2017-11-13
Implement lazy versions of Cat and Bytes

### LINKED ISSUE #225: Add checkNotNegative() checks to isAvailable() in Source implementations

### LINKED ISSUE #226: Implement a lazy version of the Cat ValueExpression
Instead of concatenating bytes, it should concatenate the sources.

### LINKED ISSUE #227: Implement a lazy version of the Bytes ValueExpression
Instead of splitting the underlying byte array, it should produce smaller slices of the source Value.

jvdb:
Resolves #225.
Resolves #226.
Resolves #227.

`Cat` and `Bytes` are now lazy. Additionally, checks are added to the `Source` implementations to assure that provided `offset` and `length` arguments are not null and not negative.

### DIFF
--- a/core/src/main/java/io/parsingdata/metal/data/ConcatenatedValueSource.java
+++ b/core/src/main/java/io/parsingdata/metal/data/ConcatenatedValueSource.java
@@ -27,8 +27,8 @@ import io.parsingdata.metal.expression.value.Value;
 
 public class ConcatenatedValueSource extends Source {
 
-    private final Value left;
-    private final Value right;
+    public final Value left;
+    public final Value right;
 
     public ConcatenatedValueSource(final Value left, final Value right) {
         this.left = checkNotNull(left, "left");
@@ -60,7 +60,7 @@ public class ConcatenatedValueSource extends Source {
 
     @Override
     public String toString() {
-        return getClass().getSimpleName() + "(" + left.toString() + "," + right.toString() + "))";
+        return getClass().getSimpleName() + "(" + left + "," + right + "))";
     }
 
     @Override
--- a/core/src/main/java/io/parsingdata/metal/data/ConcatenatedValueSource.java
+++ b/core/src/main/java/io/parsingdata/metal/data/ConcatenatedValueSource.java
@@ -1,0 +1,78 @@
+/*
+ * Copyright 2013-2016 Netherlands Forensic Institute
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.parsingdata.metal.data;
+
+import static io.parsingdata.metal.Util.checkNotNegative;
+import static io.parsingdata.metal.Util.checkNotNull;
+
+import java.math.BigInteger;
+import java.util.Objects;
+
+import io.parsingdata.metal.Util;
+import io.parsingdata.metal.expression.value.Value;
+
+public class ConcatenatedValueSource extends Source {
+
+    private final Value left;
+    private final Value right;
+
+    public ConcatenatedValueSource(final Value left, final Value right) {
+        this.left = checkNotNull(left, "left");
+        this.right = checkNotNull(right, "right");
+    }
+
+    @Override
+    protected byte[] getData(final BigInteger offset, final BigInteger length) {
+        if (!isAvailable(offset, length)) {
+            throw new IllegalStateException("Data to read is not available ([offset=" + offset + ";length=" + length + ";source=" + this + ").");
+        }
+        final byte[] outputData = new byte[length.intValueExact()];
+        if (offset.add(length).compareTo(left.getLength()) <= 0) {
+            System.arraycopy(left.getValue(), offset.intValueExact(), outputData, 0, length.intValueExact());
+        } else if (offset.compareTo(left.getLength()) >= 0) {
+            System.arraycopy(right.getValue(), offset.subtract(left.getLength()).intValueExact(), outputData, 0, length.intValueExact());
+        } else {
+            final BigInteger leftPartLength = left.getLength().subtract(offset);
+            System.arraycopy(left.getValue(), offset.intValueExact(), outputData, 0, leftPartLength.intValueExact());
+            System.arraycopy(right.getValue(), 0, outputData, leftPartLength.intValueExact(), length.subtract(leftPartLength).intValueExact());
+        }
+        return outputData;
+    }
+
+    @Override
+    protected boolean isAvailable(final BigInteger offset, final BigInteger length) {
+        return checkNotNegative(length, "length").add(checkNotNegative(offset, "offset")).compareTo(left.getLength().add(right.getLength())) <= 0;
+    }
+
+    @Override
+    public String toString() {
+        return getClass().getSimpleName() + "(" + left.toString() + "," + right.toString() + "))";
+    }
+
+    @Override
+    public boolean equals(final Object obj) {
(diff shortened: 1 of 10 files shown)


core/src/main/java/io/parsingdata/metal/data/ConcatenatedValueSource.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    @Override
+    protected boolean isAvailable(final BigInteger offset, final BigInteger length) {
+        return checkNotNegative(length, "length").add(checkNotNegative(offset, "offset")).compareTo(left.getLength().add(right.getLength())) <= 0;
+    }
+
+    @Override
+    public String toString() {
+        return getClass().getSimpleName() + "(" + left.toString() + "," + right.toString() + "))";
```
rdvdijk: Since this PR looks good to me, and I still want to comment *something*, prepare for a nitpick:

The `.toString()` calls here are not needed, so would be equivalent:

```java
return getClass().getSimpleName() + "(" + left + "," + right + "))";
```

core/src/main/java/io/parsingdata/metal/data/ConcatenatedValueSource.java
```diff
@@ -0,0 +1,78 @@
...
+import java.math.BigInteger;
+import java.util.Objects;
+
+import io.parsingdata.metal.Util;
+import io.parsingdata.metal.expression.value.Value;
+
+public class ConcatenatedValueSource extends Source {
+
+    private final Value left;
+    private final Value right;
```
rdvdijk: These fields can probably be `public`. (I noticed that the other sources have `public` fields.)

core/src/main/java/io/parsingdata/metal/data/Slice.java
```diff
@@ -39,7 +40,7 @@ private Slice(final Source source, final BigInteger offset, final BigInteger len
     }
 
     public static Optional<Slice> createFromSource(final Source source, final BigInteger offset, final BigInteger length) {
-        if (offset.compareTo(ZERO) < 0 ||
+        if (checkNotNull(offset, "offset").compareTo(ZERO) < 0 ||
```
rdvdijk: Nice. I wonder why we didn't check this before?

core/src/main/java/io/parsingdata/metal/expression/value/Cat.java
```diff
@@ -34,12 +40,8 @@ public Cat(final ValueExpression left, final ValueExpression right) {
...
     @Override
     public Optional<Value> eval(final Value left, final Value right, final ParseState parseState, final Encoding encoding) {
-        final byte[] leftBytes = left.getValue();
-        final byte[] rightBytes = right.getValue();
-        final byte[] concatenatedBytes = new byte[leftBytes.length + rightBytes.length];
-        System.arraycopy(leftBytes, 0, concatenatedBytes, 0, leftBytes.length);
-        System.arraycopy(rightBytes, 0, concatenatedBytes, leftBytes.length, rightBytes.length);
-        return Optional.of(new Value(createFromBytes(concatenatedBytes), encoding));
+        return createFromSource(new ConcatenatedValueSource(left, right), ZERO, left.getLength().add(right.getLength()))
+            .flatMap(source -> Optional.of(new Value(source, encoding)));
```
rdvdijk: :tada:

core/src/test/java/io/parsingdata/metal/data/SourceAndSliceTest.java
```diff
@@ -55,7 +57,8 @@
         return Arrays.asList(new Object[][] {
             { new ConstantSource(DATA) },
             { new DataExpressionSource(con(DATA), 0, EMPTY_PARSE_STATE, enc()) },
-            { new ByteStreamSource(new InMemoryByteStream(DATA) ) }
+            { new ByteStreamSource(new InMemoryByteStream(DATA)) },
+            { new ConcatenatedValueSource(new Value(createFromSource(new ConstantSource(DATA), ZERO, BigInteger.valueOf(2)).get(), enc()), new Value(createFromSource(new ConstantSource(DATA), BigInteger.valueOf(2), BigInteger.valueOf(2)).get(), enc())) }
```
rdvdijk: Holy long line of code, Batman! :bat:

core/src/main/java/io/parsingdata/metal/expression/value/Cat.java
ccreeten: Flatmap is not necessary I think?

ccreeten: Also, I do not see an Optional empty, perhaps you forgot to clean up imports? :D

core/src/main/java/io/parsingdata/metal/expression/value/Bytes.java
```diff
@@ -63,16 +66,16 @@ public Bytes(final ValueExpression operand) {
...
         final ImmutableList<Optional<Value>> result = output.add(
-            head.map(value -> bytesToValues(new ImmutableList<>(), value.getValue(), 0, encoding).computeResult())
+            head.map(value -> bytesToValues(new ImmutableList<>(), value, 0, encoding).computeResult())
                 .orElseGet(ImmutableList::new));
         return tail.isEmpty() ? complete(() -> result) : intermediate(() -> toBytes(result, tail.head, tail.tail, encoding));
     }
 
-    private Trampoline<ImmutableList<Optional<Value>>> bytesToValues(final ImmutableList<Optional<Value>> output, final byte[] value, final int i, final Encoding encoding) {
-        if (i >= value.length) {
+    private Trampoline<ImmutableList<Optional<Value>>> bytesToValues(final ImmutableList<Optional<Value>> output, final Value value, final int i, final Encoding encoding) {
```
ccreeten: Could you give i a more descriptive name?

core/src/main/java/io/parsingdata/metal/data/ConcatenatedValueSource.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    @Override
+    protected byte[] getData(final BigInteger offset, final BigInteger length) {
+        if (!isAvailable(offset, length)) {
+            throw new IllegalStateException("Data to read is not available ([offset=" + offset + ";length=" + length + ";source=" + this + ").");
+        }
+        final byte[] outputData = new byte[length.intValueExact()];
+        if (offset.add(length).compareTo(left.getLength()) <= 0) {
+            System.arraycopy(left.getValue(), offset.intValueExact(), outputData, 0, length.intValueExact());
```
ccreeten: Are you using statements :open_mouth:? Arrays.copyOf(Range) are expressions you could use :smile:. Sadly it is not so nice when you need to copy from multiple arrays, as in the last else block :cry: .

core/src/main/java/io/parsingdata/metal/data/ConcatenatedValueSource.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    @Override
+    public String toString() {
+        return getClass().getSimpleName() + "(" + left.toString() + "," + right.toString() + "))";
+    }
+
+    @Override
+    public boolean equals(final Object obj) {
+        return Util.notNullAndSameClass(this, obj)
```
ccreeten: With all this casting, a fun thing to do would be creating a miniature DSL for this :yum: Rather simple and could make this equality look something like:
```java
return notNullAndSameClass(this, obj)
		.equal(source -> source.left)
		.equal(source -> source.right)
		.eval();
```

core/src/main/java/io/parsingdata/metal/data/Slice.java
jvdb: I think perhaps because these were protected or package-private constructors before so we more or less know whether it's necessary. Still, I prefer to be safe with such cheap checks.

core/src/main/java/io/parsingdata/metal/data/ConcatenatedValueSource.java
jvdb: Great idea! But a little too big a change for this PR I think...

core/src/main/java/io/parsingdata/metal/data/ConcatenatedValueSource.java
jvdb: Of course I prefer expressions, but when there is no escape the code may as well look uniform...

PR #233, opened 2017-11-30
List-based ConcatenatedValueSource and FoldCat

### LINKED ISSUE #229: Make ConcatenatedValueSource list-based to improve scalability
In the 7.0.0 release a new type of `Source` was introduced in order for the `Cat` `ValueExpression` to be lazy with regards to IO: `ConcatenatedValueSource`. This implementation was a typical binary expression that concatenates the result of two `ValueExpression`s. However, with very long chains of concatenated data this created a scalability problem (mostly with regard to required stack size).

Since long chains of concatenated values are quite common in the domain of binary data file formats (e.g., file systems) we need a better solution for this. The simplest approach is to let `ConcatenatedValueSource` itself have a list of concatenated values.

### LINKED ISSUE #230: Add a folding version of the Cat ValueExpression
As discussed in #229, long chains of concatenated data are a challenge for the 7.0.0 version of `ConcatenatedValueSource`. In order for a list-based implementation to work without too much `instanceof` magic, it makes sense to have a version of `Cat` that simply takes a single `ValueExpression` operand and then concatenate all the results it evaluates to. This is essentially the same as passing that operand to the `Fold` `ValueExpression` and providing the `Cat` `ValueExpression` as reducer.

This could be generalized to all `BinaryValueExpression`s, see #231 for a discussion of that.

### LINKED ISSUE #232: Add caching to `DataExpressionSource`
Generally we try very hard to avoid mutable state in Metal, but in the case of `DataExpressionSource` this causes serious performance problems. IO laziness when parsing data with a `DataExpressionSource` means that for evaluating every `Def` or `Until`, the entire expression is evaluated. This has led to a slowdown of a factor of more than 100x in a benchmark.

So in this specific case, it makes sense for `DataExpressionSource` to cache its result.

jvdb:
Resolves #230: Added `FoldCat`, a folding version of the `Cat` `ValueExpression` that essentially implements an optimized version of calling `Fold` with `Cat` as reducer.
Resolves #229: `ConcatenatedValueSource` is now list-based, in order to improve the scalability of long chains of concatenated values.
Resolves #232: `DataExpressionSource` now caches its data, because it needs to resolve it in its entirety anyway to perform any read- or size-related operation on it. The method that fills the cache is marked `synchronized`, which may be inefficient in some cases but for all current Metal use cases this is not expected to be a problem.

### DIFF
--- a/core/src/main/java/io/parsingdata/metal/data/ConcatenatedValueSource.java
+++ b/core/src/main/java/io/parsingdata/metal/data/ConcatenatedValueSource.java
@@ -16,23 +16,28 @@
 
 package io.parsingdata.metal.data;
 
+import static java.math.BigInteger.ZERO;
+
+import static io.parsingdata.metal.Trampoline.complete;
+import static io.parsingdata.metal.Trampoline.intermediate;
 import static io.parsingdata.metal.Util.checkNotNegative;
 import static io.parsingdata.metal.Util.checkNotNull;
 
 import java.math.BigInteger;
 import java.util.Objects;
 
+import io.parsingdata.metal.Trampoline;
 import io.parsingdata.metal.Util;
 import io.parsingdata.metal.expression.value.Value;
 
 public class ConcatenatedValueSource extends Source {
 
-    public final Value left;
-    public final Value right;
+    public final ImmutableList<Value> values;
+    public final BigInteger length;
 
-    public ConcatenatedValueSource(final Value left, final Value right) {
-        this.left = checkNotNull(left, "left");
-        this.right = checkNotNull(right, "right");
+    public ConcatenatedValueSource(final ImmutableList<Value> values, final BigInteger length) {
+        this.values = checkNotNull(values, "values");
+        this.length = checkNotNegative(length, "length");
     }
 
     @Override
@@ -40,39 +45,45 @@ public class ConcatenatedValueSource extends Source {
         if (!isAvailable(offset, length)) {
             throw new IllegalStateException("Data to read is not available ([offset=" + offset + ";length=" + length + ";source=" + this + ").");
         }
-        final byte[] outputData = new byte[length.intValueExact()];
-        if (offset.add(length).compareTo(left.getLength()) <= 0) {
-            System.arraycopy(left.getValue(), offset.intValueExact(), outputData, 0, length.intValueExact());
-        } else if (offset.compareTo(left.getLength()) >= 0) {
-            System.arraycopy(right.getValue(), offset.subtract(left.getLength()).intValueExact(), outputData, 0, length.intValueExact());
-        } else {
-            final BigInteger leftPartLength = left.getLength().subtract(offset);
-            System.arraycopy(left.getValue(), offset.intValueExact(), outputData, 0, leftPartLength.intValueExact());
-            System.arraycopy(right.getValue(), 0, outputData, leftPartLength.intValueExact(), length.subtract(leftPartLength).intValueExact());
+        return getData(values, ZERO, ZERO, offset, length, new byte[length.intValueExact()]).computeResult();
+    }
+
+    private Trampoline<byte[]> getData(ImmutableList<Value> values, BigInteger currentOffset, BigInteger currentDest, BigInteger offset, BigInteger length, byte[] output) {
+        if (length.compareTo(ZERO) <= 0) {
+            return complete(() -> output);
         }
-        return outputData;
+        if (values.isEmpty()) {
+            throw new IllegalStateException("Data to read is not available ([offset =" + offset + "; length=" + length + ";source=" + this + ").");
+        }
+        if (currentOffset.add(values.head.slice.length).compareTo(offset) <= 0) {
+            return intermediate(() -> getData(values.tail, currentOffset.add(values.head.slice.length), currentDest, offset, length, output));
+        }
+        final BigInteger localOffset = offset.subtract(currentOffset).compareTo(ZERO) < 0 ? ZERO : offset.subtract(currentOffset);
+        final BigInteger toCopy = length.compareTo(values.head.slice.length.subtract(localOffset)) > 0 ? values.head.slice.length.subtract(localOffset) : length;
+        System.arraycopy(values.head.slice.getData(), localOffset.intValueExact(), output, currentDest.intValueExact(), toCopy.intValueExact());
+        return intermediate(() -> getData(values.tail, currentOffset.add(values.head.slice.length), currentDest.add(toCopy), offset, length.subtract(toCopy), output));
     }
 
     @Override
     protected boolean isAvailable(final BigInteger offset, final BigInteger length) {
-        return checkNotNegative(length, "length").add(checkNotNegative(offset, "offset")).compareTo(left.getLength().add(right.getLength())) <= 0;
+        return checkNotNegative(length, "length").add(checkNotNegative(offset, "offset")).compareTo(this.length) <= 0;
     }
 
     @Override
     public String toString() {
-        return getClass().getSimpleName() + "(" + left + "," + right + "))";
+        return getClass().getSimpleName() + "(" + values + "(" + length + ")))";
     }
 
     @Override
     public boolean equals(final Object obj) {
         return Util.notNullAndSameClass(this, obj)
-            && Objects.equals(left, ((ConcatenatedValueSource)obj).left)
-            && Objects.equals(right, ((ConcatenatedValueSource)obj).right);
+            && Objects.equals(values, ((ConcatenatedValueSource)obj).values)
+            && Objects.equals(length, ((ConcatenatedValueSource)obj).length);
     }
 
     @Override
     public int hashCode() {
-        return Objects.hash(getClass(), left, right);
+        return Objects.hash(getClass(), values, length);
     }
 
 }
--- a/core/src/main/java/io/parsingdata/metal/data/ConcatenatedValueSource.java
+++ b/core/src/main/java/io/parsingdata/metal/data/ConcatenatedValueSource.java
@@ -43,7 +43,7 @@ public class ConcatenatedValueSource extends Source {
     @Override
     protected byte[] getData(final BigInteger offset, final BigInteger length) {
         if (!isAvailable(offset, length)) {
-            throw new IllegalStateException("Data to read is not available ([offset=" + offset + ";length=" + length + ";source=" + this + ").");
+            throw new IllegalStateException("Data to read is not available (offset=" + offset + ";length=" + length + ";source=" + this + ").");
         }
         return getData(values, ZERO, ZERO, offset, length, new byte[length.intValueExact()]).computeResult();
     }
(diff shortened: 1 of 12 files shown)


core/src/test/java/io/parsingdata/metal/expression/value/FoldCatTest.java
```diff
@@ -0,0 +1,72 @@
...
+import org.junit.Test;
+
+import io.parsingdata.metal.data.Environment;
+import io.parsingdata.metal.data.ImmutableList;
+import io.parsingdata.metal.data.ParseState;
+
+public class FoldCatTest {
+
+    public static final Optional<ParseState> RESULT =
+        seq(any("a"),
```
TheUnknownCylon: At first I was a bit confused about "abc" and 3 any's with an "a".
If you know that the "a" in any is the name, that it becomes more clear what happens here.

core/src/main/java/io/parsingdata/metal/data/DataExpressionSource.java
```diff
@@ -48,28 +51,32 @@ public DataExpressionSource(final ValueExpression dataExpression, final int inde
...
+            }
+            cache = getValueAtIndex(results, index, 0)
+                .computeResult()
+                .map(Value::getValue)
+                .orElseThrow(() -> new IllegalStateException("ValueExpression dataExpression yields empty Value at index " + index + "."));
         }
-        return getValueAtIndex(results, index, 0)
-            .computeResult()
-            .orElseThrow(() -> new IllegalStateException("ValueExpression dataExpression yields empty Value at index " + index + "."));
+        return Arrays.copyOf(cache, cache.length);
```
TheUnknownCylon: 'nitpick' cache.clone()? But I think that's just a personal preference...

TheUnknownCylon:
Overall these changes looks fine to me (but keep in mind that I am not very familiar with all the implementation details on Metal, so that makes my opinion perhaps a bit less useful)

core/src/main/java/io/parsingdata/metal/data/ConcatenatedValueSource.java
```diff
@@ -16,63 +16,74 @@
...
     @Override
     protected byte[] getData(final BigInteger offset, final BigInteger length) {
         if (!isAvailable(offset, length)) {
-            throw new IllegalStateException("Data to read is not available ([offset=" + offset + ";length=" + length + ";source=" + this + ").");
+            throw new IllegalStateException("Data to read is not available (offset=" + offset + ";length=" + length + ";source=" + this + ").");
+        }
+        return getData(values, ZERO, ZERO, offset, length, new byte[length.intValueExact()]).computeResult();
+    }
+
+    private Trampoline<byte[]> getData(ImmutableList<Value> values, BigInteger currentOffset, BigInteger currentDest, BigInteger offset, BigInteger length, byte[] output) {
```
ccreeten: Missing finals?

core/src/main/java/io/parsingdata/metal/data/ConcatenatedValueSource.java
```diff
@@ -16,63 +16,74 @@
...
 
-    public final Value left;
-    public final Value right;
+    public final ImmutableList<Value> values;
+    public final BigInteger length;
 
-    public ConcatenatedValueSource(final Value left, final Value right) {
-        this.left = checkNotNull(left, "left");
-        this.right = checkNotNull(right, "right");
+    public ConcatenatedValueSource(final ImmutableList<Value> values, final BigInteger length) {
```
ccreeten: Would you rather not let the length be calculated here? Does it ever make sense for length not being equal to the slice lengths of all values?

core/src/test/java/io/parsingdata/metal/expression/value/FoldCatTest.java
```diff
@@ -0,0 +1,72 @@
...
+
+import org.junit.Test;
+
+import io.parsingdata.metal.data.Environment;
+import io.parsingdata.metal.data.ImmutableList;
+import io.parsingdata.metal.data.ParseState;
+
+public class FoldCatTest {
+
+    public static final Optional<ParseState> RESULT =
```
ccreeten: A bit odd to cache this here I think? Should the parse blow up for some reason, you'll get some class loading error instead of a test failure?

core/src/main/java/io/parsingdata/metal/data/DataExpressionSource.java
```diff
@@ -48,28 +51,32 @@ public DataExpressionSource(final ValueExpression dataExpression, final int inde
...
     protected boolean isAvailable(final BigInteger offset, final BigInteger length) {
-        return checkNotNegative(offset, "offset").add(checkNotNegative(length, "length")).compareTo(getValue().slice.length) <= 0;
+        return checkNotNegative(offset, "offset").add(checkNotNegative(length, "length")).compareTo(BigInteger.valueOf(getValue().length)) <= 0;
     }
 
-    private Value getValue() {
-        final ImmutableList<Optional<Value>> results = dataExpression.eval(parseState, encoding);
-        if (results.size <= index) {
-            throw new IllegalStateException("ValueExpression dataExpression yields " + results.size + " result(s) (expected at least " + (index + 1) + ").");
+    private synchronized byte[] getValue() {
```
ccreeten: No double checked locking to avoid useless synchronization after it has been cached? :smile:

ccreeten: I just read the pr comment, so okay :smile:

core/src/main/java/io/parsingdata/metal/data/ConcatenatedValueSource.java
```diff
@@ -16,63 +16,74 @@
...
-        return getClass().getSimpleName() + "(" + left + "," + right + "))";
+        return getClass().getSimpleName() + "(" + values + "(" + length + "))";
     }
 
     @Override
     public boolean equals(final Object obj) {
         return Util.notNullAndSameClass(this, obj)
-            && Objects.equals(left, ((ConcatenatedValueSource)obj).left)
-            && Objects.equals(right, ((ConcatenatedValueSource)obj).right);
+            && Objects.equals(values, ((ConcatenatedValueSource)obj).values)
```
ccreeten: Can I perhaps make a setup for a nice comparison utility :joy:

core/src/main/java/io/parsingdata/metal/expression/value/FoldCat.java
```diff
@@ -0,0 +1,99 @@
...
+        final BigInteger length = calculateTotalSize(values);
+        if (length.compareTo(ZERO) == 0) {
+            return ImmutableList.create(Optional.empty());
+        }
+        return createFromSource(new ConcatenatedValueSource(values, length), ZERO, length)
+            .map(source -> new ImmutableList<Optional<Value>>().add(Optional.of(new Value(source, encoding))))
+            .orElseGet(ImmutableList::new);
+    }
+
+    private static <T, U extends T> Trampoline<ImmutableList<T>> unwrap(final ImmutableList<Optional<U>> input, final ImmutableList<T> output) {
```
ccreeten: Perhaps move this to a more general class?

core/src/main/java/io/parsingdata/metal/expression/value/FoldCat.java
```diff
@@ -0,0 +1,99 @@
...
+        return input.head
+            .map(value -> intermediate(() -> unwrap(input.tail, output.add(value))))
+            .orElseGet(() -> intermediate(() -> unwrap(input.tail, output)));
+    }
+
+    private BigInteger calculateTotalSize(final ImmutableList<Value> values) {
+        return calculateTotalSize(values, ZERO).computeResult();
+    }
+
+    private Trampoline<BigInteger> calculateTotalSize(final ImmutableList<Value> values, final BigInteger size) {
```
ccreeten: Same with this? Or move to CVS class? makes this ValueExpression implementation cleaner as well :smile:

core/src/main/java/io/parsingdata/metal/expression/value/FoldCat.java
ccreeten: Also, it looks like you never use a subtype of T so you don't need an extends T for now. Also, I tend to not use type parameters which are only used once, as they can be replaces by wildcards. That would make this signature ```private static <T> Trampoline<ImmutableList<T>> unwrap(final ImmutableList<? extends Optional<? extends T>> input, final ImmutableList<T> output)```

core/src/test/java/io/parsingdata/metal/expression/value/FoldCatTest.java
jvdb: Agreed, it was an incomplete refactoring. Thanks for pointing it out!

core/src/test/java/io/parsingdata/metal/expression/value/FoldCatTest.java
jvdb: Never thought about it this way because that string being used as name is so natural to me. But I totally agree it's confusing. I've changed the name to "any" to better signal it has nothing to do with the value.

core/src/main/java/io/parsingdata/metal/data/DataExpressionSource.java
jvdb: Good point. I thought I had some reason for doing it like this with regard to shallow copies on object or multidimensional arrays, but that's all true for `Arrays.copyOf()` as well :)

core/src/main/java/io/parsingdata/metal/data/ConcatenatedValueSource.java
jvdb: Sounds like a great idea!

core/src/main/java/io/parsingdata/metal/expression/value/FoldCat.java
jvdb: As you proposed, I've removed `U` but also removed both `extends` clauses. We only use this method to remove the `Optional`s, so might as well constrain it to that. We can always generalize it if we need to.

core/src/main/java/io/parsingdata/metal/data/ConcatenatedValueSource.java
jvdb: I really want to restrict constructors to assigning fields and doing some basic checking (like `checkNotNull()` or `checkNotNegative()`, etc. So that's why the length is calculated outside the constructor. At the same time, maybe it's better suited to a factory method, that way the constructor is guarded against weird inputs.

jvdb: I ended up refactoring this: the `ConcatenatedValueSource` constructor is now private but called by a factory `create()` method. This does mean that the length is calculated inside, which means it's no longer possible to create a `ConcatenatedValueSource` with a length larger than is possible given the underlying data. It also very much cleaned up the implementations of both `FoldCat` and `Cat`.

PR #234, opened 2017-12-07
Lazy calculate utility

DavyLandman:
After a discussion with @jvdb about the recent pull-request #233 that solves issue #232, I proposed a different way to implement the cache of the `DataExpressionSource`.

In this PR I've created two implementations of this caching behaviour, one that is lock free, and one that locks only for the initial case. This is better than the current behaviour of `getValue` that locks on every invocation. Looking at the code and after a discussion with @jvdb it seems you might not even need a lock, as the edge case of sometimes calculating the array twice is not that expensive.

### DIFF
--- a/core/src/main/java/io/parsingdata/metal/Lazily.java
+++ b/core/src/main/java/io/parsingdata/metal/Lazily.java
@@ -17,7 +17,9 @@ package io.parsingdata.metal;
 
 import java.util.function.Supplier;
 
-public class Lazily {
+public final class Lazily {
+
+    private Lazily() {}
 
     /**
      * Constructs a lazily calculated value (of type T) the first time it is retrieved.  <br/><br/>
--- a/core/src/main/java/io/parsingdata/metal/Lazily.java
+++ b/core/src/main/java/io/parsingdata/metal/Lazily.java
@@ -1,0 +1,73 @@
+/*
+ * Copyright 2017 SWAT.engineering BV
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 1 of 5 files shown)


ccreeten:
I am a bit confused, because both methods of caching have been considered, but it was decided not to implement them. Has Jeroen changed his mind? :smile: 

I do like the way it is set up, having a generic reusable function.

jvdb:
@ccreeten: This is the result of a discussion I had with @DavyLandman. He decided to implement it in an elegant abstraction. I am however still conflicted on the trade-off between simplicity and performance.

I also agree that it's not entirely clear whether we need locking at all. Are there hard guarantees from the JRE that two overlapping writes of the same data do not cause a problem when reading? If that's the case, then we can just forget about synchronizing completely.

DavyLandman:
The JVM memory model is explicit around reads. When two threads are reading a field from the same object instance, there is a chance of having an outdated view on the field (for example the previous value of the field). Volatile fixes this, at the cost of more expensive reads and writes.

Depending on the use case, you want the guarantee that the calculation is only performed once, in that case you need volatile, if there is no real problem when sometimes in a heavily multi-threaded case, the calculators is run twice, and it takes a few instructions before all threads see the same value of the field, then you can use the normal fields without locking.

The current solution might be overdesigned, I was just thinking about readability, and thought, if you abstract it like this, it might be nicer to read it. And if in the future there is more laziness to certain metal primitives, the class to take care of it is already ready.

jvdb:
Like I wrote before, this is a great and elegant solution! We will introduce it once we add more caching to Metal.

DavyLandman:
Good, we'll see when we get there.

I would advice to drop the synchronized keyword and the clone call, that's over-engineering for this case.

PR #235, opened 2017-12-13
Small gardening changes for the 7.1 release.

### LINKED ISSUE #211: Add shorthand for legacy Pre behaviour
And call it `iff` or `when`. Or perhaps `imp`, because it's closer to _implication_.

jvdb:
Resolves #211: The original `Pre` behaviour is back as a shorthand called `when`.

The other changes are the removal of an unneeded `.clone()` call and some unused imports.

### DIFF
--- a/core/src/main/java/io/parsingdata/metal/Shorthand.java
+++ b/core/src/main/java/io/parsingdata/metal/Shorthand.java
@@ -152,6 +152,10 @@ public final class Shorthand {
     public static Token until(final String name, final ValueExpression initialSize, final Token terminator) { return until(name, initialSize, null, terminator, null); }
     public static Token until(final String name, final Token terminator, final Encoding encoding) { return until(name, null, terminator, encoding); }
     public static Token until(final String name, final Token terminator) { return until(name, terminator, null); }
+    public static Token when(final String name, final Token token, final Expression predicate, final Encoding encoding) { return cho(name, encoding, pre(def(EMPTY_NAME, 0), not(predicate)), token); }
+    public static Token when(final String name, final Token token, final Expression predicate) { return when(name, token, predicate, null); }
+    public static Token when(final Token token, final Expression predicate, final Encoding encoding) { return when(EMPTY_NAME, token, predicate, encoding); }
+    public static Token when(final Token token, final Expression predicate) { return when(token, predicate, null); }
 
     public static BinaryValueExpression add(final ValueExpression left, final ValueExpression right) { return new Add(left, right); }
     public static BinaryValueExpression div(final ValueExpression left, final ValueExpression right) { return new Div(left, right); }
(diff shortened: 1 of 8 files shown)


core/src/main/java/io/parsingdata/metal/data/DataExpressionSource.java
```diff
@@ -76,7 +75,7 @@ protected boolean isAvailable(final BigInteger offset, final BigInteger length)
                 .map(Value::getValue)
                 .orElseThrow(() -> new IllegalStateException("ValueExpression dataExpression yields empty Value at index " + index + "."));
         }
-        return cache.clone();
+        return cache;
```
rdvdijk: This is not needed because of the `arraycopy` above?

core/src/main/java/io/parsingdata/metal/Shorthand.java
```diff
@@ -152,6 +152,10 @@ private Shorthand() {}
     public static Token until(final String name, final ValueExpression initialSize, final Token terminator) { return until(name, initialSize, null, terminator, null); }
     public static Token until(final String name, final Token terminator, final Encoding encoding) { return until(name, null, terminator, encoding); }
     public static Token until(final String name, final Token terminator) { return until(name, terminator, null); }
+    public static Token when(final String name, final Token token, final Expression predicate, final Encoding encoding) { return cho(name, encoding, pre(def(EMPTY_NAME, 0), not(predicate)), token); }
```
rdvdijk: Just checking: this is exactly how we implemented `iff`?

core/src/test/java/io/parsingdata/metal/ShorthandsTest.java
```diff
@@ -225,4 +226,21 @@ public void mapRightWithSub() {
...
+    }
+
+    @Test
+    public void whenFalse() {
+        Optional<ParseState> result =
+            seq(
+                when(def("name1", con(1), eq(con(1))), not(TRUE)),
+                def("name2", con(1), eq(con(2)))).parse(env(stream(2)));
+        assertTrue(result.isPresent());
+        assertEquals(1, result.get().offset.intValueExact());
```
rdvdijk: Also test that the correct value was parsed? (not just the offset)

core/src/main/java/io/parsingdata/metal/data/DataExpressionSource.java
jvdb: Exactly. The only client of this method that passes data out does the copying itself.

core/src/main/java/io/parsingdata/metal/Shorthand.java
jvdb: Well, @ccreeten provided me with this idea and I think he implemented the original `iff`...

rdvdijk: I'm not familiar with the details of the `iff` implementation, I'm just checking if this exactly is the same :wink:

PR #239, opened 2018-05-22
Added Google error-prone to the build and resolved all warnings

### LINKED ISSUE #237: Use Google error-prone in the build and resolve its warnings
See https://github.com/google/error-prone and http://errorprone.info/.

jvdb:
Resolves #237.

For a description of the resolved warnings, see the following:

http://errorprone.info/bugpattern/MissingOverride
http://errorprone.info/bugpattern/ReferenceEquality
http://errorprone.info/bugpattern/OperatorPrecedence
http://errorprone.info/bugpattern/DoubleBraceInitialization
http://errorprone.info/bugpattern/ClassCanBeStatic
http://errorprone.info/bugpattern/StringSplitter

### DIFF
--- a/core/src/main/java/io/parsingdata/metal/token/Until.java
+++ b/core/src/main/java/io/parsingdata/metal/token/Until.java
@@ -101,8 +101,8 @@ public class Until extends Token {
 
     private Trampoline<Optional<ParseState>> iterate(final Environment environment, final BigInteger currentSize, final BigInteger stepSize, final BigInteger maxSize) {
         if (stepSize.compareTo(ZERO) == 0 ||
-            stepSize.compareTo(ZERO) > 0 && currentSize.compareTo(maxSize) > 0 ||
-            stepSize.compareTo(ZERO) < 0 && currentSize.compareTo(maxSize) < 0) {
+            (stepSize.compareTo(ZERO) > 0 && currentSize.compareTo(maxSize) > 0) ||
+            (stepSize.compareTo(ZERO) < 0 && currentSize.compareTo(maxSize) < 0)) {
             return complete(Util::failure);
         }
         return environment.parseState
--- a/pom.xml
+++ b/pom.xml
@@ -99,6 +99,34 @@
   <build>
     <plugins>
       <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <version>3.3</version>
+        <configuration>
+          <compilerId>javac-with-errorprone</compilerId>
+          <forceJavacCompilerUse>true</forceJavacCompilerUse>
+          <!-- maven-compiler-plugin defaults to targeting Java 5, but our
+               javac only supports >=6 -->
+          <source>8</source>
+          <target>8</target>
+          <showWarnings>true</showWarnings>
+        </configuration>
+        <dependencies>
+          <dependency>
+            <groupId>org.codehaus.plexus</groupId>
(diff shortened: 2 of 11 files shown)


core/src/main/java/io/parsingdata/metal/token/Until.java
```diff
@@ -101,8 +101,8 @@ public Until(final String name, final ValueExpression initialSize, final ValueEx
 
     private Trampoline<Optional<ParseState>> iterate(final Environment environment, final BigInteger currentSize, final BigInteger stepSize, final BigInteger maxSize) {
         if (stepSize.compareTo(ZERO) == 0 ||
-            stepSize.compareTo(ZERO) > 0 && currentSize.compareTo(maxSize) > 0 ||
-            stepSize.compareTo(ZERO) < 0 && currentSize.compareTo(maxSize) < 0) {
+            (stepSize.compareTo(ZERO) > 0 && currentSize.compareTo(maxSize) > 0) ||
+            (stepSize.compareTo(ZERO) < 0 && currentSize.compareTo(maxSize) < 0)) {
```
mvanaken: We have to say "we told you so", but... :stuck_out_tongue_winking_eye: :blush: :+1:

pom.xml
```diff
@@ -98,6 +98,34 @@
...
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <version>3.3</version>
+        <configuration>
+          <compilerId>javac-with-errorprone</compilerId>
+          <forceJavacCompilerUse>true</forceJavacCompilerUse>
+          <!-- maven-compiler-plugin defaults to targeting Java 5, but our
+               javac only supports >=6 -->
+          <source>8</source>
```
mvanaken: This version is linked to  the property 'maven.compiler.source', right? Maybe create a new property that is e.g. used by both?

pom.xml
```diff
@@ -98,6 +98,34 @@
 
   <build>
     <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <version>3.3</version>
```
mvanaken: Move version to properties?

pom.xml
```diff
@@ -98,6 +98,34 @@
...
+               javac only supports >=6 -->
+          <source>8</source>
+          <target>8</target>
+          <showWarnings>true</showWarnings>
+        </configuration>
+        <dependencies>
+          <dependency>
+            <groupId>org.codehaus.plexus</groupId>
+            <artifactId>plexus-compiler-javac-errorprone</artifactId>
+            <version>2.8.3</version>
```
mvanaken: Move version to properties?

pom.xml
```diff
@@ -98,6 +98,34 @@
...
+          </dependency>
+          <!-- override plexus-compiler-javac-errorprone's dependency on
+               Error Prone with the latest version -->
+          <dependency>
+            <groupId>com.google.errorprone</groupId>
+            <artifactId>error_prone_core</artifactId>
+            <version>2.3.1</version>
```
mvanaken: Move version to properties? (Sorry, wilde dit in 1comment doen, maar kan de vorige blijkbaar niet meer editen...)

pom.xml
jvdb: Good point! Those properties actually directly set the values for the plug-in (I think :sweat_smile: ), so I just removed both the `source` and `target` property setters there to see if they would use those: still works!

PR #241, opened 2018-07-31
Current iteration levels

mvanaken:
Implemented level option to CurrentIteration to allow referencing to the iteration loop from nested loops to its parent loops.

Since this is a continuation on the current-index branch, this PR is a diff to that branch for now. We should merge to the master when current-index branch is merged.

- [x] Adjust base to master when current-index is merged.
- [x] Add coverage for exception in getLevel method of CurrentIteration.


core/src/main/java/io/parsingdata/metal/expression/value/reference/CurrentIteration.java
```diff
@@ -73,12 +92,35 @@ public String toString() {
...
+
+    /**
+     * ParseGraphCandidates keeps track of the last n ParseGraphs.
+     */
+    private class ParseGraphCandidates {
+
+        private final Queue<ParseGraph> list;
+
+        ParseGraphCandidates(final int n) {
+            list = new ArrayDeque<>(nCopies(n, ParseGraph.EMPTY));
```
ccreeten: Hmm, this is mutable state. Could you change it? An immutable list?

ccreeten: Hmm, this is mutable state. Could you return a new instance of this class when adding an element? 

I do not know how others feel about using a stlib ArrayDeque in metal (@jvdb ?), an ImmutableList could also be used instead.

core/src/main/java/io/parsingdata/metal/expression/value/reference/CurrentIteration.java
```diff
@@ -36,12 +38,33 @@
...
+    public CurrentIteration(final ValueExpression level) {
+        this.level = level;
+    }
+
     @Override
     public ImmutableList<Optional<Value>> eval(final ParseState parseState, final Encoding encoding) {
-        if (parseState.iterations.head == null) {
+        int level = getLevel(parseState, encoding);
+        ImmutableList<BigInteger> iterations = parseState.iterations;
+        while (level > 0 && iterations.tail != null) {
```
rdvdijk: I suspect this will not pass @jvdb's style checker 😁 Reassigning two local variables... You can check up-front if the `level` value is not too large compared to the size of `iterations`, eliminating the need to decrease it inside a loop. Finding the correct iteration can be achieved using a recursive method (which I think would be safe enough, unless we parse *really* deep data structures).

core/src/main/java/io/parsingdata/metal/expression/value/reference/CurrentIteration.java
```diff
@@ -36,12 +38,33 @@
...
-        return ImmutableList.create(Optional.of(createFromNumeric(parseState.iterations.head, new Encoding())));
+        return ImmutableList.create(Optional.of(createFromNumeric(iterations.head, new Encoding())));
+    }
+
+    private int getLevel(final ParseState parseState, final Encoding encoding) {
+        ImmutableList<Optional<Value>> evaluatedLevel = level.eval(parseState, encoding);
+        if (evaluatedLevel.size != 1 || !evaluatedLevel.head.isPresent()) {
+            throw new IllegalArgumentException("Level must evaluate to a single non-empty value.");
+        }
+        return evaluatedLevel.head.get().asNumeric().intValueExact();
```
rdvdijk: Metal uses `BigInteger` wherever possible, I don't think you need to revert to using an `int` here.

core/src/test/java/io/parsingdata/metal/expression/value/reference/CurrentIterationEdgeCaseTest.java
```diff
@@ -0,0 +1,67 @@
+package io.parsingdata.metal.expression.value.reference;
```
rdvdijk: Add the license.

core/src/main/java/io/parsingdata/metal/expression/value/reference/CurrentIteration.java
```diff
@@ -36,12 +38,33 @@
  */
 public class CurrentIteration implements ValueExpression {
 
+    private final ValueExpression level;
+
+    public CurrentIteration(final ValueExpression level) {
+        this.level = level;
```
rdvdijk: Use `checkNotNull`.

core/src/test/java/io/parsingdata/metal/expression/value/reference/CurrentIterationTest.java
```diff
@@ -53,6 +56,7 @@
             { "[0, 1, 2, 3] repn=4(CURRENT_ITERATION)", repn(VALUE_EQ_ITERATION, con(4)), stream(0, 1, 2, 3), enc(), true },
             { "[255, 0, 1, 2, 3, 255] def(255), while<3(CURRENT_ITERATION), def (255)", seq(VALUE_EQ_255, whl(VALUE_EQ_ITERATION, not(eq(con(3)))), VALUE_EQ_255), stream(255, 0, 1, 2, 3, 255), enc(), true },
             { "[0, 0, 1, 2, 255, 1, 0, 1, 2, 255] repn=2(CURRENT_ITERATION, repn=3(CURRENT_ITERATION))", repn(seq(VALUE_EQ_ITERATION, repn(VALUE_EQ_ITERATION, con(3)), VALUE_EQ_255), con(2)), stream(0, 0, 1, 2, 255, 1, 0, 1, 2, 255), enc(), true },
+            { "[0, 0, 0, 0, 1, 1, 1, 1, 2, 2, 2, 2] repn=2(CURRENT_ITERATION, repn=2(PARENT_ITERATION, repn=2(GRANDPARENT_ITERATION)))", repn(seq(VALUE_EQ_ITERATION, repn(seq(VALUE_EQ_ITERATION_PARENT, repn(VALUE_EQ_ITERATION_GRANDPARENT, con(2))), con(1))), con(3)), stream(0, 0, 0, 0, 1, 1, 1, 1, 2, 2, 2, 2), enc(), true },
```
rdvdijk: The description of the test does not match the actual `Token`, right? Description shows `repn2(repn2(repn2 ))` but the actual `Token` shows `repn3(repn1(repn2))`

core/src/main/java/io/parsingdata/metal/expression/value/reference/CurrentIteration.java
mvanaken: Ok.

core/src/test/java/io/parsingdata/metal/expression/value/reference/CurrentIterationEdgeCaseTest.java
mvanaken: Ok.

core/src/test/java/io/parsingdata/metal/expression/value/reference/CurrentIterationTest.java
mvanaken: Correct. Adjusted it.

core/src/main/java/io/parsingdata/metal/expression/value/reference/CurrentIteration.java
mvanaken: Agreed. Using BigInteger instead.

core/src/main/java/io/parsingdata/metal/expression/value/reference/CurrentIteration.java
mvanaken: Ok. Using an recursive method instead.

core/src/main/java/io/parsingdata/metal/expression/value/reference/CurrentIteration.java
```diff
@@ -36,12 +41,38 @@
...
+
+    private BigInteger getLevel(final ParseState parseState, final Encoding encoding) {
+        final ImmutableList<Optional<Value>> evaluatedLevel = level.eval(parseState, encoding);
+        if (evaluatedLevel.size != 1 || !evaluatedLevel.head.isPresent()) {
+            throw new IllegalArgumentException("Level must evaluate to a single non-empty value.");
+        }
+        return evaluatedLevel.head.get().asNumeric();
+    }
+
+    private Optional<Value> getIterationRecursive(final ImmutableList<BigInteger> iterations, final BigInteger level) {
```
jvdb: This is a recursive method, so it should be implemented with a `Trampoline`. I agree it will probably never lead to scalability issues in practice, but let's avoid some weird corner case by always avoid hard recursion :)

core/src/main/java/io/parsingdata/metal/expression/value/reference/CurrentIteration.java
```diff
@@ -36,12 +41,38 @@
...
+        if (parseState.iterations.size <= level.longValue()) {
+            return Optional.empty();
+        }
+        return getIterationRecursive(parseState.iterations, level);
+    }
+
+    private BigInteger getLevel(final ParseState parseState, final Encoding encoding) {
+        final ImmutableList<Optional<Value>> evaluatedLevel = level.eval(parseState, encoding);
+        if (evaluatedLevel.size != 1 || !evaluatedLevel.head.isPresent()) {
+            throw new IllegalArgumentException("Level must evaluate to a single non-empty value.");
```
ccreeten: Perhaps separate these checks so you can return even more informative messages?

jvdb: We do this in many places, perhaps we need a utility method for this case (like `checkSingleValue` or something).

ccreeten: Yes :D Added more involved typing would also solve the size check.

PR #244, opened 2018-08-03
Parameterized parse verification

mvanaken:
During development of #241 and #242 this addition helped me a lot to check if the tests were in fact fully parsed. Decided to create a separate pull request for it.

Only had to improve the until tests, since there were 4 tests, that didn't parse until the end. For these tests I decreased the size of the stream.

Also adjusted some other descriptions within the until tests to correctly represent what the tests do.

### DIFF
--- a/core/src/test/java/io/parsingdata/metal/token/ParameterizedUntilTest.java
+++ b/core/src/test/java/io/parsingdata/metal/token/ParameterizedUntilTest.java
@@ -44,25 +44,31 @@ import io.parsingdata.metal.util.ParameterizedParse;
 
 public class ParameterizedUntilTest extends ParameterizedParse {
 
+    private static final Token ABC = def("abc", 3, eq(con("abc")));
+
     @Parameters(name = "{0} ({4})")
     public static Collection<Object[]> data() {
         return Arrays.asList(new Object[][] {
-            { "[a,b,c] i=0,s=1,m=3 ab",            untilToken(0, 1, 3, con('c'), con("ab")),     stream("abc", US_ASCII), enc(), true },
-            { "[a,b,c,a,b,c] i=3,s=1,m=6 abcab",   untilToken(3, 1, 6, con('c'), con("abcab")),  stream("abcabc", US_ASCII), enc(), true },
-            { "[a,b,c] i=0,s=2,m=3 ab",            untilToken(0, 2, 3, con('c'), con("ab")),     stream("abc", US_ASCII), enc(), true },
-            { "[a,b,c,a,b,c] i=1,s=2,m=6 abcab",   untilToken(1, 2, 6, con('c'), con("abcab")),  stream("abcabc", US_ASCII), enc(), true },
-            { "[a,b,c,a,b,c] i=0,s=3,m=6",         untilToken(0, 3, 6, con('c'), con("")),       stream("abcabc", US_ASCII), enc(), false },
-            { "[a,b,c,a,b,c] i=0,s=0,m=6",         untilToken(0, 0, 6, con(""), con("")),        stream("abcabc", US_ASCII), enc(), false },
-            { "[a,b,c,a,b,c] i=6,s=-1,m=0 abcab",  untilToken(6, -1, 0, con('c'), con("abcab")), stream("abcabc", US_ASCII), enc(), true },
-            { "[] i=0,s=-1,m=6",                   untilToken(0, -1, 6, con(""), con("")),       stream("", US_ASCII), enc(), false },
-            { "[] i=6,s=1,m=0",                    untilToken(6, 1, 0, con(""), con("")),        stream("", US_ASCII), enc(), false },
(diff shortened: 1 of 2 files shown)


core/src/test/java/io/parsingdata/metal/util/ParameterizedParse.java
```diff
@@ -44,7 +47,13 @@
 
     @Test
     public void test() throws IOException {
-        assertEquals(result, token.parse(env(parseState, encoding)).isPresent());
+        Optional<ParseState> endState = token.parse(env(parseState, encoding));
+        assertEquals(result, endState.isPresent());
+
+        endState.ifPresent(state -> {
```
jvdb: The `ifPresent()`-check is unnecessary I think, because the `assertEquals()` above it will terminate execution if it's not. So rewrite this to an `assertFalse()` over the `state.slice(ONE).isPresent()`-expression?

core/src/test/java/io/parsingdata/metal/token/ParameterizedUntilTest.java
```diff
@@ -46,21 +47,21 @@
     @Parameters(name = "{0} ({4})")
     public static Collection<Object[]> data() {
         return Arrays.asList(new Object[][] {
-            { "[a,b,c,a,b,c] i=0,s=1,m=6 ab",     untilToken(0, 1, 6, con('c'), con("ab")),     stream("abcabc", US_ASCII), enc(), true },
```
jvdb: I'm not entirely sure if it's a good idea to truncate these tests in this way... They may currently parse in the same way as with the longer stream, but the reason for example that "abcabc" was provided for a parse that runs until(c) is to make sure the implementation stops at the _first_ match. So if you remove a second instance of "c", then the test no longer validates this.

The same is probably true for some of the other tests below. I think we should consider to maybe separate these tests for the standard parameterized implementation, or maybe make that implementation configurable, or something else...

core/src/test/java/io/parsingdata/metal/util/ParameterizedParse.java
rdvdijk: I'm fine with style, else we'd need a `final ParseState state = endState.get()`.

jvdb: Oh, it's used to avoid a direct `get()`? I would just use `get()` if you do a hard check for `isPresent()` there anyway... But this is not a big thing to me.

core/src/test/java/io/parsingdata/metal/token/ParameterizedUntilTest.java
mvanaken: Yes, I agree about the missing validation that it stops at the first match. I don't think we need a to separate these tests though. We could include an additional def after the until, such that we parse the whole stream and still test if we stop at the first match. This was how I originally adjusted it, but then decided that this would make the tests too complex and I wanted to keep it simple.

Shall I change them back and then add a seq(until..., def('ABC')) around it?

core/src/test/java/io/parsingdata/metal/util/ParameterizedParse.java
mvanaken: I prefer this style. It feels more robust to me, although the difference is small. When you use `get()` then the second test _must_ precede this tests, so there is a dependency. In this style, the order doesn't matter, the first test can even be removed without changing the behaviour of this test.

jvdb: I agree there are advantages in _defensiveness_ :-)

core/src/test/java/io/parsingdata/metal/token/ParameterizedUntilTest.java
jvdb: Either that or make the added check optional in `ParameterizedParse`. It could be standard with the default constructor, but with an optional constructor that disables that check.

I don't feel strongly about either solution, so if you do, then do it :) Otherwise we can talk about it next week.

mvanaken: Ok, I have reated overload methods which introduce an argument for the expected tail of the stream, which wraps the until token and the expected tail in a seq.

jvdb: Looks good!

core/src/test/java/io/parsingdata/metal/util/ParameterizedParse.java
ccreeten: It is a test class, so I'm fine with either :) But neither of the tests can be removed right? Removing the isPresent check will make everything succeed if nothing is present?

core/src/test/java/io/parsingdata/metal/token/ParameterizedUntilTest.java
```diff
@@ -43,27 +44,33 @@
 
 public class ParameterizedUntilTest extends ParameterizedParse {
 
+    private static final Token ABC = def("abc", 3, eq(con("abc")));
```
ccreeten: Easy as 123...

Issue #246, opened 2018-08-29
Refactor 6 lines occurring 7 times in 7 files: Last.java, First.java, Count.java, UnaryValueExpression.java, FoldCat.java, UnaryLogicalExpression.java, Bytes.java


rdvdijk:
I've selected for refactoring 6 lines of code which are duplicated in 7 file(s) ([1](https://github.com/parsingdata/metal/blob/fe71ac29b9ed838fc25edb5d8a32f536647cb8d8/core/src/main/java/io/parsingdata/metal/expression/value/reference/Last.java#L49-L56), [2](https://github.com/parsingdata/metal/blob/fe71ac29b9ed838fc25edb5d8a32f536647cb8d8/core/src/main/java/io/parsingdata/metal/expression/value/reference/First.java#L56-L63), [3](https://github.com/parsingdata/metal/blob/fe71ac29b9ed838fc25edb5d8a32f536647cb8d8/core/src/main/java/io/parsingdata/metal/expression/value/reference/Count.java#L55-L62), [4](https://github.com/parsingdata/metal/blob/fe71ac29b9ed838fc25edb5d8a32f536647cb8d8/core/src/main/java/io/parsingdata/metal/expression/value/UnaryValueExpression.java#L70-L77), [5](https://github.com/parsingdata/metal/blob/fe71ac29b9ed838fc25edb5d8a32f536647cb8d8/core/src/main/java/io/parsingdata/metal/expression/value/FoldCat.java#L55-L62), [6](https://github.com/parsingdata/metal/blob/fe71ac29b9ed838fc25edb5d8a32f536647cb8d8/core/src/main/java/io/parsingdata/metal/expression/logical/UnaryLogicalExpression.java#L42-L49), [7](https://github.com/parsingdata/metal/blob/fe71ac29b9ed838fc25edb5d8a32f536647cb8d8/core/src/main/java/io/parsingdata/metal/expression/value/Bytes.java#L81-L88)).  Addressing this will make our codebase more maintainable and improve [Better Code Hub](https://bettercodehub.com)'s **Write Code Once** guideline rating! 👍 

Here's the gist of this guideline:
- **Definition** 📖 
Do not copy code.
- **Why**❓ 
When code is copied, bugs need to be fixed in multiple places. This is both inefficient and a source of regression bugs.
- **How** 🔧 
Avoid duplication by never copy/pasting blocks of code and reduce duplication by extracting shared code, either to a new unit or introduce a superclass if the language permits.

[...]

Issue #70, opened 2016-07-18
Add type mechanism for single value list expressions


rdvdijk:
Examples: In `Def` and `Nod` the `size` expression needs to evaluate to a list with a single value. Parsing will fail otherwise. `RepN` needs a list with a single value for its `n` argument.

Is there a way to enforce this using the type system?

jvdb:
In 94a8d2a12be621f9be0740d747a1e598bfae9963 I have added a `SingleValueExpression` interface that does not return a list but a single `Optional<Value>`. I have created it as a subclass of the `ValueExpression` interface, so that it can be used wherever a `ValueExpression` is required. A default method is present to convert to a list whenever needed.

Some things to consider:
1. It works quite well: almost all modifications to the test code are in order to be more precise when specifying a `ValueExpression` with a single value.
2. Currently only `Last` and `Const` have been converted as well as the `size` argument to `Def`. There are many more candidates, but in order to consider this change, a minimal set seems useful.
3. Other candidates for conversion if we want to make this permanent are, among the `ValueExpression`s, `Count`, `CurrentIteration`, `CurrentOffset`, `First`, `Self`, as well as all the `Fold` implementations. The `RepN` `Token`'s `n` argument as well as some of the `ValueExpression` arguments of the `While` `Token` can be considered for conversion as well.
4. Error-prone [complains](https://errorprone.info/bugpattern/FunctionalInterfaceMethodChanged) about the `@FunctionalInterface` annotation since it may give surprising behaviour when casting lambda expressions. I've suppressed it for now since I believe out default method's behaviour is in line with their best practice, but I'm not entirely sure.
5. One fundamental change is that when using a `SingleValueExpression` where a `ValueExpression` is required, there is no distinction between an empty list and a list containing a single empty value. In fact, using the default method to handle `ValueExpression` behaviour, an empty list is never returned. If this is a big problem the `SingleValueExpression` can override the `eval()` method, but that may be cumbersome. Right now I'm not even entirely sure if it's a problem if we lose this distinction.

rdvdijk:
Looks pretty good!

One thing, code like this is becoming a bit weird to look at (pasted from a test) :

```java
def("d", last(len(last(ref("c")))))
```

We might want to introduce `Shorthand`s for these `def` cases, where in the current code we would not need a `last` around a `len` (for example).

ccreeten:
[...]

PR #247, opened 2018-08-30
Introduce OneToManyValueExpression and OneToOneValueExpression

rdvdijk:
This removes duplicated code in these `ValueExpressions`: `Last`, `First`, `Count` and `FoldCat`. This is achieved by introducing a new abstract base class `OneToOneValueExpression`. The four classes look pretty clean now!

The old `UnaryValueExpression` has been renamed to `OneToManyValueExpression`.

The names of these classes are still open to debate, not too sure if these are the best names.
* one-to-one meaning: one operand leads to a one (optional) value
* one-to-many meaning: one operand leads to many (optional) values

I found that the `Bytes` class still contains this exact code, but is not a match for either base class, since we "expand" each result of the operand into multiple values.

### DIFF
--- a/core/src/main/java/io/parsingdata/metal/Shorthand.java
+++ b/core/src/main/java/io/parsingdata/metal/Shorthand.java
@@ -44,6 +44,7 @@ import io.parsingdata.metal.expression.value.Bytes;
 import io.parsingdata.metal.expression.value.Cat;
 import io.parsingdata.metal.expression.value.Const;
 import io.parsingdata.metal.expression.value.ConstantFactory;
+import io.parsingdata.metal.expression.value.OneToManyValueExpression;
 import io.parsingdata.metal.expression.value.reference.CurrentIteration;
 import io.parsingdata.metal.expression.value.Elvis;
 import io.parsingdata.metal.expression.value.Expand;
@@ -51,7 +52,6 @@ import io.parsingdata.metal.expression.value.FoldLeft;
 import io.parsingdata.metal.expression.value.FoldRight;
 import io.parsingdata.metal.expression.value.FoldCat;
 import io.parsingdata.metal.expression.value.Reverse;
-import io.parsingdata.metal.expression.value.UnaryValueExpression;
 import io.parsingdata.metal.expression.value.Value;
 import io.parsingdata.metal.expression.value.ValueExpression;
 import io.parsingdata.metal.expression.value.arithmetic.Add;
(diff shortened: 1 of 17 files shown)


core/src/main/java/io/parsingdata/metal/expression/value/OneToOneValueExpression.java
```diff
@@ -0,0 +1,77 @@
...
+    public OneToOneValueExpression(final ValueExpression operand) {
+        this.operand = checkNotNull(operand, "operand");
+    }
+
+    @Override
+    public ImmutableList<Optional<Value>> eval(final ParseState parseState, final Encoding encoding) {
+        return ImmutableList.create(eval(operand.eval(parseState, encoding), parseState, encoding));
+    }
+
+    public abstract Optional<Value> eval(final ImmutableList<Optional<Value>> list, final ParseState parseState, final Encoding encoding);
```
rdvdijk: Self-review: I think we do not need to pass `ParseState` into this method, we do not use it anywhere.

core/src/test/java/io/parsingdata/metal/ValueExpressionSemanticsTest.java
```diff
@@ -33,13 +33,13 @@
 import java.io.IOException;
 import java.util.Optional;
 
+import io.parsingdata.metal.expression.value.OneToManyValueExpression;
```
rdvdijk: Self-review: Fix import order.

core/src/main/java/io/parsingdata/metal/expression/value/FoldCat.java
```diff
@@ -16,14 +16,11 @@
 
 package io.parsingdata.metal.expression.value;
 
-import static java.math.BigInteger.ZERO;
-
 import static io.parsingdata.metal.data.Slice.createFromSource;
+import static java.math.BigInteger.ZERO;
```
rdvdijk: Self-review: fix import order?

core/src/main/java/io/parsingdata/metal/expression/value/OneToOneValueExpression.java
rdvdijk: @jvdb @mvanaken I think that in both this new class as in `OneToManyValueExpression`, there is no need to have the `ParseState` argument. We never access it anywhere. Even the `Encoding` is rarely accessed; I found only a single use of it in the code I refactored here.

core/src/main/java/io/parsingdata/metal/Shorthand.java
```diff
@@ -164,10 +164,10 @@ private Shorthand() {}
...
     public static BinaryValueExpression sub(final ValueExpression left, final ValueExpression right) { return new Sub(left, right); }
     public static BinaryValueExpression mod(final ValueExpression left, final ValueExpression right) { return new Mod(left, right); }
-    public static UnaryValueExpression neg(final ValueExpression operand) { return new Neg(operand); }
+    public static OneToManyValueExpression neg(final ValueExpression operand) { return new Neg(operand); }
     public static BinaryValueExpression and(final ValueExpression left, final ValueExpression right) { return new io.parsingdata.metal.expression.value.bitwise.And(left, right); }
     public static BinaryValueExpression or(final ValueExpression left, final ValueExpression right) { return new io.parsingdata.metal.expression.value.bitwise.Or(left, right); }
-    public static UnaryValueExpression not(final ValueExpression operand) { return new io.parsingdata.metal.expression.value.bitwise.Not(operand); }
+    public static OneToManyValueExpression not(final ValueExpression operand) { return new io.parsingdata.metal.expression.value.bitwise.Not(operand); }
```
ccreeten: Referring to your naming debate, reading this does make it seem a bit funny in that it now reads like `negate/not` have a different cardinality between input and output (consider an identity function, it would be even more weird imo), although both are simply (and mostly unbounded) ValueExpressions. Before, `Unary` meant a single operand, with implicitly a single output value (as always). Now I think it is a mix of two concepts. 

But thinking of a good name is hard. UnaryManyToMany to me feels like I would know better what it all is about. But maybe there may be better names. Or I am just thinking about it wrong.

Also fun times thinking about naming `exp`, two operands, with the first one able to be any size, and the second one exactly one element. :sos:

jvdb: I agree that because of our list semantics, the notion of what is a _single_ operand is a bit complex...

ccreeten: Sounds good :) It is also interesting for expressions with more than one operand, where e.g. `mul` can have two lists, but `exp` can have a list as the first operand, but requires a single element as the second.

It would probably be cool to encode that information in the operand types, i.e. `BinaryExpr&ToManyExpr exp(ToManyExpr value, ToOneExpr second)`. It can then allow us to statically check for correctness as well, as per #70 (although there are few cases where it is necessary?). Downside is that we can not pass a list expression where we know that it will only ever contain a single value, so in the `exp` case, it would need a wrap with a `last` for example.

ccreeten:
The cleanup bit is nice I think!

jvdb:
Closing this PR for now, we'll keep it as reference for issue #246 and others.

PR #252, opened 2018-10-08
Gardening/cleanup for upcoming 8.0.0 release

### LINKED ISSUE #245: Clean up instantiation of default Encoding objects.
There are quite some `new Encoding()` calls throughout the library. This can be cleaned up by using a fixed constant, also to make sure there are no incorrect situations.

### LINKED ISSUE #249: Use isEmpty() instead of checking for nulls when using an ImmutableList or ParseGraph
Improves the abstraction of using `ImmutableList`.

jvdb:
Resolves #249: Using only `isEmpty()` on immutable containers/collections instead of `null` checking.
Resolves #245: Remove default `Encoding` constructor and replace with constant.

Additionally, the other commits are mostly concerned with cleanup of test code: improve assert statements and generics use as well as remove unnecessary `import` statements and `throws` clauses.


ccreeten:
Ahh, too late. But I had one question, perhaps you could look at adding accessor methods for all variables? That way, it can play nice with method references for example, or creating overloads in subclasses.

jvdb:
Good point! I've made an issue for this: #253.

Issue #250, opened 2018-09-19
Add a "locally scoped" ref to avoid unpredictable references in recursive Token definitions


DavyLandman:
Given the following pseudo-java code which has a recursive Token `A`:

```java
final Token A = seq("A",
    def("b", con(1)), 
    token("AorB"), 
    def("c", con(1), ge(last(ref("A.b")))
);
final Token B = seq("B", def("a", con(1)));
final Token AorB = cho("AorB", A, B);
```

It seems impossible to calculate how far back in the `ref("A.b")` stream we have to look to get the `b` field that was parsed at the same level of recursion (or is nesting the better term)?

I would like to propose an alternative to `ref` which is a ref that only searches the local stack and higher, not in the children.

(problem discovered together with @pinoval )

DavyLandman:
After some more discussion and considering cases with @pinoval, it might be that we need a kind of stackfull reference. So that we can also pass along a value expression to a function, that keeps on referring to this stack, even though it's used at a deeper level

jvdb:
Sounds interesting!

I think my initial approach will be to solve the case as described at the top. I have some fairly simple ideas, but I'm not sure if they will work. It is quite possible that if those solutions solve this case that they are not adequate for what you are doing -- but that would be the time to come up with a new sample case :)

The most basic approach will be to implement a `ValueExpression` that refers to a value's current _scope_ with regard to referencing other values. So basically this would allow you to create scopes on the fly based on where you are. I will create a branch for the working test case and possible solutions.

DavyLandman:
I've tried to create a horrible example:

```java
    final Token A = seq("A", def("b", con(1)), ABC, d(last(ref("A.b"))));
    final Token B = seq("B", def("b", con(1)), d(last(ref("B.b"))));
    final Token C = seq("C", def("b", con(1)));
    final Token ABC = cho("ABC", A, B, C);

    final Token d(ValueExpression x) {
        return seq("D_dynamic", ABC, def("c", con(1), eq(x)));
    }
```

The `d` function is a way to reduce duplication, we've seen this pattern before in other metal examples, and its quite a useful one. But with the recursive/nested structures, it's impossible to pass it the right ValueExpression that points to the correct field at the correct nesting level.

> The most basic approach will be to implement a ValueExpression that refers to a value's current scope with regard to referencing other values. So basically this would allow you to create scopes on the fly based on where you are. I will create a branch for the working test case and possible solutions.

This sounds like it might be good enough, even for the above example.

[...]

PR #256, opened 2018-11-04
Small improvements and test case for #250.

jvdb:
Three commits and three small and unrelated changes:

1. Removed a bunch of obsolete JUnit-annotations.
2. Added a base test case related to issue #250.
3. Removed the `TokenCallback` class and replaced its use with `ImmutablePair<Token, Callback>`.


rdvdijk:
I think the third change deserved its own PR :-)

core/src/main/java/io/parsingdata/metal/data/ImmutablePair.java
```diff
@@ -34,7 +34,7 @@ public ImmutablePair(final L left, final R right) {
 
     @Override
     public String toString() {
-        return left + "=" + right;
+        return left + "->" + right;
```
rdvdijk: Another option:

```suggestion
        return "(" + left + "," + right + ")";
```

core/src/test/java/io/parsingdata/metal/ScopeTest.java
```diff
@@ -0,0 +1,77 @@
...
+import io.parsingdata.metal.data.ParseState;
+import io.parsingdata.metal.token.Token;
+
+public class ScopeTest {
+
+    private final Token nested =
+        seq("nested",
+            def("left", 1),
+            token("nestedOrTerminator"),
+            def("right", 1, eq(nth(ref("left"), sub(count(ref("left")), count(ref("right"))))))
```
rdvdijk: Grin, pretty cool that this works.

PR #268, opened 2018-12-02
Added and cleaned up tests for local reference.

jvdb:
Having these tests is useful independent of a local reference implementation since they both contain a calculated solution that does not depend on local reference. However, they are good cases because they would be much simpler with an `lref` available, as shown in comments at the bottom of both tests.

### DIFF
--- a/core/src/test/java/io/parsingdata/metal/expression/value/reference/LocalReferenceNestedScopeTest.java
+++ b/core/src/test/java/io/parsingdata/metal/expression/value/reference/LocalReferenceNestedScopeTest.java
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package io.parsingdata.metal;
+package io.parsingdata.metal.expression.value.reference;
 
 import static java.math.BigInteger.ONE;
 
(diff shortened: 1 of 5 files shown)


core/src/test/java/io/parsingdata/metal/expression/value/reference/LocalReferenceNestedScopeTest.java
```diff
@@ -40,38 +40,46 @@
...
         );
+    }
 
-    @Test
-    public void nestedScopes() {
-        Optional<ParseState> parseState = format.parse(env(stream(42, 1, 2, 3, 42, 3, 2, 1), enc()));
+    private void nestedScopes(final Expression rightExpression) {
+        Optional<ParseState> parseState = topLevelNestedScopes(rightExpression).parse(env(stream(42, 1, 2, 3, 42, 3, 2, 1), enc()));
         assertTrue(parseState.isPresent());
         assertFalse("The test has not parsed the whole stream. It ended at offset " + parseState.get().offset + ".", parseState.get().slice(ONE).isPresent());
```
rdvdijk: Do we test for this often? If so, maybe extract a helper method, something like `assertStreamFullyConsumed(ParseState)` or something like that.

jvdb: `ParameterizedParse` already does this, as @mvanaken added this a while back. At some point we'll need a standard method for this somewhere I agree.

PR #270, opened 2018-12-06
Gardening of tooling and configurations

jvdb:
1. Added support for newer versions of Maven (3.5+) and cleaned up pom.xml (the eclipse plug-in was causing warnings and was entirely unneeded).
2. Updated PMD configuration (it wasn't working properly even though it didn't complain) and handled the issues arising from that (one of which was also reported by error-prone).
3. Cleaned up Travis configuration (it seems we did a full build twice? The install phase script was unnecessary) and added Sonar analysis to it.


.travis.yml
```diff
@@ -1,16 +1,16 @@
 language: java
 sudo: false # faster builds
 
-install: mvn clean install -DskipTests
-
-script: mvn clean verify
+addons:
+  sonarcloud:
```
rdvdijk: I must be blind 🙈, but where can I see the results of this?

jvdb: https://sonarcloud.io/project/issues?id=parsingdata_metal&pullRequest=270&resolved=false

Issue #274, opened 2018-12-17
Unify naming of ValueExpression instances


jvdb:
The smells (from Sonar) about hiding field names with variables actually expose that our naming scheme is a little messy. We often have this problem where we have a `ValueExpression` operand called `something` and then do this:

`final ImmutableList<Optional<Value>> something = this.something.eval(parseState, encoding);`

Which then hides the field. I understand why because after the `eval()`-call it's more or less the _new_ version of the same value that we use from then on. But it may still be error-prone if we ever somehow want to use the field again (e.g., for another different evaluation).

Additionally, we use different naming schemes in some locations, like calling the variable `somethingValues` or `evaluatedSomething`.

In addition to this, there is a distinction between `ValueExpression`s where we are capable of processing a list of evaluated values and where we explicitly forbid returning a list and only accept a single, non-empty value (e.g., the `size` argument on `Def`). Still, we tend to name both types using singular nouns instead of plurals. With the singular values, we often also extract the final value, often as an `int` or `BigInteger`.

My proposal is to unify everything using the following style (example is from `Expand`, which has a _plural_ argument (`base`) and a _singular_ one (`count`)):

1. Name constructor arguments and fields according to whether they are plural or singular. So the arguments and fields would be called `bases` and `count`.
2.~~When evaluating the fields, prefix the names with `evaluated`, so in this case `evaluatedBases` and `evaluatedCount`.~~ It's probably better to name these things -`list` instead of `evaluated`-. So it would then be `baseList` and `countList`.
3. When passing the full list around to other methods, the argument names become suffixed with `Values` to in this case `baseValues`. There seems to be no reason to pass `evaluatedCount` around as a list (see next item). Not sure whether we should keep this or keep using the `list`-suffix.
4. When extracting one actual value from the list, use the _singular_ version of the noun (because it always represents a single value) and suffix the names with `Value`, so in this case `baseValue` (for example to bind a variable when iterating the list) and `countValue`.

The only question that remains is what we do with the many instances of the names `operand`, `left` and `right`? In many cases with basic operators there are no other names for the operands. But should we rename the cases where lists are acceptable then to use `operands`, `lefts` and `rights`? It somehow feels a bit weird, but it is the direct consequence of this naming scheme.

_Originally posted by @jvdb in https://github.com/parsingdata/metal/pull/273#issuecomment-447499291_

As discussed in #275, there are some problems with this proposal and we've decided:

1. These names resolve the shadowing issues but they are far from perfect or desirable.
[...]

PR #275, opened 2018-12-20
Unify ValueExpression instance naming and update JavaDoc

jvdb:
Implemented the naming scheme as proposed in #274.

~~Additionally, when updating the JavaDoc to reflect the new names, also updated it. Most important change there is that in places where constructor arguments/fields are referred to as _operand(s)_, this has been changed to _field()s_ since that word is more accurate in the context of the Java code it describes.~~ <- Undoing this because it just makes it more unreadable...

### DIFF
--- a/core/src/main/java/io/parsingdata/metal/expression/value/BinaryValueExpression.java
+++ b/core/src/main/java/io/parsingdata/metal/expression/value/BinaryValueExpression.java
@@ -33,16 +33,16 @@ import io.parsingdata.metal.encoding.Encoding;
 /**
  * Base class for {@link ValueExpression}s with two operands.
  * <p>
- * A BinaryValueExpression implements a ValueExpression that has two operands:
+ * A BinaryValueExpression implements a ValueExpression that has two fields:
  * <code>lefts</code> and <code>rights</code> (both {@link ValueExpression}s).
- * Both operands are themselves first evaluated. If at least one of the
- * operands evaluates to {@link Optional#empty()}, the result of the
- * ValueExpression itself will be empty as well.
+ * Both fields are first evaluated. If at least one of the operands evaluates to
+ * {@link Optional#empty()}, the result of the ValueExpression itself will be
+ * empty as well.
  * <p>
- * For lists, values with the same index are evaluated in this manner. If
- * lists are of unequal length, the result is a list with evaluated values the
- * same size as the shortest list, appended with empty values to match the
- * size of the longest list.
+ * For lists, values with the same index are evaluated in this manner. If lists
+ * are of unequal length, the result is a list with evaluated values the same
+ * size as the shortest list, appended with empty values to match the size of the
+ * longest list.
  * <p>
  * To implement a BinaryValueExpression, only the
  * {@link #eval(Value, Value, ParseState, Encoding)} method must be implemented,
--- a/core/src/main/java/io/parsingdata/metal/expression/value/BinaryValueExpression.java
+++ b/core/src/main/java/io/parsingdata/metal/expression/value/BinaryValueExpression.java
@@ -34,7 +34,7 @@ import io.parsingdata.metal.encoding.Encoding;
  * Base class for {@link ValueExpression}s with two operands.
  * <p>
  * A BinaryValueExpression implements a ValueExpression that has two operands:
- * <code>left</code> and <code>right</code> (both {@link ValueExpression}s).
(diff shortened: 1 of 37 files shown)


rdvdijk:
To be honest, this does not look like an improvement to me.

Let's take a look at the first class in the diff:

```java
public abstract class BinaryValueExpression implements ValueExpression {

    public final ValueExpression left;
    public final ValueExpression right;

    public BinaryValueExpression(final ValueExpression left, final ValueExpression right) {
        this.left = checkNotNull(left, "left");
        this.right = checkNotNull(right, "right");
    }
```

Now turns into:

```java
public abstract class BinaryValueExpression implements ValueExpression {
    public final ValueExpression lefts;
    public final ValueExpression rights;

    public BinaryValueExpression(final ValueExpression lefts, final ValueExpression rights) {
        this.lefts = checkNotNull(lefts, "lefts");
        this.rights = checkNotNull(rights, "rights");
    }
```

The fact that a single `ValueExpression` results in a collection of values, multiple "`lefts`", does not warrant the field name change. We have a single `ValueExpression`, so the singular `left` make total sense to me.

If we continue with this, we will end up with Gollum-ish variable names, like `leftses` for `ValueExpression[]` or `List<ValueExpression>`.

> The Precious will be ours once the hobbitses are dead.

Isn't the problem mostly with the variable names we use for the result of these expressions, instead of the names of the fields?

rdvdijk:
Maybe it would make more sense to give the resulting variables more meaningful names.

From `Bytes`:

```java
final ImmutableList<Optional<Value>> evaluatedOperands = operands.eval(parseState, encoding);
```

`evaluatedOperands` says nothing about what we are actually dealing with here.

In this case, using a more meaningful name could be:

```java
final ImmutableList<Optional<Value>> bytes = operand.eval(parseState, encoding);
```

(Note the singular `operand` here.)

jvdb:
> In this case, using a more meaningful name could be:
> 
> ```java
> final ImmutableList<Optional<Value>> bytes = operand.eval(parseState, encoding);
> ```
Unfortunately this is a good example of why `evaluatedOperands` is a better name: this line you are quoting in no way represents anything that could accurately be described by the name `bytes`. It's just a completely unknown result of evaluating a `ValueExpression` that we will transform into a representation of byte values. But at this point in the processing, `bytes` completely misrepresents what it is, because it's a list of `Optional<Value>` instances without any additional meaning attached to it.

jvdb:
To respond to your first comment, I totally agree that `lefts`, `rights` and even `operands` looks horrible. I was trying to apply the scheme without thinking in order to find some edge situations that would be unclear or strange. The `lefts`-case is the most glaring one, but there are some others.

For example, when extracting a value from a list of `evaluatedBases` we would refer to it as a `baseValue`, but once we convert that from a `Value` to a `BigInteger` (or something else) we still use `baseValue`? It's a little difficult to determine a good scheme there.

rdvdijk:
> Unfortunately this is a good example of why `evaluatedOperand` (with or without the s) is a better name: this line you are quoting in no way represents anything that could accurately be described by the name `bytes`.

I agree that the name `bytes` in this case is also not ideal. But `evaluatedOperand` is also rather confusing to me. In many (if not all) cases we should be able to come up with a more meaningful name. Concatenating the variable name (`operand`) with the method we invoked on it (`eval`) does not improve our code much, IMHO.

As we know, [naming things](https://martinfowler.com/bliki/TwoHardThings.html) is hard :grin:

rdvdijk:
> This makes sense on some level: in the case of `left` or `operand`, the name describes what it itself _is_, a left operand or the operand. In the case of `bases` or `offsets`, the name describes what it _represents_, which can be singular or plurar. It just bugs me that in many cases we don't have a real description (Unless we want to give the operands specific names like _divisor_ for the right operand of `Div`?)

I feel your pain. But I guess that's the problem with all the very generic code inside a parser combinator implementation.

Using the name `divisor` in `Div` makes a lot of sense to me, and I think we should do things like that wherever we can (`left` would become `dividend`). But within classes like `BinaryValueExpression`, we have no choice but to use those very generic names.

jvdb:
Ok this is it. I totally agree that stuff like `methodedVariable` are not really nice names. But I do think that this PR moves things a little bit in the right direction, so I hope you'll agree @rdvdijk @mvanaken @ccreeten :)

jvdb:
I'm going to incorporate the best stuff from this branch/PR into a single commit on the other PR (#273). Thanks for the feedback!

Issue #280, opened 2019-02-05
Create a clean implementation for NOT_A_VALUE


jvdb:
The work on #279 will lead to a rather simple approach to `NOT_A_VALUE` that is a little error-prone (basically, because it is possible for a user of the library to create a valid `Value` object that is the same as the `NOT_A_VALUE`-constant).

A clean approach will be to split `Value` into a `Value` interface and a `ValueImpl` (or `BaseValue`, `CoreValue`, ...) class. A `NotAValue` class can then be created with only a `private` constructor that is called when instantiating the constant. The `ValueImpl`'s constructor(s) can then forbid constructing a zero-sized `Value`.

jvdb:
#282 is merged, so issue closed (no autoclose it seems because it was merged through another branch/PR first...)

PR #281, opened 2019-02-06
Change return value of ValueExpression.eval(...) to ImmutableList<Value>

### LINKED ISSUE #279: Change ImmutableList<Optional<Value>> to ImmutableList<Value>
This would entail adding a `NOT_A_VALUE` constant for non-values (e.g., NaN) to take the place of empty `Optional`s in the list. It's probably best if we mimic typical NaN behaviour, so that when some value of `NOT_A_VALUE` is encountered, the entire computation results in `NOT_A_VALUE`. Additionally comparing `NOT_A_VALUE` to `NOT_A_VALUE` results in `false`.

Two issues that will follow this refactoring are: #280 (`NOT_A_VALUE` cleanup) and #70 (`SingleValueExpression`).

jvdb:
This refactoring changes the return value of `ValueExpression.eval(...)` from `ImmutableList<Optional<Value>>` to `ImmutableList<Value>`. Putting optionals in a list doesn't really make sense, instead, there is now a `NOT_A_VALUE` constant to express results that have no value, such as the result of a division-by-zero.

In the earliest commits (not visible in the full diff now) the return value was changed to `Optional<ImmutableList<Value>>` however since the original interface already required a non-null return value, wrapping the new return value into an optional appeared unnecessary. Furthermore, it becomes quite tricky to distinguish between return values such as `Optional.empty()`, an empty `ImmutableList`, a list containing a single `NOT_A_VALUE`, etc.

The current semantics should be that:
1. Null is never returned. This was already the case previously.
2. A valid request with no results (such as a `ref("a")` when no values with name "a" exist) returns an empty list.
3. Any impossible computation will return a `NOT_A_VALUE`. In the context of processing lists, only the specific computation that results in `NOT_A_VALUE` will do so, not the surrounding ones. For example, the result of div([1, 1, 1], [1, 0, 1]) will evaluate to [1, NOT_A_VALUE, 1].
4. A comparison involving `NOT_A_VALUE` results in false, so `eq(con(NOT_A_VALUE), con(NOT_A_VALUE))` evaluates to `false`. This is in line with behaviour defined by [IEEE-754](https://en.wikipedia.org/wiki/IEEE_754).

Some other points to consider:

1. In situations where a `SingleValueExpression` is required, sometimes an exception is thrown upon encountering a `NOT_A_VALUE`, but not always (e.g., `CurrentIteration` and `Expand` do it, but `Ref` doesn't). Let's unify this when we implement #70.
2. There are quite a lot of error-prone warnings about using reference equality when checking for `NOT_A_VALUE`. This is currently by design to simplify the size of the diff this issue generates and will be fixed in #280.

Finally, I've come across some things to consider and perhaps change:
1. If `Sub` (the `Token`) or `Tie` is invoked with an empty list of `ValueExpressions` (for the `offset`s and `dataExpression`s respectively) that doesn't fail the parse, which makes sense, but it does insert an empty `ParseGraph`. I'm not sure that's a good idea.
2. `CurrentIteration` returns an empty list when a nonsensical `level` is provided such as a negative number. It also does this when the requested level does not exist (e.g., 5 when there are only 3 levels). Perhaps it makes more sense to return `NOT_A_VALUE` for a negative number (i.e., impossible) and 0 for a too high number (i.e., we are technically in iteration 0 for everything above the current high level)?

If we decide to change behaviour, let's make new issues so they don't make this diff larger. Additionally, I will pick up any issues resulting from the analysis tools on this PR.

This PR resolves #279.


jvdb:
Issues identified by tools are resolved. Only remaining one that we can ignore for now is the issues that BetterCodeHub raises, since those issues mostly relate to the introduction of `NOT_A_VALUE`. The current implementation is just a placeholder for when we continue with #280.

core/src/main/java/io/parsingdata/metal/data/DataExpressionSource.java
```diff
@@ -67,19 +66,20 @@ protected boolean isAvailable(final BigInteger offset, final BigInteger length)
...
             if (results.size <= index) {
-                throw new IllegalStateException(format("ValueExpression dataExpression yields %d result(s) (expected at least %d).", results.size, index+1));
+                throw new IllegalStateException(format("ValueExpression dataExpression must yield at least %d results.", index+1));
             }
-            cache = getValueAtIndex(results, index, 0)
-                .computeResult()
-                .map(Value::getValue)
-                .orElseThrow(() -> new IllegalStateException(format("ValueExpression dataExpression yields empty Value at index %d.", index)));
+            final Value cacheValue = getValueAtIndex(results, index, 0).computeResult();
+            if (cacheValue == Value.NOT_A_VALUE) {
```
rdvdijk: You can `static import` `NOT_A_VALUE`.

core/src/main/java/io/parsingdata/metal/expression/value/Value.java
```diff
@@ -31,6 +32,7 @@
 
 public class Value {
 
+    public static final Value NOT_A_VALUE = new Value(Slice.createFromBytes(new byte[]{}), DEFAULT_ENCODING);
```
rdvdijk: Use `new byte[0]` instead.

core/src/main/java/io/parsingdata/metal/expression/value/Expand.java
```diff
@@ -51,19 +51,19 @@ public Expand(final ValueExpression bases, final ValueExpression count) {
...
             return baseList;
         }
-        final ImmutableList<Optional<Value>> countList = count.eval(parseState, encoding);
-        if (countList.size != 1 || !countList.head.isPresent()) {
+        final ImmutableList<Value> countList = count.eval(parseState, encoding);
+        if (countList.size != 1 || countList.head == NOT_A_VALUE) {
             throw new IllegalArgumentException("Count must evaluate to a single non-empty value.");
         }
-        return expand(baseList, countList.head.get().asNumeric().intValueExact(), new ImmutableList<>()).computeResult();
+        return expand(baseList, countList.head.asNumeric().intValueExact(), new ImmutableList<>()).computeResult();
```
rdvdijk: I really like that this `.get()` has now disappeared.

core/src/main/java/io/parsingdata/metal/expression/value/Fold.java
```diff
@@ -57,42 +57,47 @@ public Fold(final ValueExpression values, final BinaryOperator<ValueExpression>
...
 
     @Override
-    public ImmutableList<Optional<Value>> eval(final ParseState parseState, final Encoding encoding) {
-        final ImmutableList<Optional<Value>> initialList = initial != null ? initial.eval(parseState, encoding) : new ImmutableList<>();
-        if (initialList.size > 1) {
-            return new ImmutableList<>();
+    public ImmutableList<Value> eval(final ParseState parseState, final Encoding encoding) {
+        final ImmutableList<Value> initialList = initial != null ? initial.eval(parseState, encoding) : new ImmutableList<>();
+        if (initialList.size > 1 || initialList.head == NOT_A_VALUE) {
+            return ImmutableList.create(NOT_A_VALUE);
```
rdvdijk: There are other cases where we throw an `Exception` in this case. Shouldn't we do so here? (Probably a separate issue.)

core/src/main/java/io/parsingdata/metal/expression/value/UnaryValueExpression.java
```diff
@@ -54,19 +55,26 @@ public UnaryValueExpression(final ValueExpression operand) {
...
         if (values.isEmpty()) {
             return complete(() -> result);
         }
-        return intermediate(() -> eval(values.tail, parseState, encoding, result.add(values.head.flatMap(value -> eval(value, parseState, encoding)))));
+        return intermediate(() -> eval(values.tail, parseState, encoding, result.add(checkEval(values.head, parseState, encoding))));
     }
 
     public abstract Optional<Value> eval(final Value value, final ParseState parseState, final Encoding encoding);
 
+    private Value checkEval(final Value value, final ParseState parseState, final Encoding encoding) {
```
rdvdijk: `checkEval` might not be the best name for this method... maybe `checkedEval`? or `safeEval`?

core/src/main/java/io/parsingdata/metal/expression/value/Value.java
rdvdijk: And maybe extract an `EMPTY_SLICE`.

core/src/main/java/io/parsingdata/metal/expression/value/reference/CurrentIteration.java
```diff
@@ -56,8 +57,10 @@ public CurrentIteration(final ValueExpression level) {
     }
 
     @Override
-    public ImmutableList<Optional<Value>> eval(final ParseState parseState, final Encoding encoding) {
-        return ImmutableList.create(getIteration(parseState, encoding));
+    public ImmutableList<Value> eval(final ParseState parseState, final Encoding encoding) {
+        return getIteration(parseState, encoding)
+            .map(ImmutableList::create)
+            .orElseGet(ImmutableList::new);
```
rdvdijk: How common is this pattern of `.map(ImmutableList::create).orElseGet(ImmutableList::new);`? If it happens very often we could consider extracting a helper method... Although that would mean passing an `Optional`. Nevermind. :grinning:

core/src/main/java/io/parsingdata/metal/expression/value/reference/Ref.java
```diff
@@ -68,26 +68,29 @@ private Ref(final T reference, final Predicate<ParseValue> predicate, final Valu
...
 
     @Override
-    public ImmutableList<Optional<Value>> eval(final ParseState parseState, final Encoding encoding) {
+    public ImmutableList<Value> eval(final ParseState parseState, final Encoding encoding) {
         if (limit == null) {
             return evalImpl(parseState, NO_LIMIT);
         }
-        ImmutableList<Optional<Value>> evaluatedLimit = limit.eval(parseState, encoding);
-        if (evaluatedLimit.size != 1 || !evaluatedLimit.head.isPresent()) {
+        final ImmutableList<Value> evaluatedLimit = limit.eval(parseState, encoding);
```
rdvdijk: (Missing `final`, nice find.)

core/src/main/java/io/parsingdata/metal/data/ConcatenatedValueSource.java
```diff
@@ -42,24 +44,15 @@ private ConcatenatedValueSource(final ImmutableList<Value> values, final BigInte
         this.length = checkNotNegative(length, "length");
     }
 
-    public static Optional<ConcatenatedValueSource> create(final ImmutableList<Optional<Value>> optionalValues) {
-        final ImmutableList<Value> values = unwrap(optionalValues, new ImmutableList<>()).computeResult();
+    public static Optional<ConcatenatedValueSource> create(final ImmutableList<Value> optionalValues) {
+        final ImmutableList<Value> values = reverse(optionalValues);
```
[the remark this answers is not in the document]
ccreeten: Yea, it added the head to the accumulator each recursive call, so the first element here would be the last in the returned list, etc.

core/src/main/java/io/parsingdata/metal/data/DataExpressionSource.java
```diff
@@ -67,19 +66,20 @@ protected boolean isAvailable(final BigInteger offset, final BigInteger length)
...
+                throw new IllegalStateException(format("ValueExpression dataExpression yields empty Value at index %d.", index));
```
ccreeten: Perhaps make it more explicitly known it is `NOT_A_VALUE`?

core/src/main/java/io/parsingdata/metal/expression/comparison/ComparisonExpression.java
```diff
@@ -58,22 +56,22 @@ public ComparisonExpression(final ValueExpression value, final ValueExpression p
 
     @Override
     public boolean eval(final ParseState parseState, final Encoding encoding) {
-        final ImmutableList<Optional<Value>> values = value == null ? ImmutableList.create(parseState.order.current().map(identity())) : value.eval(parseState, encoding);
+        final ImmutableList<Value> values = value == null ? parseState.order.current().map(ImmutableList::<Value>create).orElseGet(ImmutableList::new) : value.eval(parseState, encoding);
```
ccreeten: Could you not instead put in the constructor `this.value = value == null ? SELF : value`. Or perhaps even pass SELF through Shorthand (preferably?).

core/src/main/java/io/parsingdata/metal/expression/value/BinaryValueExpression.java
```diff
@@ -64,26 +65,33 @@ public BinaryValueExpression(final ValueExpression left, final ValueExpression r
...
-    private Trampoline<ImmutableList<Optional<Value>>> padList(final ImmutableList<Optional<Value>> list, final long size) {
+    private Trampoline<ImmutableList<Value>> padList(final ImmutableList<Value> list, final long size) {
         if (size <= 0) {
             return complete(() -> list);
         }
-        return intermediate(() -> padList(list.add(Optional.empty()), size - 1));
+        return intermediate(() -> padList(list.add(NOT_A_VALUE), size - 1));
+    }
+
+    private Value checkEval(final Value leftValue, final Value rightValue, final ParseState parseState, final Encoding encoding) {
```
ccreeten: `check` seems a bit weird for the method name, though I can't think of a better one.

core/src/main/java/io/parsingdata/metal/expression/value/Fold.java
ccreeten: Or should we not throw exceptions elsewhere? Returning `NOT_A_VALUE` here seems fine to me, at first though anyway.

core/src/main/java/io/parsingdata/metal/expression/value/Fold.java
```diff
@@ -57,42 +57,47 @@ public Fold(final ValueExpression values, final BinaryOperator<ValueExpression>
...
 
-    private Trampoline<Boolean> containsEmpty(final ImmutableList<Optional<Value>> list) {
+    private Trampoline<Boolean> containsNotAValue(final ImmutableList<Value> list) {
         if (list.isEmpty()) {
             return complete(() -> false);
         }
-        return list.head
-            .map(t -> intermediate(() -> containsEmpty(list.tail)))
-            .orElseGet(() -> complete(() -> true));
+        if (list.head != NOT_A_VALUE) {
```
ccreeten: Stupidly minor, but switch around the statements? More positivity!

core/src/main/java/io/parsingdata/metal/expression/value/UnaryValueExpression.java
ccreeten: Oh, I commented the same above. I was thinking of `safeEval` as well. `doEval`? :stuck_out_tongue_closed_eyes:

core/src/main/java/io/parsingdata/metal/expression/value/ValueExpression.java
```diff
@@ -27,15 +25,14 @@
  * <p>
  * A ValueExpression is an expression that is evaluated by executing its
  * {@link #eval(ParseState, Encoding)} method. It yields a list of
- * {@link Value} objects encapsulated in {@link Optional} objects (to guard
- * against <code>null</code>s).
+ * {@link Value} objects.
  * <p>
- * As context, it receives the current <code>parseState</code> object that as
- * well as the current <code>encoding</code> object.
+ * As context, it receives the current <code>ParseState</code> object as
```
ccreeten: Use `@link`s?

core/src/main/java/io/parsingdata/metal/expression/value/arithmetic/Div.java
```diff
@@ -42,7 +44,7 @@ public Div(final ValueExpression left, final ValueExpression right) {
     @Override
     public Optional<Value> eval(final Value leftValue, final Value rightValue, final ParseState parseState, final Encoding encoding) {
         if (rightValue.asNumeric().equals(ZERO)) {
-            return Optional.empty();
+            return Optional.of(NOT_A_VALUE);
```
ccreeten: Prepare to add `Infinity` some time... :smile:

core/src/main/java/io/parsingdata/metal/expression/value/FoldCat.java
```diff
@@ -45,11 +44,11 @@ public FoldCat(final ValueExpression operand) {
...
 
     @Override
-    public ImmutableList<Optional<Value>> eval(final ParseState parseState, final Encoding encoding) {
+    public ImmutableList<Value> eval(final ParseState parseState, final Encoding encoding) {
         return ConcatenatedValueSource.create(operand.eval(parseState, encoding))
             .flatMap(source -> createFromSource(source, ZERO, source.length))
-            .map(slice -> new ImmutableList<Optional<Value>>().add(Optional.of(new Value(slice, encoding))))
-            .orElseGet(() -> ImmutableList.create(Optional.empty()));
+            .map(slice -> ImmutableList.create(new Value(slice, encoding)))
+            .orElseGet(ImmutableList::new);
```
mvanaken: Wow, this looks so much cleaner! :) I like it!

core/src/main/java/io/parsingdata/metal/expression/value/reference/Self.java
```diff
@@ -35,8 +31,8 @@
 public class Self implements ValueExpression {
 
     @Override
-    public ImmutableList<Optional<Value>> eval(final ParseState parseState, final Encoding encoding) {
-        return ImmutableList.create(parseState.order.current().map(identity()));
+    public ImmutableList<Value> eval(final ParseState parseState, final Encoding encoding) {
+        return parseState.order.current().map(ImmutableList::<Value>create).orElseGet(ImmutableList::new);
```
mvanaken: I like the way you moved it to three lines in CurrentIteration. I would prefer to keep that consistent and also move it to three lines here.

core/src/main/java/io/parsingdata/metal/data/DataExpressionSource.java
jvdb: (Was already resolved when rewriting `==` checks to `equals(...)`.

core/src/main/java/io/parsingdata/metal/expression/value/Fold.java
jvdb: We need to evaluate where we throw an exception and where we return `NOT_A_VALUE`. @ccreeten: I tend to agree that the latter will usually be entirely acceptable.

core/src/main/java/io/parsingdata/metal/expression/value/reference/Self.java
jvdb: Both `Self` and `CurrentIteration` have been changed in the other two PRs quite a bit, but thanks for pointing me towards `CurrentIteration`! It turns out I forgot to convert its `level` operand to a `SingleValueExpression`!

core/src/main/java/io/parsingdata/metal/data/DataExpressionSource.java
```diff
@@ -67,19 +66,20 @@ protected boolean isAvailable(final BigInteger offset, final BigInteger length)
 
     private synchronized byte[] getValue() {
         if (cache == null) {
-            final ImmutableList<Optional<Value>> results = dataExpression.eval(parseState, encoding);
+            final ImmutableList<Value> results = dataExpression.eval(parseState, encoding);
             if (results.size <= index) {
-                throw new IllegalStateException(format("ValueExpression dataExpression yields %d result(s) (expected at least %d).", results.size, index+1));
+                throw new IllegalStateException(format("ValueExpression dataExpression must yield at least %d results.", index+1));
```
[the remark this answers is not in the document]
jvdb: I have no idea... I will revert it. I do remember changing it around, probably because I was trying to make the `format(...)`-method in `Util` simple by only accepting a single substitution argument or something like that.

core/src/main/java/io/parsingdata/metal/data/ConcatenatedValueSource.java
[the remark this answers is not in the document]
jvdb: As @ccreeten states, unwrapping also involved reversing :sweat_smile:

core/src/main/java/io/parsingdata/metal/data/ConcatenatedValueSource.java
```diff
@@ -42,24 +44,15 @@ private ConcatenatedValueSource(final ImmutableList<Value> values, final BigInte
         this.length = checkNotNegative(length, "length");
     }
 
-    public static Optional<ConcatenatedValueSource> create(final ImmutableList<Optional<Value>> optionalValues) {
-        final ImmutableList<Value> values = unwrap(optionalValues, new ImmutableList<>()).computeResult();
+    public static Optional<ConcatenatedValueSource> create(final ImmutableList<Value> optionalValues) {
```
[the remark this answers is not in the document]
jvdb: Renamed to `inputValues` to prevent name collision.

core/src/main/java/io/parsingdata/metal/expression/value/Value.java
jvdb: As you know, I ended up extracting an entire class :sweat_smile:

core/src/main/java/io/parsingdata/metal/expression/value/UnaryValueExpression.java
jvdb: Well if you _both_ thought of `safeEval(...)` I'm going to assume that's a better choice and go with that! Here and also at `BinaryValueExpression`.

core/src/main/java/io/parsingdata/metal/expression/value/BinaryValueExpression.java
jvdb: `safeEval(...)` it is (see `UnaryValueExpression`-discussion).

core/src/main/java/io/parsingdata/metal/expression/comparison/ComparisonExpression.java
jvdb: Agreed, but I want to keep this PR focussed, but we need to pick this up as well as part of #187.

jvdb:
I have checked the BetterCodeHub-results. They seem to indicate some worsening of their metrics but everything still seems quite well within their own quality parameters (still scoring 9/10).

Issue #170, opened 2017-05-17
Add javadoc around ByteStream and Environment


DavyLandman:
I'm trying out metal, and found some parts less documented than others.

undocumented:
- `ByteStream` (for example, what should be the return value for the read when it reads past the EOF?)
- `Environment`
- `ParseGraph`
- `Source`
- `CallBacks`

next to it, there is no reference implementation for`ByteStream`.

jvdb:
Will be added for the upcoming 6.0 release.

A basic implementation for `ByteStream` exists in the `util` package in the test code of the `core` component. It is exposed in a test-jar.

PR #282, opened 2019-02-10
Convert Value to interface, extract functionality to CoreValue and create a NotAValue class

### LINKED ISSUE #280: Create a clean implementation for NOT_A_VALUE

jvdb:
This PR merges to the branch of #281. So please review that first.

The conversion in #281 was complicated by the need for `NOT_A_VALUE` to be a direct instance of `Value`, which was a class and not an interface. This is fixed in this PR. `Value` is now an interface, and `NOT_A_VALUE` is an instance of `NotAValue`, which implements `Value` but which itself cannot be instantiated (private constructor), guaranteeing the uniqueness of `NOT_A_VALUE`.

This PR resolves #280.

### DIFF
--- a/core/src/main/java/io/parsingdata/metal/expression/value/CoreValue.java
+++ b/core/src/main/java/io/parsingdata/metal/expression/value/CoreValue.java
@@ -1,0 +1,99 @@
+/*
+ * Copyright 2013-2016 Netherlands Forensic Institute
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.parsingdata.metal.expression.value;
+
+import static io.parsingdata.metal.Util.bytesToHexString;
+import static io.parsingdata.metal.Util.checkNotNull;
+import static io.parsingdata.metal.encoding.Encoding.DEFAULT_ENCODING;
+
+import java.math.BigInteger;
+import java.util.BitSet;
+import java.util.Objects;
+
+import io.parsingdata.metal.Util;
+import io.parsingdata.metal.data.Slice;
+import io.parsingdata.metal.encoding.ByteOrder;
+import io.parsingdata.metal.encoding.Encoding;
+import io.parsingdata.metal.encoding.Sign;
+
+public class CoreValue implements Value {
+
+    public static final Value NOT_A_VALUE = new CoreValue(Slice.createFromBytes(new byte[]{}), DEFAULT_ENCODING);
+    public static final BigInteger TO_STRING_BYTE_COUNT = BigInteger.valueOf(4);
+
+    private final Slice slice;
+    private final Encoding encoding;
+
+    public CoreValue(final Slice slice, final Encoding encoding) {
+        this.slice = checkNotNull(slice, "slice");
+        this.encoding = checkNotNull(encoding, "encoding");
+    }
+
+    @Override
+    public Slice getSlice() {
+        return slice;
+    }
+
+    @Override
+    public Encoding getEncoding() {
+        return encoding;
+    }
+
+    @Override
+    public byte[] getValue() {
+        return slice.getData();
+    }
+
+    @Override
+    public BigInteger getLength() {
+        return slice.length;
+    }
+
+    @Override
+    public BigInteger asNumeric() {
+        return encoding.sign == Sign.SIGNED ? new BigInteger(encoding.byteOrder.apply(getValue()))
+                                            : new BigInteger(1, encoding.byteOrder.apply(getValue()));
+    }
+
+    @Override
+    public String asString() {
+        return new String(getValue(), encoding.charset);
+    }
+
+    @Override
+    public BitSet asBitSet() {
+        return BitSet.valueOf(encoding.byteOrder == ByteOrder.BIG_ENDIAN ? ByteOrder.LITTLE_ENDIAN.apply(getValue()) : getValue());
+    }
+
+    @Override
+    public String toString() {
+        return "0x" + bytesToHexString(slice.getData(TO_STRING_BYTE_COUNT)) + (getLength().compareTo(TO_STRING_BYTE_COUNT) > 0 ? "..." : "");
+    }
+
+    @Override
+    public boolean equals(final Object obj) {
+        return Util.notNullAndSameClass(this, obj)
+            && Objects.equals(slice, ((CoreValue)obj).slice)
+            && Objects.equals(encoding, ((CoreValue)obj).encoding);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(getClass(), slice, encoding);
+    }
+
+}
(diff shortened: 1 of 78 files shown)


core/src/main/java/io/parsingdata/metal/expression/value/NotAValue.java
```diff
@@ -0,0 +1,57 @@
...
+import java.util.Objects;
+
+import io.parsingdata.metal.Util;
+import io.parsingdata.metal.data.Slice;
+import io.parsingdata.metal.encoding.Encoding;
+
+public final class NotAValue implements Value {
+
+    public static final Value NOT_A_VALUE = new NotAValue();
+    private static final UnsupportedOperationException UNSUPPORTED = new UnsupportedOperationException("NOT_A_VALUE does not support any Value operation.");
```
rdvdijk: Just wondering, because I've never "cached" an exception in a constant like this: what does the stack trace look like when this exception is thrown from one of the methods below?

core/src/main/java/io/parsingdata/metal/expression/value/NotAValue.java
```diff
@@ -0,0 +1,57 @@
...
+
+import java.math.BigInteger;
+import java.util.BitSet;
+import java.util.Objects;
+
+import io.parsingdata.metal.Util;
+import io.parsingdata.metal.data.Slice;
+import io.parsingdata.metal.encoding.Encoding;
+
+public final class NotAValue implements Value {
```
rdvdijk: Add Javadoc.

core/src/main/java/io/parsingdata/metal/expression/value/Value.java
```diff
@@ -16,69 +16,20 @@
...
-        return "0x" + bytesToHexString(slice.getData(TO_STRING_BYTE_COUNT)) + (getLength().compareTo(TO_STRING_BYTE_COUNT) > 0 ? "..." : "");
-    }
 
-    @Override
-    public boolean equals(final Object obj) {
-        return Util.notNullAndSameClass(this, obj)
-            && Objects.equals(slice, ((Value)obj).slice)
-            && Objects.equals(encoding, ((Value)obj).encoding);
-    }
+public interface Value {
```
rdvdijk: Add Javadoc here, and also add Javadoc on the methods in this interface.

core/src/main/java/io/parsingdata/metal/expression/value/Value.java
```diff
@@ -16,69 +16,20 @@
...
 
-    @Override
-    public int hashCode() {
-        return Objects.hash(getClass(), slice, encoding);
-    }
+    Slice getSlice();
```
rdvdijk: We discussed this before, I think, but I'd like to see shorter names here, and just use `slice()`, `encoding()`, `value()`, and `length()`. We can do this in a separate issue.

core/src/main/java/io/parsingdata/metal/expression/value/NotAValue.java
ccreeten: Good one, pretty sure the stack trace is filled at construction time. This behaviour could be overridden of course. But I would instead just add a method throwing this exception.

core/src/main/java/io/parsingdata/metal/expression/value/Value.java
ccreeten: I agree. Also add spaces between the methods? :smile:

core/src/main/java/io/parsingdata/metal/expression/value/NotAValue.java
```diff
@@ -0,0 +1,57 @@
...
+    @Override public BitSet asBitSet() { throw UNSUPPORTED; }
+
+    @Override
+    public String toString() {
+        return "NOT_A_VALUE";
+    }
+
+    @Override
+    public boolean equals(final Object obj) {
+        return Util.notNullAndSameClass(this, obj);
```
mvanaken: I think it could be a very intersting discussion wheter `NotAValue` should be equal to itself, which in this implementation it is. `NaN` compared to itself results into false. However, there are different opions about this if this is wrong or not. Still, many languages implements it this way.

We may consider to include a method `isValid()` or `isNotAValue()` in the `Value` and let `CoreValue` and `NotAValue` implement it. Then this method here could always return false and as such mimics the same behaviour as `NaN`. I'm not saying we should, but it is worth a thought. :)

jvdb: I've implemented that `NotAValue == NotAValue` results in `false`, but that is in the `Eq` operator, which is at the language-level of Metal, where these NaN-semantics are. At the Java implementation-level, `NOT_A_VALUE.equals(NOT_A_VALUE)` _must_ return `true` otherwise it breaks the defined semantics of all `Object`-implementations. For example, otherwise it would wreck maps that depend on the `equals(...)`-method to determine whether they already have a key.

core/src/main/java/io/parsingdata/metal/expression/value/CoreValue.java
```diff
@@ -0,0 +1,98 @@
...
+import java.util.BitSet;
+import java.util.Objects;
+
+import io.parsingdata.metal.Util;
+import io.parsingdata.metal.data.Slice;
+import io.parsingdata.metal.encoding.ByteOrder;
+import io.parsingdata.metal.encoding.Encoding;
+import io.parsingdata.metal.encoding.Sign;
+
+public class CoreValue implements Value {
```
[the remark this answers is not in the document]
jvdb: I chose this name because it contains the _core_ functionality of all regular `Value` implementations. Not just the constructed values used it, but `ParseValue` derives from it as well. So that's what I was aiming at with the "core" part.

I do agree it's difficult to find a good name. The only other one I could come up with was `ValueImpl` (which is normally ok, but not for something that actually gets instantiated like this). I'm open to changing names if we discover something that really fits well.

core/src/main/java/io/parsingdata/metal/expression/value/NotAValue.java
jvdb: The JLS indeed specifies that it is filled in when initialized. So I will fix this!

core/src/main/java/io/parsingdata/metal/expression/value/CoreValue.java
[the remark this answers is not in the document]
ccreeten: DIfficult. `AnyValue`? `JustAValue`?

core/src/main/java/io/parsingdata/metal/expression/value/NotAValue.java
jvdb: Agreed, but `CoreValue`, `Value`, etc. and many other mostly internal classes don't have JavaDoc either. So let's pick this up as part of #170.

core/src/main/java/io/parsingdata/metal/expression/value/Value.java
jvdb: I fixed both the names and the spacing. In some other places as well by the way, so it should be universal now that there is a newline _between_ methods :)

core/src/main/java/io/parsingdata/metal/expression/value/CoreValue.java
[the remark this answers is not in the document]
jvdb: Let's revisit this if the name really bothers us going forward :)

PR #283, opened 2019-02-11
Add SingleValueExpression to type ValueExpressions that always return a single value

### LINKED ISSUE #70: Add type mechanism for single value list expressions

jvdb:
This PR merges to the branch of #282. So please review that first.

Introduces `SingleValueExpression` as proposed in #70.

Instead of the `ImmutableList<Value>` that is returned by `ValueExpression.eval(...)`, `SingleValueExpression` has a `evalSingle(...)` method with the same arguments that returns an `Optional<Value>`. `SingleValueExpression` extends `ValueExpression`, so that it can be used wherever a `ValueExpression` is required.

The default implementation of the `eval(...)` method on `SingleValueExpression` converts an empty result of `evalSingle(...)` to an empty list, otherwise the returned value is stored as single value in the returned list.

Converted to `SingleValueExpression` are:

1. `Token` arguments:
   1. the `size` argument of `Def`
   2. the `size` argument of `Nod` (essentially just a proxy for `Def`)
   2. the `n` argument of `RepN`
2. `ValueExpression` implementations:
   1. `CurrentOffset`
   2. `CurrentIteration`
   3. `Const`
   4. `First`
   5. `Last`
   6. `Count`
   7. `Self`
3. `ValueExpression` arguments:
   1. the `count` argument of `Expand`
   2. the `initial` argument of `Fold`

Tasks:
 - [x] The `Fold` implementations themselves should also be converted, not just (some of) the argument(s).
 - [x] Resolve static analysis tools results.

The unaddressed tasks that was in the list concerning refactoring the type of `Fold`'s `reducer` argument now has its own issue: #284.

This PR resolves #70.

### DIFF
--- a/core/src/main/java/io/parsingdata/metal/Shorthand.java
+++ b/core/src/main/java/io/parsingdata/metal/Shorthand.java
@@ -91,7 +91,7 @@ import io.parsingdata.metal.token.While;
 public final class Shorthand {
 
     public static final Token EMPTY = def(EMPTY_NAME, 0L);
-    public static final ValueExpression SELF = new Self();
+    public static final SingleValueExpression SELF = new Self();
     public static final SingleValueExpression CURRENT_OFFSET = new CurrentOffset();
     public static final SingleValueExpression CURRENT_ITERATION = new CurrentIteration(con(0));
     public static final Expression TRUE = new True();
--- a/core/src/main/java/io/parsingdata/metal/Shorthand.java
+++ b/core/src/main/java/io/parsingdata/metal/Shorthand.java
@@ -200,11 +200,11 @@ public final class Shorthand {
     public static ValueExpression elvis(final ValueExpression left, final ValueExpression right) { return new Elvis(left, right); }
     public static SingleValueExpression count(final ValueExpression operand) { return new Count(operand); }
     public static ValueExpression foldLeft(final ValueExpression values, final BinaryOperator<ValueExpression> reducer) { return new FoldLeft(values, reducer, null); }
-    public static ValueExpression foldLeft(final ValueExpression values, final BinaryOperator<ValueExpression> reducer, final ValueExpression initial) { return new FoldLeft(values, reducer, initial); }
+    public static ValueExpression foldLeft(final ValueExpression values, final BinaryOperator<ValueExpression> reducer, final SingleValueExpression initial) { return new FoldLeft(values, reducer, initial); }
     public static ValueExpression foldRight(final ValueExpression values, final BinaryOperator<ValueExpression> reducer) { return new FoldRight(values, reducer, null); }
-    public static ValueExpression foldRight(final ValueExpression values, final BinaryOperator<ValueExpression> reducer, final ValueExpression initial) { return new FoldRight(values, reducer, initial); }
+    public static ValueExpression foldRight(final ValueExpression values, final BinaryOperator<ValueExpression> reducer, final SingleValueExpression initial) { return new FoldRight(values, reducer, initial); }
     public static ValueExpression fold(final ValueExpression values, final BinaryOperator<ValueExpression> reducer) { return foldRight(values, reducer); }
-    public static ValueExpression fold(final ValueExpression values, final BinaryOperator<ValueExpression> reducer, final ValueExpression initial) { return foldRight(values, reducer, initial); }
+    public static ValueExpression fold(final ValueExpression values, final BinaryOperator<ValueExpression> reducer, final SingleValueExpression initial) { return foldRight(values, reducer, initial); }
     public static ValueExpression rev(final ValueExpression values) { return new Reverse(values); }
     public static ValueExpression exp(final ValueExpression base, final SingleValueExpression count) { return new Expand(base, count); }
     public static BinaryValueExpression mapLeft(final BiFunction<ValueExpression, ValueExpression, BinaryValueExpression> func, final ValueExpression left, final ValueExpression rightExpand) { return func.apply(left, exp(rightExpand, count(left))); }
--- a/core/src/main/java/io/parsingdata/metal/Shorthand.java
+++ b/core/src/main/java/io/parsingdata/metal/Shorthand.java
@@ -196,15 +196,15 @@ public final class Shorthand {
     public static SingleValueExpression iteration(final int level) { return iteration(con(level)); }
     public static SingleValueExpression iteration(final ValueExpression level) { return new CurrentIteration(level); }
     public static ValueExpression cat(final ValueExpression left, final ValueExpression right) { return new Cat(left, right); }
-    public static ValueExpression cat(final ValueExpression operand) { return new FoldCat(operand); }
+    public static SingleValueExpression cat(final ValueExpression operand) { return new FoldCat(operand); }
     public static ValueExpression elvis(final ValueExpression left, final ValueExpression right) { return new Elvis(left, right); }
     public static SingleValueExpression count(final ValueExpression operand) { return new Count(operand); }
-    public static ValueExpression foldLeft(final ValueExpression values, final BinaryOperator<ValueExpression> reducer) { return new FoldLeft(values, reducer, null); }
-    public static ValueExpression foldLeft(final ValueExpression values, final BinaryOperator<ValueExpression> reducer, final SingleValueExpression initial) { return new FoldLeft(values, reducer, initial); }
-    public static ValueExpression foldRight(final ValueExpression values, final BinaryOperator<ValueExpression> reducer) { return new FoldRight(values, reducer, null); }
-    public static ValueExpression foldRight(final ValueExpression values, final BinaryOperator<ValueExpression> reducer, final SingleValueExpression initial) { return new FoldRight(values, reducer, initial); }
-    public static ValueExpression fold(final ValueExpression values, final BinaryOperator<ValueExpression> reducer) { return foldRight(values, reducer); }
-    public static ValueExpression fold(final ValueExpression values, final BinaryOperator<ValueExpression> reducer, final SingleValueExpression initial) { return foldRight(values, reducer, initial); }
+    public static SingleValueExpression foldLeft(final ValueExpression values, final BinaryOperator<ValueExpression> reducer) { return new FoldLeft(values, reducer, null); }
+    public static SingleValueExpression foldLeft(final ValueExpression values, final BinaryOperator<ValueExpression> reducer, final SingleValueExpression initial) { return new FoldLeft(values, reducer, initial); }
+    public static SingleValueExpression foldRight(final ValueExpression values, final BinaryOperator<ValueExpression> reducer) { return new FoldRight(values, reducer, null); }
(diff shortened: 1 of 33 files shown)


core/src/main/java/io/parsingdata/metal/expression/value/SingleValueExpression.java
```diff
@@ -0,0 +1,46 @@
...
+
+import io.parsingdata.metal.data.ImmutableList;
+import io.parsingdata.metal.data.ParseState;
+import io.parsingdata.metal.encoding.Encoding;
+
+/**
+ * Interface for all SingleValueExpression implementations.
+ * <p>
+ * A SingleValueExpression is an expression that is evaluated by executing its
+ * {@link #eval(ParseState, Encoding)} method. It yields an {@link Optional}
```
rdvdijk: ```#evalSingle```

core/src/main/java/io/parsingdata/metal/expression/value/reference/Self.java
```diff
@@ -16,22 +16,23 @@
...
  * added to the parse state.
  */
-public class Self implements ValueExpression {
+public class Self implements SingleValueExpression {
 
     @Override
-    public ImmutableList<Value> eval(final ParseState parseState, final Encoding encoding) {
-        return parseState.order.current().map(ImmutableList::<Value>create).orElseGet(ImmutableList::new);
+    public Optional<Value> evalSingle(final ParseState parseState, final Encoding encoding) {
+        return parseState.order.current().flatMap(Optional::of);
```
rdvdijk: Ah, here is the upcast :wink:

core/src/main/java/io/parsingdata/metal/token/RepN.java
```diff
@@ -20,48 +20,42 @@
...
-            return failure();
-        }
-        final BigInteger count = counts.head.asNumeric();
-        return parse(environment, env -> env.parseState.iterations.head.right.compareTo(count) >= 0, env -> failure());
+        return n.evalSingle(environment.parseState, environment.encoding)
+            .filter(count -> !count.equals(NOT_A_VALUE))
+            .flatMap(count -> parse(environment, env -> env.parseState.iterations.head.right.compareTo(count.asNumeric()) >= 0, env -> failure()));
     }
-
+
```
rdvdijk: (Extreme nitpick: remove spaces.)

core/src/main/java/io/parsingdata/metal/expression/value/Expand.java
```diff
@@ -56,11 +58,11 @@ public Expand(final ValueExpression bases, final ValueExpression count) {
...
         }
-        final ImmutableList<Value> countList = count.eval(parseState, encoding);
-        if (countList.size != 1 || countList.head.equals(NOT_A_VALUE)) {
-            throw new IllegalArgumentException("Count must evaluate to a single non-empty value.");
+        final Optional<Value> countValue = count.evalSingle(parseState, encoding);
+        if (!countValue.isPresent() || countValue.get().equals(NOT_A_VALUE)) {
+            throw new IllegalArgumentException("Count must evaluate to a non-empty countable value.");
         }
-        return expand(baseList, countList.head.asNumeric().intValueExact(), new ImmutableList<>()).computeResult();
+        return expand(baseList, countValue.get().asNumeric().intValueExact(), new ImmutableList<>()).computeResult();
```
rdvdijk: Is it possible to do `return countValue.filter(!NOT_A_VALUE).map(expand(...)).orElseThrow()` here?

core/src/test/java/io/parsingdata/metal/ByteLengthTest.java
```diff
@@ -53,8 +54,8 @@
     // but Len will become useful when Let is implemented
     private static final Token STRING = seq(
         def("length", 1),
-        def("text1", ref("length")),
-        def("text2", len(ref("text1"))));
+        def("text1", last(ref("length"))),
+        def("text2", last(len(ref("text1")))));
```
rdvdijk: (You mentioned this offline, and I must it does create some noise...)

core/src/test/java/io/parsingdata/metal/IterateTest.java
```diff
@@ -41,11 +42,11 @@
 
     private static final Token repNToken =
         seq(any("n"),
-            repn(def("x", con(1), gtNum(con(1))), ref("n")),
+            repn(def("x", con(1), gtNum(con(1))), last(ref("n"))),
             def("f", con(1), eq(con(42))));
 
     private static final Token repBrokenNToken =
-        seq(repn(any("x"), div(con(1), con(0))),
+        seq(repn(any("x"), last(div(con(1), con(0)))),
```
rdvdijk: (This looks pretty confusing, to be honest, even though it makes total sense.)

jvdb: Agreed! This is one of those cases where we'd like to have `SingleValueExpression`-_versions_ of common operations such as the logical and math operators.

core/src/main/java/io/parsingdata/metal/Shorthand.java
```diff
@@ -90,22 +91,22 @@
...
+    public static final SingleValueExpression CURRENT_ITERATION = new CurrentIteration(con(0));
     public static final Expression TRUE = new True();
 
     private Shorthand() {}
 
-    public static Token def(final String name, final ValueExpression size, final Expression predicate, final Encoding encoding) { return post(def(name, size, encoding), predicate); }
-    public static Token def(final String name, final ValueExpression size, final Expression predicate) { return def(name, size, predicate, null); }
-    public static Token def(final String name, final ValueExpression size, final Encoding encoding) { return new Def(name, size, encoding); }
-    public static Token def(final String name, final ValueExpression size) { return def(name, size, (Encoding)null); }
+    public static Token def(final String name, final SingleValueExpression size, final Expression predicate, final Encoding encoding) { return post(def(name, size, encoding), predicate); }
```
ccreeten: Now I see this, and I can't recall why: what is the reason we do not allow a list of sizes? Apart from that it could be written down differently (and better?), and that there are not really use cases for it then. Is it an arbitrary choice? Just curious, I think it makes sense of course.

core/src/main/java/io/parsingdata/metal/Shorthand.java
```diff
@@ -172,40 +173,40 @@ private Shorthand() {}
...
+    public static SingleValueExpression con(final long value, final Encoding encoding) { return con(ConstantFactory.createFromNumeric(value, encoding)); }
+    public static SingleValueExpression con(final String value) { return con(value, DEFAULT_ENCODING); }
+    public static SingleValueExpression con(final String value, final Encoding encoding) { return con(ConstantFactory.createFromString(value, encoding)); }
+    public static SingleValueExpression con(final Value value) { return new Const(value); }
+    public static SingleValueExpression con(final Encoding encoding, final int... values) { return new Const(new CoreValue(createFromBytes(toByteArray(values)), encoding)); }
+    public static SingleValueExpression con(final int... values) { return con(DEFAULT_ENCODING, values); }
+    public static SingleValueExpression con(final byte[] value) { return con(value, DEFAULT_ENCODING); }
+    public static SingleValueExpression con(final byte[] value, final Encoding encoding) { return con(ConstantFactory.createFromBytes(value, encoding)); }
     public static ValueExpression len(final ValueExpression operand) { return new Len(operand); }
```
ccreeten: I recall reading you didn't want to change all expressions at once, can't seem to find it now though. But to make sure: `len` will also be a single expression right?

core/src/main/java/io/parsingdata/metal/expression/value/Fold.java
```diff
@@ -31,46 +32,46 @@
...
-public abstract class Fold implements ValueExpression {
+public abstract class Fold implements SingleValueExpression {
 
     public final ValueExpression values;
     public final BinaryOperator<ValueExpression> reducer;
-    public final ValueExpression initial;
+    public final SingleValueExpression initial;
 
-    public Fold(final ValueExpression values, final BinaryOperator<ValueExpression> reducer, final ValueExpression initial) {
+    public Fold(final ValueExpression values, final BinaryOperator<ValueExpression> reducer, final SingleValueExpression initial) {
```
ccreeten: Could we sometime update the shorthand to pass an expression evaluation to empty instead of these nulls? Cleans up the implementation below as well.

core/src/main/java/io/parsingdata/metal/expression/value/reference/Nth.java
```diff
@@ -44,9 +43,9 @@
  * <code>indices</code> (both {@link ValueExpression}s). Both operands are
  * evaluated. Next, the resulting values of evaluating <code>indices</code> is
  * used as a list of integer indices into the results of evaluating
- * <code>values</code>. For every invalid index (such as
- * {@link Optional#empty()}, a negative value or an index that is out of
- * bounds) empty is returned.
+ * <code>values</code>. For every invalid index (<code>NOT_A_VALUE</code>, a
```
ccreeten: `{@link NotAValue#NOT_A_VALUE}`?

core/src/main/java/io/parsingdata/metal/expression/value/reference/Self.java
ccreeten: Not `.map(identity())`? :stuck_out_tongue_closed_eyes:

core/src/test/java/io/parsingdata/metal/ByteLengthTest.java
ccreeten: For len it will dissappear I guess. For ref, well... Java does support return type overloading on the bytecode level :smiling_imp:

core/src/main/java/io/parsingdata/metal/Shorthand.java
jvdb: I think we just decided that if you want a list of `def`s, you just need to use `seq` (with `repn` as dynamic alternative).

core/src/main/java/io/parsingdata/metal/Shorthand.java
jvdb: No, but you're talking about `count` 😅

ccreeten: Doh! Little bit confusing... :yum:

core/src/main/java/io/parsingdata/metal/expression/value/reference/Self.java
jvdb: @ccreeten: Oh I actually like that one better...

core/src/main/java/io/parsingdata/metal/expression/value/Expand.java
jvdb: Wow indeed!

core/src/main/java/io/parsingdata/metal/expression/value/Fold.java
jvdb: This is part of #187, but I agree it's a reasonably small part and can quickly make things better.

PR #289, opened 2019-03-26
Add a Scope ValueExpression to handle scoping aspects of predicates

### LINKED ISSUE #250: Add a "locally scoped" ref to avoid unpredictable references in recursive Token definitions

jvdb:
Resolves #250.

### DIFF
--- a/core/src/main/java/io/parsingdata/metal/expression/value/Scope.java
+++ b/core/src/main/java/io/parsingdata/metal/expression/value/Scope.java
@@ -1,0 +1,88 @@
+/*
+ * Copyright 2013-2019 Netherlands Forensic Institute
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.parsingdata.metal.expression.value;
+
+import static java.math.BigInteger.ZERO;
+
+import static io.parsingdata.metal.Trampoline.complete;
+import static io.parsingdata.metal.Trampoline.intermediate;
+import static io.parsingdata.metal.Util.checkNotNull;
+import static io.parsingdata.metal.expression.value.NotAValue.NOT_A_VALUE;
+
+import java.util.Objects;
+
+import io.parsingdata.metal.Trampoline;
+import io.parsingdata.metal.Util;
+import io.parsingdata.metal.data.ImmutableList;
+import io.parsingdata.metal.data.ParseGraph;
+import io.parsingdata.metal.data.ParseState;
+import io.parsingdata.metal.encoding.Encoding;
+
+public class Scope implements ValueExpression {
+
+    public final ValueExpression scopedValueExpression;
+    public final SingleValueExpression scopeSize;
+
+    public Scope(final ValueExpression scopedValueExpression, final SingleValueExpression scopeSize) {
+        this.scopedValueExpression = checkNotNull(scopedValueExpression, "scopedValueExpression");
+        this.scopeSize = checkNotNull(scopeSize, "scopeSize");
+    }
+
+    @Override
+    public ImmutableList<Value> eval(final ParseState parseState, final Encoding encoding) {
+        return scopeSize.evalSingle(parseState, encoding)
+            .filter(sizeValue -> !sizeValue.equals(NOT_A_VALUE) && sizeValue.asNumeric().compareTo(ZERO) >= 0)
+            .map(sizeValue -> scopedValueExpression.eval(parseState.withOrder(calculateScope(parseState.order, sizeValue.asNumeric().intValueExact())), encoding))
+            .orElseThrow(() -> new IllegalArgumentException("Argument scopeSize must evaluate to a positive, non-empty countable value."));
+
+    }
+
+    private ParseGraph calculateScope(final ParseGraph order, final int size) {
+        final ImmutableList<ParseGraph> scopeList = createScopeList(order, new ImmutableList<>()).computeResult();
+        if (size >= scopeList.size) { return order; }
+        return findScope(scopeList, size).computeResult();
+    }
+
+    private Trampoline<ImmutableList<ParseGraph>> createScopeList(final ParseGraph order, final ImmutableList<ParseGraph> list) {
+        final ImmutableList<ParseGraph> newList = order.getDefinition().isScopeDelimiter() ? list.add(order) : list;
+        if (!order.head.isGraph()) { return complete(() -> list); }
+        return intermediate(() -> createScopeList(order.head.asGraph(), newList));
+    }
+
+    private Trampoline<ParseGraph> findScope(final ImmutableList<ParseGraph> scopeList, final int size) {
+        if (size == 0) { return complete(() -> scopeList.head); }
+        return intermediate(() -> findScope(scopeList, size - 1));
+    }
+
+    @Override
+    public String toString() {
+        return getClass().getSimpleName() + "(" + scopedValueExpression + "," + scopeSize + ")";
+    }
+
+    @Override
+    public boolean equals(final Object obj) {
+        return Util.notNullAndSameClass(this, obj)
+            && Objects.equals(scopedValueExpression, ((Scope)obj).scopedValueExpression)
+            && Objects.equals(scopeSize, ((Scope)obj).scopeSize);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(getClass(), scopedValueExpression, scopeSize);
+    }
+
+}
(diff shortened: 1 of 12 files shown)


DavyLandman:
Nice that we are getting the concept of scopes in metal!

In the original issue (#250) we reported some example of how you could need this, would it be possible to add these to the tests? It would make it easier for us to understand how this new feature works.

core/src/main/java/io/parsingdata/metal/expression/value/Scope.java
```diff
@@ -0,0 +1,105 @@
...
+    @Override
+    public ImmutableList<Value> eval(final ParseState parseState, final Encoding encoding) {
+        return scopeSize.evalSingle(parseState, encoding)
+            .filter(sizeValue -> !sizeValue.equals(NOT_A_VALUE) && sizeValue.asNumeric().compareTo(ZERO) >= 0)
+            .map(sizeValue -> scopedValueExpression.eval(parseState.withOrder(calculateScope(parseState.order, sizeValue.asNumeric().intValueExact())), encoding))
+            .orElseThrow(() -> new IllegalArgumentException("Argument scopeSize must evaluate to a positive, non-empty countable value."));
+
+    }
+
+    private ParseGraph calculateScope(final ParseGraph order, final int size) {
```
rdvdijk: Is `calculate` the right verb here? Maybe use `determine` or `select`?

core/src/main/java/io/parsingdata/metal/token/Token.java
```diff
@@ -75,6 +75,10 @@ protected Token(final String name, final Encoding encoding) {
 
     protected abstract Optional<ParseState> parseImpl(final Environment environment);
 
+    public boolean isScopeDelimiter() {
```
rdvdijk: I know we haven't documented the other methods here (like `isLocal()` below), but these really do need proper Javadoc. The meaning of these `boolean`s are now "hidden" in the Javadoc in the classes that need these fields. We could add some short documentation and refer to the other classes for more detailed explanations.

core/src/test/java/io/parsingdata/metal/expression/value/reference/LocalReferenceNestedScopeTest.java
```diff
@@ -77,9 +79,9 @@ public void testNestedScopesCalculated() {
         nestedScopes(eq(nth(ref("left"), sub(count(ref("left")), count(ref("right"))))));
     }
 
-//    @Test
-//    public void testNestedScopesLocalReferenced() {
-//        nestedScopes(eq(last(lref("left"))));
-//    }
+    @Test
+    public void testNestedScopesLocalReferenced() {
+        nestedScopes(eq(first(scope(ref("left"), con(0)))));
```
rdvdijk: lebowski

core/src/test/java/io/parsingdata/metal/expression/value/reference/LocalReferenceFollowingScopeTest.java
```diff
@@ -74,9 +75,9 @@ public void testFollowingScopesCalculated() {
         followingScopes(eqNum(sub(count(ref("ten")), sub(fold(ref("tenCount"), Shorthand::add), last(ref("tenCount"))))));
     }
 
-//    @Test
-//    public void testFollowingScopesLocalReferenced() {
-//        followingScopes(eqNum(count(lref("ten"))));
-//    }
+    @Test
+    public void testFollowingScopesLocalReferenced() {
+        followingScopes(eqNum(count(scope(ref("ten"), con(0)))));
```
rdvdijk: Awesomesauce!

rdvdijk:
> Nice that we are getting the concept of scopes in metal!
> 
> In the original issue (#250) we reported some example of how you could need this, would it be possible to add these to the tests? It would make it easier for us to understand how this new feature works.

Yes, I see that in #250, @jvdb already mentioned that we would add those cases as unit tests. I think we can fix those now, so adding them here would be in order.

core/src/main/java/io/parsingdata/metal/token/Token.java
ccreeten: Perhaps also adding the rationale behind the choice of when something would be a scope delimiter.

core/src/main/java/io/parsingdata/metal/expression/value/Scope.java
```diff
@@ -0,0 +1,105 @@
...
+
+/**
+ * A {@link ValueExpression} that prunes the {@link ParseState#order} field to
+ * let a nested {@link ValueExpression} have only a scoped view.
+ * <p>
+ * A Scope expression has two operands: <code>scopedValueExpression</code> (a
+ * {@link ValueExpression} and <code>scopeSize</code> (a
+ * {@link SingleValueExpression}). The <code>scopeSize</code> operand is first
+ * evaluated, if it evaluates to <code>NOT_A_VALUE</code> or a negative value
+ * an {@link IllegalArgumentException} is thrown. Otherwise the
```
ccreeten: I feel the doc does talk quite a bit in terms of how it is implemented (which could of course also change).  It is probably good here where the token implementation resides, but I would expect at the point where people use it (i.e. the Shorthand), it would describe more about what it does, what the use case is and how you use it, with perhaps an example of how it can be used. But since there is no doc in the Shorthand yet, perhaps add here (though ideally Shorthand :smile:)? Because I think it would make it more clear how to use this expression, especially for people not familiar with  implementation details of Metal.

core/src/main/java/io/parsingdata/metal/expression/value/Scope.java
```diff
@@ -0,0 +1,105 @@
...
+        this.scopedValueExpression = checkNotNull(scopedValueExpression, "scopedValueExpression");
+        this.scopeSize = checkNotNull(scopeSize, "scopeSize");
+    }
+
+    @Override
+    public ImmutableList<Value> eval(final ParseState parseState, final Encoding encoding) {
+        return scopeSize.evalSingle(parseState, encoding)
+            .filter(sizeValue -> !sizeValue.equals(NOT_A_VALUE) && sizeValue.asNumeric().compareTo(ZERO) >= 0)
+            .map(sizeValue -> scopedValueExpression.eval(parseState.withOrder(calculateScope(parseState.order, sizeValue.asNumeric().intValueExact())), encoding))
+            .orElseThrow(() -> new IllegalArgumentException("Argument scopeSize must evaluate to a positive, non-empty countable value."));
```
ccreeten: Truly minor, but we don't really deal in empty values any more, right?

core/src/main/java/io/parsingdata/metal/expression/value/Scope.java
```diff
@@ -0,0 +1,105 @@
...
+
+    }
+
+    private ParseGraph calculateScope(final ParseGraph order, final int size) {
+        final ImmutableList<ParseGraph> scopeList = createScopeList(order, new ImmutableList<>()).computeResult();
+        if (size >= scopeList.size) { return order; }
+        return findScope(scopeList, size).computeResult();
+    }
+
+    private Trampoline<ImmutableList<ParseGraph>> createScopeList(final ParseGraph order, final ImmutableList<ParseGraph> list) {
```
ccreeten: Really concise, neat!

core/src/test/java/io/parsingdata/metal/expression/value/ScopeTest.java
```diff
@@ -0,0 +1,84 @@
...
+    public void negativeScopeSize() {
+        thrown.expect(IllegalArgumentException.class);
+        thrown.expectMessage("Argument scopeSize must evaluate to a positive, non-empty countable value.");
+        new Scope(con(0), con(-1, signed())).eval(EMPTY_PARSE_STATE, enc());
+    }
+
+    @Test
+    public void scopeSizes() {
+        final Token scopesToken =
+            seq(any("value"),
```
ccreeten: Could you add a couple more simple tests and/or some doc here?

jvdb:
> Nice that we are getting the concept of scopes in metal!
> 
> In the original issue (#250) we reported some example of how you could need this, would it be possible to add these to the tests? It would make it easier for us to understand how this new feature works.

2 of the 3 examples you provided (1 and 3) are already in the unit tests! The first one, which is in the issue description of #250 is in `LocalReferenceNestedScopeTest.java`. The third one, which involves the `CheckedREP` `Token` is in `LocalReferenceFollowingScopeTest.java`. I will see if I can turn the second one into a representative and understandable unit test and add that one too.

core/src/main/java/io/parsingdata/metal/expression/value/Scope.java
jvdb: I totally agree that we need documentation that describes _what_ things do and _how_ to use them in practice. I'm not entirely sure we should make `Shorthand` even more unwieldy by adding a massive amount of Javadoc to it.

I think that what we could use best is just some user guide that explains what all the Tokens/Expressions/ValueExpressions do in combination with 1-2 practical examples for everything. Let's first discuss what we need :)

core/src/main/java/io/parsingdata/metal/expression/value/Scope.java
jvdb: I was thinking about `createScope`, but we already have `createScopeList` and `findScope` further down. I agree _calculate_ is a bit dodgy, but I tried to capture the intuition of having a graph, picking out parts and then determine which one is needed. I think _select_ matches that the best but it's a little ambiguous with _find_. _Calculate_ at least communicates that it's the thing that comes up with the final answer.

core/src/main/java/io/parsingdata/metal/token/Token.java
jvdb: Agreed again on rationale, but like I said earlier, we really need more comprehensive documentation for users of the entire library. I would rather focus on that than make lots of Javadoc.

core/src/test/java/io/parsingdata/metal/expression/value/reference/LocalReferenceNestedScopeTest.java
jvdb: I'm a little hesitant to make this resolved since I like scrolling past it so much :sweat_smile:

PR #331, opened 2020-11-25
Implemented memoization to improve cycle detection performance.

### LINKED ISSUE #330: Improve the algorithm used by the Sub Token to detect potential cycles
In files with a large amount of pointer dereferences and especially a wide structure, cycle detection slows parsing down considerably. The current implementation does not cache any intermediate results, so every time a Sub is parsed, all possible locations that might have been jumped to before and enumerated.

jvdb:
Resolves #330.

This change boils down to the addition of `CycleToken` and the handling of it: maintaining a collection of `ParseReference`s in the `ParseState` and checking that collection instead of searching by the `Sub` token.

BTW: BetterCodeHub complains about some added complexity, but I propose to not address that in this PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,8 +10,6 @@ jdk:
   - oraclejdk11
   - openjdk11
 
-before_install: pip install --user codecov
-
 script: mvn clean verify sonar:sonar -Dsonar.projectKey=parsingdata_metal
 
-after_success: codecov
+after_success: bash <(curl -s https://codecov.io/bash)
--- a/core/src/main/java/io/parsingdata/metal/data/ParseState.java
+++ b/core/src/main/java/io/parsingdata/metal/data/ParseState.java
@@ -101,7 +101,7 @@ public class ParseState {
     @Override
     public String toString() {
         final String iterationString = iterations.isEmpty() ? "" : ";iterations:" + iterations.toString();
-        final String targetsString = references.isEmpty() ? "" : ";targets:" + references.toString();
+        final String targetsString = references.isEmpty() ? "" : ";references:" + references.toString();
         return getClass().getSimpleName() + "(source:" + source + ";offset:" + offset + ";order:" + order + iterationString + targetsString + ")";
     }
 
--- a/core/src/main/java/io/parsingdata/metal/data/ParseState.java
+++ b/core/src/main/java/io/parsingdata/metal/data/ParseState.java
@@ -66,6 +66,10 @@ public class ParseState {
         return new ParseState(order.closeBranch(), source, offset, token.isIterable() ? iterations.tail : iterations, references);
     }
 
+    public ParseState addCycleReference(final ParseReference parseReference) {
+        return new ParseState(order, source, offset, iterations, references.add(parseReference));
+    }
+
     public ParseState add(final ParseValue parseValue) {
         return new ParseState(order.add(parseValue), source, offset, iterations, references);
     }
@@ -88,10 +92,6 @@ public class ParseState {
 
     public ParseState withSource(final Source source) {
         return new ParseState(order, source, ZERO, iterations, references);
-    }
-
-    public ParseState withReferences(final ImmutableList<ParseReference> references) {
-        return new ParseState(order, source, offset, iterations, references);
     }
 
     public Optional<Slice> slice(final BigInteger length) {
--- a/core/src/main/java/io/parsingdata/metal/data/ParseState.java
+++ b/core/src/main/java/io/parsingdata/metal/data/ParseState.java
@@ -66,7 +66,7 @@ public class ParseState {
         return new ParseState(order.closeBranch(), source, offset, token.isIterable() ? iterations.tail : iterations, references);
     }
 
-    public ParseState addCycleReference(final ParseReference parseReference) {
+    public ParseState add(final ParseReference parseReference) {
         return new ParseState(order, source, offset, iterations, references.add(parseReference));
     }
 
@@ -74,7 +74,7 @@ public class ParseState {
         return new ParseState(order.add(parseValue), source, offset, iterations, references);
     }
 
-    public ParseState add(final ParseReference parseReference) {
+    public ParseState createCycle(final ParseReference parseReference) {
         return new ParseState(order.add(parseReference), source, offset, iterations, references);
     }
 
--- a/core/src/main/java/io/parsingdata/metal/data/ParseState.java
+++ b/core/src/main/java/io/parsingdata/metal/data/ParseState.java
@@ -100,9 +100,9 @@ public class ParseState {
 
     @Override
     public String toString() {
-        final String iterationString = iterations.isEmpty() ? "" : ";iterations:" + iterations.toString();
-        final String targetsString = references.isEmpty() ? "" : ";references:" + references.toString();
-        return getClass().getSimpleName() + "(source:" + source + ";offset:" + offset + ";order:" + order + iterationString + targetsString + ")";
+        final String iterationsString = iterations.isEmpty() ? "" : ";iterations:" + iterations.toString();
+        final String referencesString = references.isEmpty() ? "" : ";references:" + references.toString();
+        return getClass().getSimpleName() + "(source:" + source + ";offset:" + offset + ";order:" + order + iterationsString + referencesString + ")";
     }
 
     @Override
--- a/core/src/main/java/io/parsingdata/metal/data/ParseState.java
+++ b/core/src/main/java/io/parsingdata/metal/data/ParseState.java
@@ -37,16 +37,18 @@ public class ParseState {
     public final BigInteger offset;
     public final Source source;
     public final ImmutableList<ImmutablePair<Token, BigInteger>> iterations;
+    public final ImmutableList<ParseReference> references;
 
-    public ParseState(final ParseGraph order, final Source source, final BigInteger offset, final ImmutableList<ImmutablePair<Token, BigInteger>> iterations) {
+    public ParseState(final ParseGraph order, final Source source, final BigInteger offset, final ImmutableList<ImmutablePair<Token, BigInteger>> iterations, final ImmutableList<ParseReference> references) {
         this.order = checkNotNull(order, "order");
         this.source = checkNotNull(source, "source");
         this.offset = checkNotNegative(offset, "offset");
         this.iterations = checkNotNull(iterations, "iterations");
+        this.references = checkNotNull(references, "references");
     }
 
     public static ParseState createFromByteStream(final ByteStream input, final BigInteger offset) {
-        return new ParseState(ParseGraph.EMPTY, new ByteStreamSource(input), offset, new ImmutableList<>());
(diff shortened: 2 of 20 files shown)


core/src/main/java/io/parsingdata/metal/data/ParseState.java
```diff
@@ -95,7 +101,8 @@ public ParseState withSource(final Source source) {
     @Override
     public String toString() {
         final String iterationString = iterations.isEmpty() ? "" : ";iterations:" + iterations.toString();
-        return getClass().getSimpleName() + "(source:" + source + ";offset:" + offset + ";order:" + order + iterationString + ")";
+        final String targetsString = references.isEmpty() ? "" : ";references:" + references.toString();
```
rdvdijk, line 104: I guess that the 'references' concept was renamed somewhere along the line? I would expect `referencesString` here.

core/src/main/java/io/parsingdata/metal/token/CycleToken.java
```diff
@@ -0,0 +1,47 @@
...
+    /**
+     * Parse this {@link Token}. Adds itself to the list of references maintained in the {@link Environment}'s
+     * {@link ParseState} object to quickly allow detection of potential cycles.
+     *
+     * @param environment the environment to apply the parse to
+     * @return a non-empty {@code Optional<ParseState>} if successful, otherwise {@code Optional.empty()}
+     */
+    @Override
+    public Optional<ParseState> parse(Environment environment) {
+        return super.parse(environment.withParseState(environment.parseState.withReferences(environment.parseState.references.add(new ParseReference(environment.parseState.offset, environment.parseState.source, this.getCanonical(environment.parseState))))));
```
rdvdijk, line 45: That's a pretty long line! Maybe reformat this a bit?

My proposal, but maybe I'm talking it too far in the other direction :wink: 

```java
         return super.parse(
             environment.withParseState(
                 environment.parseState.withReferences(
                     environment.parseState.references.add(
                         new ParseReference(
                             environment.parseState.offset, 
                             environment.parseState.source, 
                             this.getCanonical(environment.parseState)
                         )
                     )
                 )
             )
         );
```

core/src/main/java/io/parsingdata/metal/token/CycleToken.java
```diff
@@ -0,0 +1,47 @@
...
+import java.util.Optional;
+
+import io.parsingdata.metal.data.Environment;
+import io.parsingdata.metal.data.ParseReference;
+import io.parsingdata.metal.data.ParseState;
+import io.parsingdata.metal.encoding.Encoding;
+
+/**
+ * Base class for {@link Token} implementations that can potentially
+ * cause a cycle if used in conjunction with {@link Sub}.
```
rdvdijk, line 28: What does `in conjunction with Sub` mean?

I understand the change of the PR, but this Javadoc seems a bit vague.

Maybe we can elaborate more what the `CycleToken`s typically are, or how they behave, maybe something like "tokens that contain other tokens" or "tokens that use other tokens as building blocks" (also pretty vague, I admit).

rdvdijk: And related to that: Which tokens are *not* `CycleToken`s?

core/src/test/java/io/parsingdata/metal/ToStringTest.java
```diff
@@ -203,6 +207,19 @@ public void callback() {
...
+    @Test
+    public void toStringOnParseStateCollections() {
+        final ParseState parseState = stream(1, 2);
+        assertFalse(parseState.toString().contains(";iterations:"));
+        assertFalse(parseState.toString().contains(";references:"));
+        final ImmutableList<ImmutablePair<Token, BigInteger>> iterationsList = ImmutableList.create(new ImmutablePair<>(t(), BigInteger.ZERO));
+        final ParseState parseStateWithIteration = new ParseState(parseState.order, parseState.source, parseState.offset, iterationsList, new ImmutableList<>());
+        assertTrue(parseStateWithIteration.toString().contains(";iterations:" + iterationsList.toString()));
+        final ImmutableList<ParseReference> referencesList = ImmutableList.create(new ParseReference(BigInteger.ZERO, parseState.source, t()));
+        final ParseState parseStateWithReference = new ParseState(parseState.order, parseState.source, parseState.offset, new ImmutableList<>(),referencesList);
```
rdvdijk, line 219: **code style check engage**
```java
,referencesList
```
Should be:
```java
, referencesList
```
**code style check disengage**

core/src/test/java/io/parsingdata/metal/data/SelectionTest.java
```diff
@@ -42,12 +42,21 @@
         @Override protected boolean isAvailable(BigInteger offset, BigInteger length) { return true; }
     };
 
+    private final Source otherSource = new Source() {
+        @Override protected byte[] getData(BigInteger offset, BigInteger length) { return new byte[0]; }
+        @Override protected boolean isAvailable(BigInteger offset, BigInteger length) { return true; }
+    };
```
rdvdijk, line 48: Minor, just wondering: is this the first time we need an "other" source? If so, maybe we can reuse it here.

core/src/test/java/io/parsingdata/metal/data/selection/ByTokenTest.java
```diff
@@ -239,9 +239,9 @@ public void getSubRef() {
         final Optional<ParseState> result = composition.parse(env(stream(0)));
         assertTrue(result.isPresent());
         final ImmutableList<ParseItem> items = getAll(result.get().order, DEF2);
-        // should return the ParseGraph created by the Sub and the ParseReference that refers to the existing ParseItem
+        // should return two values created by the Sub, no ParseReference because a Def cannot create a cycle
```
rdvdijk, line 242: Nice! :+1:

core/src/main/java/io/parsingdata/metal/token/CycleToken.java
mvanaken, line 45: This is a really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really really long line. ;) ~~Maybe create a method on the environment instead? E.g. `environment.addReference(new ParseReference...)`, which does this logic, because you really have to know the environment here to understand what is happening here. The method in environment can get a more explanatory name.~~ Can't `add(ParseReference)` within ParseState can be reused?

core/src/main/java/io/parsingdata/metal/data/ParseState.java
```diff
@@ -37,55 +37,61 @@
...
-        return new ParseState(order.closeBranch(), source, offset, token.isIterable() ? iterations.tail : iterations);
+        return new ParseState(order.closeBranch(), source, offset, token.isIterable() ? iterations.tail : iterations, references);
     }
 
     public ParseState add(final ParseValue parseValue) {
-        return new ParseState(order.add(parseValue), source, offset, iterations);
+        return new ParseState(order.add(parseValue), source, offset, iterations, references);
     }
 
     public ParseState add(final ParseReference parseReference) {
```
mvanaken, line 73: Can't this the method where we actually add a parseReference to the references list? I know this is only used within Sub, but we could use this also within the CycleToken, I think.

mvanaken (commented):
I like how minimal the change actually is. :) I still think we can make it a little more readable and am curious if we can reuse the `add(ParseReference)` within `ParseState` for this. That would also make the method `withReferences` redundant.

core/src/main/java/io/parsingdata/metal/data/ParseState.java
jvdb: (This is a response to both your comments!) The problem is that not every invocation of `ParseState.add(...)` should lead to the creation of a `ParseReference` to add to the `references` collection. This is something specific to when a `CycleToken` is parsed, so that's why it happens there. For example a `Def` is not a `CycleToken` (because it cannot create a cycle) so adding a `Def` to the `ParseState` should not lead to adding something to the `references` collection.

I do totally agree though that the line in `CycleToken.parse(...)` looks quite ridiculous, so I will try to see if there is some way to shorten that in an intuitive manner that doesn't move `CycleToken`'s responsibility somewhere where it doesn't belong.

core/src/test/java/io/parsingdata/metal/data/SelectionTest.java
jvdb: There's one other use but it's a little different. Instead I've extracted the construction into a method so that it's no longer a duplicate blob.

core/src/main/java/io/parsingdata/metal/token/CycleToken.java
jvdb: I've reformatted it, but @mvanaken below suggests to maybe integrate all these chained helper invocations into something a little easier to follow (and shorter!). So it might change into something better 😅

core/src/main/java/io/parsingdata/metal/token/CycleToken.java
jvdb: Before you know it, we're back in the discussion around whether it's actually a non-terminal or not. But I agree it should be explained to the user here!

jvdb: I've reworded the description and expanded it to explicitly explain that all non-terminal `Token`s are of this type except `TokenRef` and `Tie` along with the rationale for that.

.travis.yml
```diff
@@ -10,8 +10,6 @@ jdk:
   - oraclejdk11
   - openjdk11
 
-before_install: pip install --user codecov
-
 script: mvn clean verify sonar:sonar -Dsonar.projectKey=parsingdata_metal
 
-after_success: codecov
+after_success: bash <(curl -s https://codecov.io/bash)
```
rdvdijk, line 15: Saw your comments [here](https://community.codecov.io/t/weird-ssl-error-from-codecov-python-when-uploading-from-travis-ci/2521). Weird!

rdvdijk (approved):

core/src/main/java/io/parsingdata/metal/token/CycleToken.java
jvdb: See my response below!

core/src/main/java/io/parsingdata/metal/data/ParseState.java
jvdb: I've mirrored the pattern we used on `Environment.addBranch(...)` and `ParseState.addBranch(...)`: there are now helper methods `Environment.addCycleReference(...)` and also `ParseState.addCycleReference(...)` that together compose what originally happened in `CycleToken.parse(...)`.

.travis.yml
jvdb: 😬

core/src/main/java/io/parsingdata/metal/data/ParseState.java
jvdb: We ended up improving this further by renaming the original `add(ParseReference)` to `createCycle(ParseReference)` and the `addCycleReference(...)` to just `add(...)` to better reflect what they actually do.

PR #332, opened 2020-12-17
Make ImmutableList iterable

gertjanal:
Make the ImmutableList iterable, so it can be used in for loops and streams

### DIFF
--- a/core/src/main/java/io/parsingdata/metal/data/ImmutableList.java
+++ b/core/src/main/java/io/parsingdata/metal/data/ImmutableList.java
@@ -109,8 +109,7 @@ public class ImmutableList<T> implements Iterable<T> {
 
             @Override
             public boolean hasNext() {
-                final ImmutableList<T> list = next.get();
-                return !list.isEmpty();
+                return !next.get().isEmpty();
             }
 
             @Override
--- a/core/src/main/java/io/parsingdata/metal/data/ImmutableList.java
+++ b/core/src/main/java/io/parsingdata/metal/data/ImmutableList.java
@@ -21,12 +21,14 @@ import static io.parsingdata.metal.Trampoline.intermediate;
 import static io.parsingdata.metal.Util.checkNotNull;
 import static io.parsingdata.metal.data.Selection.reverse;
 
+import java.util.Iterator;
 import java.util.Objects;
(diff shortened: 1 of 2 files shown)


core/src/test/java/io/parsingdata/metal/ImmutableListTest.java
```diff
@@ -146,6 +146,15 @@ public void sizeEmpty() {
         assertEquals(0, new ImmutableList<ParseValue>().size);
     }
 
+    @Test
+    public void iterable() {
+        final StringBuilder builder = new StringBuilder();
+        for(final ParseValue value : l5) {
+            builder.append(value.name);
+        }
+        assertEquals("abade", builder.toString());
```
gertjanal, line 155: Is this the expected order?

core/src/main/java/io/parsingdata/metal/data/ImmutableList.java
```diff
@@ -100,4 +102,27 @@ public int hashCode() {
...
+            @Override
+            public boolean hasNext() {
+                return !next.get().isEmpty();
+            }
+
+            @Override
+            public T next() {
+                final ImmutableList<T> list = next.get();
+                if (list.isEmpty()) {
+                    return null;
```
mvanaken, line 119: Or throw an exception, e.g. `NoSuchElementException`, since there is a `hasNext`method?

core/src/test/java/io/parsingdata/metal/ImmutableListTest.java
```diff
@@ -146,6 +146,15 @@ public void sizeEmpty() {
         assertEquals(0, new ImmutableList<ParseValue>().size);
     }
 
+    @Test
+    public void iterable() {
```
mvanaken, line 150: Include a test to iterate beyond the end?

mvanaken (commented):
Have a iterable list is very convenient! I think we can simplify e.g. implementations within Structure, but was there a specific usecase that you implemented this now? (if you told me, sorry I forgot...)

jvdb:
Closing this because adding this directly to Metal would only make sense if all the classes using the `ImmutableList` inside Metal would also be refactored to use this functionality. This may be a good idea but requires some thought and discussion before we put effort into doing it.

PR #335, opened 2021-02-07
Improve API usability by adding SingleValueExpression-typed versions of common ValueExpressions

### LINKED ISSUE #285: Implement SingleValueExpression versions of common ValueExpressions
Examples are the logical and math `UnaryValueExpression`s and `BinaryValueExpression`s such as `Add`, `Or` and `Not`.

Important point is whether this can be implemented cleanly without requiring too complicated handling with generics and type casting.

jvdb:
Resolves #285.

The `Shorthand` helper methods that we can safely create are the following:

1. Mathematical operators: `add`, `div`, `mul`, `sub`, `mod` and `neg`.
2. Bit-wise operators: `and`, `or`, `not`, `shl` and `shr`.
3. Reference operators: `len`, `nth`*, `offset`, `cat`** and `elvis`.

Note*: For `nth` this only concerns the second argument and the return value. The first argument needs to be a list to index into.
Note**: For `cat` this refers to the version that has two arguments. The single argument version of `cat` explicitly operates on lists.

There are some additional `ValueExpression`s that could be converted but I believe it's better if we don't create far-fetched cases. For example, the `mapLeft` and `mapRight` could get a `SingleValueExpression` as input argument (and then as a result also as return value), but if you know beforehand that you only need to apply something to a single value, why even use map?

Some additional things following from this:
 - [x] In the earlier issue introducing `SingleValueExpression` the `limit` argument of `Ref` should have been converted too but was not.
 - [x] For the `Fold` `ValueExpression`s, the `reducer` argument should probably be converted from `BinaryOperator<ValueExpression>` to `BinaryOperator<SingleValueExpression>`. This was noticed earlier but not changed due to usability concerns (we wanted to be able to use method references to the `Shorthand` helper methods here, which will stay possible with this change once we add the additional helper methods).
 - [x] Remove the added `last(...)` wrappings from test code that were added when `SingleValueExpression` was added (in #283, issue #70) that are now no longer needed.

### DIFF
--- a/core/src/main/java/io/parsingdata/metal/Shorthand.java
+++ b/core/src/main/java/io/parsingdata/metal/Shorthand.java
@@ -216,12 +216,12 @@ public final class Shorthand {
     public static ValueExpression elvis(final ValueExpression left, final ValueExpression right) { return new Elvis(left, right); }
     public static SingleValueExpression elvis(final SingleValueExpression left, final SingleValueExpression right) { return last(new Elvis(left, right)); }
     public static SingleValueExpression count(final ValueExpression operand) { return new Count(operand); }
-    public static SingleValueExpression foldLeft(final ValueExpression values, final BinaryOperator<ValueExpression> reducer) { return new FoldLeft(values, reducer, null); }
-    public static SingleValueExpression foldLeft(final ValueExpression values, final BinaryOperator<ValueExpression> reducer, final SingleValueExpression initial) { return new FoldLeft(values, reducer, initial); }
-    public static SingleValueExpression foldRight(final ValueExpression values, final BinaryOperator<ValueExpression> reducer) { return new FoldRight(values, reducer, null); }
-    public static SingleValueExpression foldRight(final ValueExpression values, final BinaryOperator<ValueExpression> reducer, final SingleValueExpression initial) { return new FoldRight(values, reducer, initial); }
-    public static SingleValueExpression fold(final ValueExpression values, final BinaryOperator<ValueExpression> reducer) { return foldRight(values, reducer); }
-    public static SingleValueExpression fold(final ValueExpression values, final BinaryOperator<ValueExpression> reducer, final SingleValueExpression initial) { return foldRight(values, reducer, initial); }
+    public static SingleValueExpression foldLeft(final ValueExpression values, final BinaryOperator<SingleValueExpression> reducer) { return new FoldLeft(values, reducer, null); }
+    public static SingleValueExpression foldLeft(final ValueExpression values, final BinaryOperator<SingleValueExpression> reducer, final SingleValueExpression initial) { return new FoldLeft(values, reducer, initial); }
+    public static SingleValueExpression foldRight(final ValueExpression values, final BinaryOperator<SingleValueExpression> reducer) { return new FoldRight(values, reducer, null); }
+    public static SingleValueExpression foldRight(final ValueExpression values, final BinaryOperator<SingleValueExpression> reducer, final SingleValueExpression initial) { return new FoldRight(values, reducer, initial); }
+    public static SingleValueExpression fold(final ValueExpression values, final BinaryOperator<SingleValueExpression> reducer) { return foldRight(values, reducer); }
+    public static SingleValueExpression fold(final ValueExpression values, final BinaryOperator<SingleValueExpression> reducer, final SingleValueExpression initial) { return foldRight(values, reducer, initial); }
     public static ValueExpression rev(final ValueExpression values) { return new Reverse(values); }
     public static ValueExpression exp(final ValueExpression base, final SingleValueExpression count) { return new Expand(base, count); }
     public static BinaryValueExpression mapLeft(final BiFunction<ValueExpression, ValueExpression, BinaryValueExpression> func, final ValueExpression left, final ValueExpression rightExpand) { return func.apply(left, exp(rightExpand, count(left))); }
--- a/core/src/main/java/io/parsingdata/metal/Shorthand.java
+++ b/core/src/main/java/io/parsingdata/metal/Shorthand.java
@@ -185,9 +185,9 @@ public final class Shorthand {
     public static SingleValueExpression con(final byte[] value, final Encoding encoding) { return con(ConstantFactory.createFromBytes(value, encoding)); }
     public static ValueExpression len(final ValueExpression operand) { return new Len(operand); }
     public static NameRef ref(final String name) { return ref(name, null); }
-    public static NameRef ref(final String name, final ValueExpression limit) { return new NameRef(name, limit); }
+    public static NameRef ref(final String name, final SingleValueExpression limit) { return new NameRef(name, limit); }
     public static DefinitionRef ref(final Token definition) { return ref(definition, null); }
-    public static DefinitionRef ref(final Token definition, final ValueExpression limit) { return new DefinitionRef(definition, limit); }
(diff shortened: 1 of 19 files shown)


core/src/main/java/io/parsingdata/metal/Shorthand.java
```diff
@@ -184,28 +195,33 @@ private Shorthand() {}
     public static SingleValueExpression con(final byte[] value) { return con(value, DEFAULT_ENCODING); }
     public static SingleValueExpression con(final byte[] value, final Encoding encoding) { return con(ConstantFactory.createFromBytes(value, encoding)); }
     public static ValueExpression len(final ValueExpression operand) { return new Len(operand); }
+    public static SingleValueExpression len(final SingleValueExpression operand) { return last(new Len(operand)); }
     public static NameRef ref(final String name) { return ref(name, null); }
-    public static NameRef ref(final String name, final ValueExpression limit) { return new NameRef(name, limit); }
+    public static NameRef ref(final String name, final SingleValueExpression limit) { return new NameRef(name, limit); }
```
rdvdijk, line 200: Nice catch :+1:

core/src/main/java/io/parsingdata/metal/Shorthand.java
```diff
@@ -184,28 +195,33 @@ private Shorthand() {}
...
     public static DefinitionRef ref(final Token definition) { return ref(definition, null); }
-    public static DefinitionRef ref(final Token definition, final ValueExpression limit) { return new DefinitionRef(definition, limit); }
+    public static DefinitionRef ref(final Token definition, final SingleValueExpression limit) { return new DefinitionRef(definition, limit); }
```
rdvdijk, line 202: :+1:

core/src/main/java/io/parsingdata/metal/Shorthand.java
```diff
@@ -184,28 +195,33 @@ private Shorthand() {}
...
+    public static SingleValueExpression foldLeft(final ValueExpression values, final BinaryOperator<SingleValueExpression> reducer) { return new FoldLeft(values, reducer, null); }
+    public static SingleValueExpression foldLeft(final ValueExpression values, final BinaryOperator<SingleValueExpression> reducer, final SingleValueExpression initial) { return new FoldLeft(values, reducer, initial); }
+    public static SingleValueExpression foldRight(final ValueExpression values, final BinaryOperator<SingleValueExpression> reducer) { return new FoldRight(values, reducer, null); }
+    public static SingleValueExpression foldRight(final ValueExpression values, final BinaryOperator<SingleValueExpression> reducer, final SingleValueExpression initial) { return new FoldRight(values, reducer, initial); }
+    public static SingleValueExpression fold(final ValueExpression values, final BinaryOperator<SingleValueExpression> reducer) { return foldRight(values, reducer); }
+    public static SingleValueExpression fold(final ValueExpression values, final BinaryOperator<SingleValueExpression> reducer, final SingleValueExpression initial) { return foldRight(values, reducer, initial); }
```
rdvdijk, line 224: As Borat would say: Very nice!

core/src/test/java/io/parsingdata/metal/ShorthandOverloadsTest.java
```diff
@@ -0,0 +1,92 @@
...
+public class ShorthandOverloadsTest {
+
+    public static final ParseState PARSE_STATE = ParseState.createFromByteStream(new ByteStream() {
+        @Override public byte[] read(BigInteger offset, int length) { return new byte[0]; }
+        @Override public boolean isAvailable(BigInteger offset, BigInteger length) { return false; }
+    });
+
+    @Parameterized.Parameter public String description;
+    @Parameterized.Parameter(1) public SingleValueExpression toExecute;
+    @Parameterized.Parameter(2) public SingleValueExpression expectedResult;
```
rdvdijk, line 63: I tend to import these for improved readability.

```java
    @Parameter public String description;
    @Parameter(1) public SingleValueExpression toExecute;
    @Parameter(2) public SingleValueExpression expectedResult;

    @Parameters(name="SingleValueExpression {0}")
```

rdvdijk: And even though it's the default, in this case `@Parameter(0)` looks better than omitting the `(0)`.

rdvdijk (commented):
Looks good to me!

I was wondering: There must be `Token`s in tests where the `last()` was introduced recently, and can now be removed again in favor of these new overloads, right? I wonder if we should clean up some of those,

rdvdijk (approved):

core/src/main/java/io/parsingdata/metal/expression/value/reference/Ref.java
```diff
@@ -49,37 +50,32 @@
...
-        final ImmutableList<Value> evaluatedLimit = limit.eval(parseState, encoding);
-        if (evaluatedLimit.size != 1) {
-            throw new IllegalArgumentException("Limit must evaluate to a single non-empty value.");
-        }
-        if (evaluatedLimit.head.equals(NOT_A_VALUE)) {
-            return ImmutableList.create(NOT_A_VALUE);
-        }
-        return evalImpl(parseState, evaluatedLimit.head.asNumeric().intValueExact());
+        return limit.evalSingle(parseState, encoding)
+            .map(limitValue -> limitValue.equals(NOT_A_VALUE) ? ImmutableList.create(NOT_A_VALUE) : evalImpl(parseState, limitValue.asNumeric().intValueExact()))
```
mvanaken, line 77: The error message in combination with this logic confuses me. In this case, `limitValue` could be `NOT_A_VALUE`, but the error message states that the limit must evaluate to a non-empty value. But we guard that case here?
So, either the error message is incorrect and we can rewrite this as:
```
limit.evalSingle(parseState, encoding)
    .filter(not(limitValue -> limitValue.equals(NOT_A_VALUE)))
    .map(limitValue -> evalImpl(parseState, limitValue.asNumeric().intValueExact()))
    .orElseThrow(() -> new IllegalArgumentException("Ref must evaluate to a non-empty value."))
```
or we can use `orElseGet` instead:
```
limit.evalSingle(parseState, encoding)
    .filter(not(limitValue -> limitValue.equals(NOT_A_VALUE)))
    .map(limitValue -> evalImpl(parseState, limitValue.asNumeric().intValueExact()))
    .orElseGet(() -> ImmutableList.create(NOT_A_VALUE))
```
You may even consider adding an extra `map` step to first convert `limitValue` to int value, before passing it to `evalImpl`, because that may improve readability:
```
.map(limitValue -> limitValue.asNumeric().intValueExact())
```
This latter is just a thought. :)

mvanaken (approved):
There is a small question I have about the behaviour within `Ref`, but the rest looks good!

core/src/main/java/io/parsingdata/metal/expression/value/reference/Ref.java
jvdb: The problem is that there are three cases that must be handled based on what `limit` evaluates to:
1. An empty `Optional`, in which case an exception is thrown.
2. `NOT_A_VALUE`, in which case we return a list with that in it.
3. Something else, in which case we pass that to `evalImpl(...)` and return the result.
What you propose always handles only two of these cases I think, or do you propose to do something else next to this?

jvdb: After discussion, we postpone cleaning this until later.

PR #336, opened 2021-02-15
Add basic JavaDoc to Shorthand

### LINKED ISSUE #334: Add minimal JavaDoc to Shorthand to improve IDE experience
Possibly just an {@see ...} to each helper method to refer to the class it functions as a factory method for.

jvdb:
Resolves #334.

Some explanation:
1. For the helper methods that do not have a backing implementation and are solely compositions of existing parts, I've added a full explanation of what they do. These are: `nod`, `opt`, `when` and `mapLeft`/`mapRight`.
2. For everything else, I have added a reference to the class that implements its functionality.
3. Two exceptions: `def` and `con` as examples of what (in my opinion) a more extensive but still manageable documentation could look like. I've chosen those two because they are probably the most used `Token` and `ValueExpression` helper methods respectively.

### DIFF
--- a/core/src/main/java/io/parsingdata/metal/Shorthand.java
+++ b/core/src/main/java/io/parsingdata/metal/Shorthand.java
@@ -100,29 +100,29 @@ public final class Shorthand {
 
     private Shorthand() {}
 
-    /** Instantiates a {@link Def} nested in a {@link Post}. */
-    public static Token def(final String name, final SingleValueExpression size, final Expression predicate, final Encoding encoding) { return post(def(name, size, encoding), predicate); }
-
-    /** Instantiates a {@link Def} with {@code encoding = null} nested in a {@link Post}. */
-    public static Token def(final String name, final SingleValueExpression size, final Expression predicate) { return def(name, size, predicate, null); }
-
-    /** Instantiates a {@link Def}. */
+    /** "DEFinition": Instantiates a {@link Def}. */
     public static Token def(final String name, final SingleValueExpression size, final Encoding encoding) { return new Def(name, size, encoding); }
 
-    /** Instantiates a {@link Def} with {@code encoding = null}. */
-    public static Token def(final String name, final SingleValueExpression size) { return def(name, size, (Encoding)null); }
-
-    /** Instantiates a {@link Def} with {@code size = con(size)}, nested in a {@link Post}. */
-    public static Token def(final String name, final long size, final Expression predicate, final Encoding encoding) { return def(name, con(size), predicate, encoding); }
-
-    /** Instantiates a {@link Def} with {@code size = con(size)} and {@code encoding = null}, nested in a {@link Post}. */
-    public static Token def(final String name, final long size, final Expression predicate) { return def(name, size, predicate, null); }
-
-    /** Instantiates a {@link Def} with {@code size = con(size)}. */
+    /** "DEFinition": Instantiates a {@link Def} with {@code size = con(size)}. */
     public static Token def(final String name, final long size, final Encoding encoding) { return def(name, con(size), encoding); }
 
-    /** Instantiates a {@link Def} with {@code size = con(size)} and {@code encoding = null}. */
+    /** "DEFinition": Instantiates a {@link Def} with {@code encoding = null}. */
+    public static Token def(final String name, final SingleValueExpression size) { return def(name, size, (Encoding)null); }
+
+    /** "DEFinition": Instantiates a {@link Def} with {@code size = con(size)} and {@code encoding = null}. */
     public static Token def(final String name, final long size) { return def(name, size, (Encoding)null); }
+
+    /** "DEFinition": Instantiates a {@link Def} nested in a {@link Post}. */
+    public static Token def(final String name, final SingleValueExpression size, final Expression predicate, final Encoding encoding) { return post(def(name, size, encoding), predicate); }
+
+    /** "DEFinition": Instantiates a {@link Def} with {@code size = con(size)}, nested in a {@link Post}. */
+    public static Token def(final String name, final long size, final Expression predicate, final Encoding encoding) { return def(name, con(size), predicate, encoding); }
+
+    /** "DEFinition": Instantiates a {@link Def} with {@code encoding = null} nested in a {@link Post}. */
+    public static Token def(final String name, final SingleValueExpression size, final Expression predicate) { return def(name, size, predicate, null); }
+
+    /** "DEFinition": Instantiates a {@link Def} with {@code size = con(size)} and {@code encoding = null}, nested in a {@link Post}. */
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/parsingdata/metal/Shorthand.java
```diff
@@ -99,154 +100,215 @@
 
     private Shorthand() {}
 
+    /** Instantiates a {@link Def} nested in a {@link Post}. */
     public static Token def(final String name, final SingleValueExpression size, final Expression predicate, final Encoding encoding) { return post(def(name, size, encoding), predicate); }
+
+    /** Instantiates a {@link Def} with {@code encoding = null} nested in a {@link Post}. */
```
mvanaken, line 106: I would not mention the post part here, but simply reference that shorthand instead. I also thought that you did that... Am I reviewing too soon?

jvdb: Actually, I thought that referencing those helpers methods would create a kind of click-through jungle that would not really help users of Metal to understand what they need to know. In this case the JavaDoc says that what you end up with is a `Def` and a `Post`, with links to the explanations of what both do. And in cases where relevant arguments are filled in with a default, that is also mentioned here.

rdvdijk: I wonder if the Javadoc tags introduced in Java 8 would help here: `@implSpec` and `@implNote`.

```java
/**
 * Instantiates a {@link Def} that succeeds when the predicate succeeds.
 *
 * @imlpSpec Instantiates a {@link Def} with {@code encoding = null} nested in a {@link Post}.
 */
```

core/src/main/java/io/parsingdata/metal/Shorthand.java
```diff
@@ -99,154 +100,215 @@
...
     public static SingleValueExpression con(final long value, final Encoding encoding) { return con(ConstantFactory.createFromNumeric(value, encoding)); }
+
+    /** "CONstant": instantiates {@link Const} with a {@link Value} object using {@code value} and {@code DEFAULT_ENCODING}. */
     public static SingleValueExpression con(final String value) { return con(value, DEFAULT_ENCODING); }
+
+    /** "CONstant": instantiates {@link Const} with a {@link Value} object using {@code value} the provided {@code encoding}. */
     public static SingleValueExpression con(final String value, final Encoding encoding) { return con(ConstantFactory.createFromString(value, encoding)); }
+
+    /** "CONstant": instantiates {@link Const} with {@code value}. */
     public static SingleValueExpression con(final Value value) { return new Const(value); }
```
rdvdijk, line 240: Minor: maybe this should be the first `con()` in the list of shorthands. With most (or all) other shorthands here, the 'first in each list' references the class, where all the other eventually delegate to it.

core/src/main/java/io/parsingdata/metal/Shorthand.java
```diff
@@ -99,154 +100,215 @@
 
     private Shorthand() {}
 
+    /** Instantiates a {@link Def} nested in a {@link Post}. */
```
rdvdijk, line 103: To keep this similar with `nod`,`opt` and `con`: Maybe start with:

```
/** "DEFinition" : ...
```

core/src/main/java/io/parsingdata/metal/Shorthand.java
```diff
@@ -99,154 +100,215 @@
...
+    /** @see Seq */ public static Token seq(final Encoding encoding, final Token token1, final Token token2, final Token... tokens) { return seq(NO_NAME, encoding, token1, token2, tokens); }
+    /** @see Seq */ public static Token seq(final Token token1, final Token token2, final Token... tokens) { return seq((Encoding)null, token1, token2, tokens); }
+    /** @see io.parsingdata.metal.token.Sub */ public static Token sub(final String name, final Token token, final ValueExpression address, final Encoding encoding) { return new io.parsingdata.metal.token.Sub(name, token, address, encoding); }
+    /** @see io.parsingdata.metal.token.Sub */ public static Token sub(final String name, final Token token, final ValueExpression address) { return sub(name, token, address, null); }
+    /** @see io.parsingdata.metal.token.Sub */ public static Token sub(final Token token, final ValueExpression address, final Encoding encoding) { return sub(NO_NAME, token, address, encoding); }
+    /** @see io.parsingdata.metal.token.Sub */ public static Token sub(final Token token, final ValueExpression address) { return sub(token, address, null); }
```
rdvdijk, line 152: Oh, the need for a fully-qualified class name here... :rofl:

core/src/main/java/io/parsingdata/metal/Shorthand.java
```diff
@@ -99,154 +100,215 @@
...
+    /** @see Reverse */ public static ValueExpression rev(final ValueExpression values) { return new Reverse(values); }
+    /** @see Expand */ public static ValueExpression exp(final ValueExpression base, final SingleValueExpression count) { return new Expand(base, count); }
+
+    /** Denotes a map operation using the provided {@code func}, applied once for each result of evaluating {@code left} and reusing the single result of evaluating {@code rightExpand}. */
+    public static BinaryValueExpression mapLeft(final BiFunction<ValueExpression, ValueExpression, BinaryValueExpression> func, final ValueExpression left, final SingleValueExpression rightExpand) { return func.apply(left, exp(rightExpand, count(left))); }
+
+    /** Denotes a map operation using the provided {@code func}, applied once for each result of evaluating {@code right} and reusing the single result of evaluating {@code leftExpand}. */
+    public static BinaryValueExpression mapRight(final BiFunction<ValueExpression, ValueExpression, BinaryValueExpression> func, final SingleValueExpression leftExpand, final ValueExpression right) { return func.apply(exp(leftExpand, count(right)), right); }
```
rdvdijk, line 289: (Mental note: As you explained on Webex, this is a subtle API change: `rightExpand` and `leftExpand` are now changed to `SingleValueExpression`s.)

core/src/main/java/io/parsingdata/metal/Shorthand.java
jvdb: Sounds like a good approach for a future extension!

core/src/main/java/io/parsingdata/metal/Shorthand.java
jvdb: I've added them everywhere for consistency, even when the shorthand is not really an abbreviation.

core/src/main/java/io/parsingdata/metal/Shorthand.java
jvdb: Agreed, but to do this everywhere required lots of stuff to be reordered. But good idea!

rdvdijk (approved):
Looks good to me!

PR #345, opened 2022-11-29
Added size() to Source and size() token to obtain the size of the data

gertjanal:
ValueExpression to easily read a footer

```
sub(
    def("footer", con(1), eqNum(con(0x01))),
    sub(size(), con(1)))
```

We could consider renaming it to `eof()` or `end()` to keep the method length 3 characters. We could also consider returning the last byte position instead of the size (so `size - 1`), but on the other hand, keeping `size()` might come in handy when we want to paginate data blocks (size == 64 might be better than lastByte == 63)

### DIFF
--- a/core/src/main/java/io/parsingdata/metal/Shorthand.java
+++ b/core/src/main/java/io/parsingdata/metal/Shorthand.java
@@ -77,6 +77,7 @@ import io.parsingdata.metal.expression.value.reference.Ref;
 import io.parsingdata.metal.expression.value.reference.Ref.DefinitionRef;
 import io.parsingdata.metal.expression.value.reference.Ref.NameRef;
 import io.parsingdata.metal.expression.value.reference.Self;
+import io.parsingdata.metal.expression.value.reference.Size;
 import io.parsingdata.metal.token.Cho;
 import io.parsingdata.metal.token.Def;
 import io.parsingdata.metal.token.Post;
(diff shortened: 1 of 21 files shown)


jvdb (commented):
I think there are several problems with adding this `ValueExpression` to Metal. I will enumerate them:

1. The reason `ByteStream` does not already have a `size()`-method is because we want a broad range of underlying implementations to be candidates for encapsulation in a `ByteStream` and use in Metal. Some stream-implementations necessarily have no knowledge of their size (e.g., network streams, legacy storage devices). So by requiring the `ByteStream`-interface to implement this it would remove compatibility with those streams.
2. If you want to do something like this I would suggest to accumulate all the data you want to parse starting at the back with something like `Until(...)`, then references this data with a `ValueExpression` such as `Ref(...)` from which you can obtain the size using the `Len(...)` `ValueExpression`. In order to parse this data again with Metal, you can insert the referenced data into a `Tie(...)`.
3. From a forensic perspective, I can't really see the use case where you want to parse a fixed piece of data from the end. If you are parsing trusted data and know 100% that there is not some random piece of data attached at the end (which happens a lot) it's not a forensic scenario and maybe Metal should not be used for this. If you are in a real forensic scenario then assuming there is some data at the end is risky and I would consider it a bad practice.

Finally, I fear this may also have some bad interactions with attempts to be lazy with regard to when data is read. But I have not researched this fully and the 3 points above I think are clear in why this is not a good addition to Metal.

PR #361, opened 2023-02-23
#360: Add variant to Until to allow parsing excluding terminator

### LINKED ISSUE #360: Extend Until to exclude terminator from ParseGraph
There are three scenarios where you would like to use Until:

1. when there is a separate terminator between structure A and B (A|T|B)
2. when you know how the structure A ends (AT|B)
3. when you know how the next structure B starts (A|TB)

The current implementation of Until parses two structures: a def and the terminator.
The current implementation can handle 1. correctly, but for 2 and 3 requires some hacky solutions with some drawbacks. See examples solutions below.

1. until("A", T)
2. until("A", post(EMPTY, endsWith(T)))
3. until("A", sub(T, CURRENT_OFFSET))

Especially for scenario 3, the "T" part is parsed twice and is also available twice in the parseGraph.
This can be resolved by allowing two implementations of Until where you can specify if the terminator should be part of the parseGraph.

mvanaken:
There are three scenarios where you would like to use Until:

1. when there is a separate terminator between structure A and B (A|T|B)
2. when you know how structure A ends (AT|B)
3. when you know how structure B starts (A|TB)

The current implementation of Until parses two structures: a def and the terminator.
The current implementation can handle 1. correctly, but for 2 and 3 requires some hacky solutions with some drawbacks. See examples solutions below.

1. until("A", T)
2. until("A", post(EMPTY, endsWith(T)))
3. until("A", sub(T, CURRENT_OFFSET))

Especially for scenario 3, the "T" part is parsed twice and is also available twice in the parseGraph.
This can be resolved by allowing two implementations of Until where you can specify if the terminator should be part of the parseGraph.

This PR includes the new shorthands "defU" (name still in progress) that parses the def, but not the terminator. It uses the same Until class, but with an additional boolean indicating if the terminator should be included or not.

Resolves #360.

### DIFF
--- a/core/src/main/java/io/parsingdata/metal/Shorthand.java
+++ b/core/src/main/java/io/parsingdata/metal/Shorthand.java
@@ -215,7 +215,7 @@ public final class Shorthand {
 
 
     /** "DEFinition": Instantiates a {@link Until} and its terminator nested in a {@link Seq}. */
-    public static Token until(final String name, final ValueExpression initialSize, final ValueExpression stepSize, final ValueExpression maxSize, final Token terminator, final Encoding encoding) { return seq(name, def(name, initialSize, stepSize, maxSize, terminator, encoding), terminator); }
+    public static Token until(final String name, final ValueExpression initialSize, final ValueExpression stepSize, final ValueExpression maxSize, final Token terminator, final Encoding encoding) { return seq(def(name, initialSize, stepSize, maxSize, terminator, encoding), terminator); }
 
     /** "DEFinition": Instantiates a {@link Until} and its terminator nested in a {@link Seq} with {@code encoding = null}. */
     public static Token until(final String name, final ValueExpression initialSize, final ValueExpression stepSize, final ValueExpression maxSize, final Token terminator) { return until(name, initialSize, stepSize, maxSize, terminator, null); }
--- a/core/src/main/java/io/parsingdata/metal/Shorthand.java
+++ b/core/src/main/java/io/parsingdata/metal/Shorthand.java
@@ -180,7 +180,7 @@ public final class Shorthand {
     /** @see Tie */ public static Token tie(final String name, final Token token, final ValueExpression dataExpression) { return tie(name, token, dataExpression, null); }
     /** @see Tie */ public static Token tie(final Token token, final ValueExpression dataExpression, final Encoding encoding) { return tie(NO_NAME, token, dataExpression, encoding); }
     /** @see Tie */ public static Token tie(final Token token, final ValueExpression dataExpression) { return tie(token, dataExpression, null); }
-    /** @see Until */ public static Token until(final String name, final ValueExpression initialSize, final ValueExpression stepSize, final ValueExpression maxSize, final Token terminator, final Encoding encoding) { return new Until(name, initialSize, stepSize, maxSize, terminator, encoding); }
+    /** @see Until */ public static Token until(final String name, final ValueExpression initialSize, final ValueExpression stepSize, final ValueExpression maxSize, final Token terminator, final Encoding encoding) { return new Until(name, initialSize, stepSize, maxSize, terminator, true, encoding); }
     /** @see Until */ public static Token until(final String name, final ValueExpression initialSize, final ValueExpression stepSize, final ValueExpression maxSize, final Token terminator) { return until(name, initialSize, stepSize, maxSize, terminator, null); }
     /** @see Until */ public static Token until(final String name, final ValueExpression initialSize, final ValueExpression stepSize, final Token terminator, final Encoding encoding) { return until(name, initialSize, stepSize, null, terminator, encoding); }
     /** @see Until */ public static Token until(final String name, final ValueExpression initialSize, final ValueExpression stepSize, final Token terminator) { return until(name, initialSize, stepSize, null, terminator, null); }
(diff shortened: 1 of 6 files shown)


core/src/main/java/io/parsingdata/metal/Shorthand.java
```diff
@@ -189,6 +189,15 @@ private Shorthand() {}
     /** @see Until */ public static Token until(final String name, final Token terminator, final Encoding encoding) { return until(name, null, terminator, encoding); }
     /** @see Until */ public static Token until(final String name, final Token terminator) { return until(name, terminator, null); }
 
+    public static Token defU(final String name, final ValueExpression initialSize, final ValueExpression stepSize, final ValueExpression maxSize, final Token terminator, final Encoding encoding) { return new Until(name, initialSize, stepSize, maxSize, terminator, false, encoding); }
```
mvanaken, line 192: We do not like the name `defU`, U meaning Unterminated, anymore. Mostly, because the def is not unterminated, it is terminated, but we just do not know where yet. We are considering the following names:
* tdefEx: terminated def excluding terminator. this allows for the equivalent `tdefIn` for which we want to rename the original until shorthand to (and keeping the original name for backward compatibility)
* defn: in same line of repn, although this can be confusing, since repn uses a valueexpression to determine how far it should go, while rep is dynamic. so then def should actually be defn and this new token is def... let's not go there.

core/src/main/java/io/parsingdata/metal/token/Until.java
```diff
@@ -114,12 +116,17 @@ private Trampoline<Optional<ParseState>> iterate(final Environment environment,
...
             .orElseGet(Util::failure)
-            .map(parsedParseState -> complete(() -> success(parsedParseState)))
+            .map(parseState -> complete(() -> success(parseState)))
             .orElseGet(() -> intermediate(() -> iterate(environment, currentSize.add(stepSize), stepSize, maxSize)));
     }
 
+    private Optional<ParseState> parseTerminator(final Environment environment, final ParseState parseStateExcludingTerminator) {
+        return terminator.parse(environment.withParseState(parseStateExcludingTerminator))
+            .map(parseStateIncludingTerminator -> includeTerminator ? parseStateIncludingTerminator : parseStateExcludingTerminator);
+    }
```
mvanaken, line 128: We decide to move the logic to a separate method, so we can give the different parseStates meaningful names and without cluttering the method above. But if that is not the case, then we can inline this method ofcourse.

core/src/test/java/io/parsingdata/metal/token/UntilTest.java
```diff
@@ -113,6 +118,21 @@ public void untilExclusiveWithTerminatorInParseGraph() {
         assertEquals(3, newLines.size);
     }
 
+    @Test
+    public void defUnterminated() {
```
mvanaken, line 122: We should rename this method after a good name is determined.

core/src/main/java/io/parsingdata/metal/Shorthand.java
rdvdijk: > ["There are only two hard things in Computer Science: cache invalidation and naming things"](https://skeptics.stackexchange.com/questions/19836/has-phil-karlton-ever-said-there-are-only-two-hard-things-in-computer-science)

I like your first suggestion: `tdef`.

As a side note: Kaitai has also encountered this exact case, and they have [the special keyword `include`](https://doc.kaitai.io/user_guide.html#delimited-struct-advanced) that indicates whether a configured `terminator` should be part of the parsed value or not.

jvdvegt: I like the idea of appending `Ex` or `In`, that makes it really clear what the method does. So my vote goes to `tdexIn` and `tdefEx`.

jvdb: defu

jvdvegt:
I'm reviewing by email, but is that really a single line? It could sure use some newlines!

jvdb ***@***.***> schreef op 25 februari 2023 12:07:09 CET:

core/src/main/java/io/parsingdata/metal/Shorthand.java
mvanaken: The confusing part of the current `until` implementation is that it is actually a combination token that parses two defs. The proposed alternative implementation, `defU`, only parses a single `def`, which is why we would like to let the shorthand of this variant also start with `def`.

In addition, we are able to write `until` behaviour using `defU` icw `seq` and `post`, so we could decide to let the `Until` class always set `includeTerminator` to false and if a terminator is desired to be included, we can resolve that with shorthands only.

I would therefor propose to rename `Until.java` to `DefUntil.java` (or `DefBoundedByToken.java`) as is will be implemented to parse a single def until what ever token was defined and then define the following shorthands from the user's perspective:
* def(name, token): (an overload method of the current def) the specified terminator should not be part of the resulting def and should not be part of the parsegraph. the token simply defines the boundary of the def.
* defIn(name, token): the specified terminator should be part of the resulting def, e.g. line + \n
* defs(name, token): a shorthand for seq(def(name, token), token), such that the terminator is also parsed and part of the parsegraph.

I think we then cover all the possible until variants.

core/src/test/java/io/parsingdata/metal/token/UntilTest.java
```diff
@@ -58,7 +60,7 @@ public class UntilTest {
 
     @Test
     public void threeNewLines() {
-        final Optional<ParseState> parseState = createToken(con(0), NEWLINE).parse(env(stream(INPUT, US_ASCII)));
+        final Optional<ParseState> parseState = rep(until("line", NEWLINE)).parse(env(stream(INPUT, US_ASCII)));
```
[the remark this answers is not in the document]
mvanaken: Converted it to a parameterized test instead. ;)

core/src/test/java/io/parsingdata/metal/token/UntilTest.java
mvanaken: No need anymore, converted it to a parameterized test instead.

core/src/main/java/io/parsingdata/metal/Shorthand.java
jvdb: I think calling them all `def` is perhaps a good idea after all. Essentially it retains this idea that there are basically two types of `Token`s: combinators and value definitions. The latter is then always called `Def`, regardless of how it is defined -- as either something with a provided size, a provided terminator or through some provided `Expression` that does whatever it needs to determine whether a growing candidate value is "complete".

core/src/main/java/io/parsingdata/metal/Shorthand.java
```diff
@@ -124,6 +124,38 @@ private Shorthand() {}
...
+    public static Token def(final String name, final ValueExpression initialSize, final Token terminator) { return def(name, initialSize, null, terminator, null); }
+
+    /** "DEFinition": Instantiates a {@link Until} where the size of the def is dynamically determined with {@code initialSize = null}, {@code stepSize = null} and {@code maxSize = null}. */
+    public static Token def(final String name, final Token terminator, final Encoding encoding) { return def(name, null, terminator, encoding); }
+
+    /** "DEFinition": Instantiates a {@link Until} where the size of the def is dynamically determined with {@code initialSize = null}, {@code stepSize = null}, {@code maxSize = null} and {@code encoding = null}. */
+    public static Token def(final String name, final Token terminator) { return def(name, terminator, null); }
+
+    /** "DEFinition": Instantiates a {@link Until} with the expression nested in a {@link Post} and {@code initialSize = con(1)}. */
+    public static Token def(final String name, final Expression predicate, final Encoding encoding) { return def(name, con(1), post(EMPTY, predicate), encoding); }
```
jvdb, line 153: Maybe we need more overloads for this pattern? Now the `Expression`-variant of the "growing" `def` only supports all default values for `initialSize`, `stepSize` and `maxSize`. I don't mind if we add those later, but it's a choice to make.

core/src/main/java/io/parsingdata/metal/Shorthand.java
rdvdijk: Agreed. For the longest time we only had `def` as our value definition in Metal. Until `until` came along (pun intended). Ever since then we've had to explain this whenever we've shown Metal to new team members.

jvdb (approved):

mvanaken:
Run code coverage locally and this was again 100%. The codecov hickup is resolved in #362.

PR #363, opened 2023-02-24
#362: Build with GitHub actions

### LINKED ISSUE #362: Move CI to GitHub Actions
We're currently using Travis (for Linux) and AppVeyor (for Windows) as CI-platforms. Travis is no longer free, so we need to move away from that. We can consolidate Linux and Windows builds on GitHub Actions.

jvdb:
Resolves #362.

The issue proposes to replace both Travis and AppVeyor with GitHub Actions but I suppose we can keep AppVeyor while it works well and is free.


rdvdijk:
Looks good.

This PR can point to the `master` branch, I think?

mvanaken:
> Looks good.
> 
> This PR can point to the `master` branch, I think?

Unfortunately, no. This issue is branched of until-revised. 
However, there is a problem, because we cannot merge until-revised without these changes, because the code coverage check is broken there, so my proposal is this:
* we merge until-revised into this branch
* if this branch is green, then 
  * first merge until-revised branch
  * then merge this branch

jvdb:
It depends a bit on whether we think the `until-revised`-PR will remain open for much longer. If we're close to merging that too, we might as well opt for 1.

.github/workflows/build.yml
```diff
@@ -0,0 +1,51 @@
...
+          java-version: ${{ matrix.java-version }}
+          distribution: ${{ matrix.distribution }}
+
+      - name: Build with Maven and analyze with Sonar
+        # Analyze only the Temurin 11 build
+        if: matrix.distribution == 'temurin' && matrix.java-version == '11'
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
+        run: mvn -B --update-snapshots -Dmaven.test.failure.ignore=true package org.sonarsource.scanner.maven:sonar-maven-plugin:sonar -Dsonar.projectKey=parsingdata_metal
```
mvanaken, line 42: As discussed offline: maybe add a comment to explain that `parsingdata_metal` projectKey is generated by the SonarCloud GitHub App during install of the app? This to avoid that this key is renamed. The value is not found anywhere else, so one may assume that this key is defined here and that it can be adjusted.

mvanaken (approved):

PR #368, opened 2023-03-01
#367 Use scope of environment for ParseValue in Until instead of the direct name.

### LINKED ISSUE #367: Bug in Until: the name does not contain the scope.
The resulting def of an Until will recieve the exact name without the names of the parent tokens.

E.g.: 
```
seq("struct", 
    until("value", terminator), 
    def("afterValue", con(1))
);
```
The name of `until` is _value_, while it should be _struct.value_. The `def` gets the correct name _struct.afterValue_.

The problem now is that you are unable to select these values specifically if you are only interested in a value within in named structure.

The fix is simple: the ParseValue in until should use `environment.scope` instead of `name`

mvanaken:
Due to this bug you are unable to select a specific value from the `ParseGraph` in an other structure.

Included a test, which failed without the fix.

Resolves #367

### DIFF
--- a/core/src/test/java/io/parsingdata/metal/token/UntilTest.java
+++ b/core/src/test/java/io/parsingdata/metal/token/UntilTest.java
@@ -17,6 +17,7 @@
 package io.parsingdata.metal.token;
 
 import static io.parsingdata.metal.Shorthand.CURRENT_OFFSET;
+import static io.parsingdata.metal.Shorthand.seq;
 import static io.parsingdata.metal.Shorthand.sub;
 import static java.nio.charset.StandardCharsets.US_ASCII;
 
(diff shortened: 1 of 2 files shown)


jvdb (approved):
Kind of mindblowing that the original implementation of `Until` just completely threw away the scope's name an inserted its own short name instead.

The tests clearly illustrate that the current workings on `Until` are not really in line with the general ideas we normally espouse around how combinators and values should interact in Metal. You can see this in `UntilTest`: the name "struct.value.terminator" doesn't make sense (the terminator is not part of the value!) but it's a result of the fact that `Until` acts as both a combinator (putting its name in the enclosing scope) and a value definition (putting its name on the parsed value).

core/src/test/java/io/parsingdata/metal/token/UntilTest.java
```diff
@@ -127,4 +128,16 @@ private Token createToken(final ValueExpression initialSize, final Token termina
         return repn(until("line", initialSize, terminator), con(3));
     }
 
+    @Test
+    public void nameScope() {
```
rdvdijk, line 132: Woah 😲

rdvdijk (approved):
Great find!

mvanaken:
Ran code coverage Jacoco locally, and everything was 100%. This codecov hickup will be resolved in #362.

Issue #383, opened 2023-05-22
Join is missing equality methods.


mvanaken:
In #375 , Join is introduced, but it did not overwrite the methods "toString" "equals" and "hashCode". This could introduce bugs if not implemented. It was also missing from the AutoEqualityTest.

I could be interesting if we could find the classes to test automatically and fail if the automatically found classes does not match the expected classes. Then AutoEqualityTest would have failed the moment Join was added.

It turns out that other values are also missing from this test, e.g. NotAValue.

PR #388, opened 2023-07-24
#383 join equality

mvanaken:
There was a bug introduced when the Join class was merged, because the join class did not contain equality methods. 
Since it is easy to forget them, I thought it may be a good idea to help ourselves and let the AutoEqualityTest tell us if we forgot to include a class.

This Pull Request contains:
1. Add missing equality methods for [Join](core/src/main/java/io/parsingdata/metal/expression/value/Join.java) class.
2. Automatically check that we test all classes for equality in [AutoEqualityTest](core/src/test/java/io/parsingdata/metal/AutoEqualityTest.java).
I extended the AutoEqualityClass that collects all the classes from certain packages to force us to add a new class to the CLASSES_TO_TEST or CLASSES_TO_IGNORE constant.

It turned out, `Join.class` was not the only class missing in the AutoEqualityTest. I added:
Join.class, Self.class, Ref.NameRef.class, Ref.DefinitionRef.class, NotAValue.class, ParseGraph.class, ImmutableList.class and Slice.class. See also [this commit](https://github.com/parsingdata/metal/commit/8cb3ff83ef06a6dca924a790e46437f3bf942124).

Let me know what you think.

### DIFF
--- a/core/src/test/java/io/parsingdata/metal/AutoEqualityTest.java
+++ b/core/src/test/java/io/parsingdata/metal/AutoEqualityTest.java
@@ -19,6 +19,8 @@ package io.parsingdata.metal;
 import static java.math.BigInteger.ONE;
 import static java.math.BigInteger.TEN;
 import static java.math.BigInteger.ZERO;
+import static java.util.Spliterators.spliteratorUnknownSize;
+import static java.util.function.Predicate.not;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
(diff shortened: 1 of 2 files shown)


core/src/test/java/io/parsingdata/metal/AutoEqualityTest.java
```diff
@@ -152,6 +208,7 @@ public class AutoEqualityTest {
     private static final List<Supplier<Object>> TOKEN_ARRAYS = Arrays.asList(() -> new Token[] { any("a"), any("b")}, () -> new Token[] { any("b"), any("c") }, () -> new Token[] { any("a"), any("b"), any("c") });
```
rdvdijk, line 208: Maybe it's time to replace `Arrays.asList` with `List.of`, or do we want to keep Metal Java 8 compatible?

rdvdijk (approved):

core/src/test/java/io/parsingdata/metal/AutoEqualityTest.java
mvanaken: Good idea, but maybe not in this issue.

core/src/test/java/io/parsingdata/metal/AutoEqualityTest.java
```diff
@@ -197,40 +255,71 @@ private static Map<Class, List<Supplier<Object>>> buildMap() {
...
+        } catch (IOException e) {
+            throw new UncheckedIOException(e);
+        }
+    }
+
+    private static Stream<? extends Class<?>> getClasses(final String packageName, final URL url) {
+        try (BufferedReader reader = new BufferedReader(new InputStreamReader(url.openStream()))) {
+            return reader.lines()
+                .filter(line -> line.endsWith(".class"))
+                .map(className -> packageName + "." + className.substring(0, className.lastIndexOf('.')))
```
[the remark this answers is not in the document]
mvanaken: Yes! e.g. `NameRef` and `DefinitionRef` are inner classes of `Ref`.

Issue #392, opened 2023-07-27
Allow varargs for Ref to reference to defs with different names and return them in parsing order.


mvanaken:
You can only specify one argument to ref to reference tokens with a specific name.
If you want to select tokens with different names simultanuously, you can use Join, eg:
`join(ref("a"), ref("b"), ref("c"))`

This will have the following properties:
1. this will select the tokens in the order specified, in this example thatt could be "a", "a", "a", "b", "b", "c"
2. this will evaluate the separate expressions, and with multiple refs, this would mean traversing the ParseGraph multiple times, in this example 3 times.

If Ref accepts a varargs for the selected elements, e.g.:
`ref("a", "b", "c")`

This will impact to above properties:
1. this will select the tokens in the parsing order, in this example that could be "a", "b", "c", "b", "a", "a"
2. to find these elements only one traversal of the ParseGraph is required.

PR #393, opened 2023-07-28
#392 RefAny implementation by allowing varargs for the Ref class.

mvanaken:
The main changes are in Ref

### DIFF
--- a/core/src/main/java/io/parsingdata/metal/expression/value/reference/Ref.java
+++ b/core/src/main/java/io/parsingdata/metal/expression/value/reference/Ref.java
@@ -16,15 +16,21 @@
 
 package io.parsingdata.metal.expression.value.reference;
 
+import static io.parsingdata.metal.Shorthand.con;
+import static io.parsingdata.metal.Shorthand.ref;
 import static io.parsingdata.metal.Trampoline.complete;
 import static io.parsingdata.metal.Trampoline.intermediate;
+import static io.parsingdata.metal.Util.checkContainsNoNulls;
 import static io.parsingdata.metal.Util.checkNotNull;
+import static io.parsingdata.metal.data.ImmutableList.create;
 import static io.parsingdata.metal.data.Selection.NO_LIMIT;
 import static io.parsingdata.metal.data.Selection.getAllValues;
 import static io.parsingdata.metal.expression.value.NotAValue.NOT_A_VALUE;
 
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Objects;
-import java.util.function.Predicate;
+import java.util.function.BiPredicate;
 
 import io.parsingdata.metal.Trampoline;
 import io.parsingdata.metal.Util;
@@ -48,24 +54,30 @@ import io.parsingdata.metal.token.Token;
  */
 public class Ref<T> implements ValueExpression {
 
-    public final T reference;
-    public final Predicate<ParseValue> predicate;
+    public final ImmutableList<T> references;
+    public final BiPredicate<ParseValue, T> predicate;
     public final SingleValueExpression limit;
 
-    private Ref(final T reference, final Predicate<ParseValue> predicate, final SingleValueExpression limit) {
-        this.reference = checkNotNull(reference, "reference");
+    @SafeVarargs
+    private Ref(final BiPredicate<ParseValue, T> predicate, final SingleValueExpression limit, final T reference,  final T... references) {
         this.predicate = checkNotNull(predicate, "predicate");
         this.limit = limit;
+        this.references = create(checkContainsNoNulls(references, "references"))
+            .add(checkNotNull(reference, "reference"));
     }
 
     public static class NameRef extends Ref<String> {
-        public NameRef(final String reference) { this(reference, null); }
-        public NameRef(final String reference, final SingleValueExpression limit) { super(reference, value -> value.matches(reference), limit); }
+        public NameRef(final SingleValueExpression limit, final ImmutableList<String> references) { this(limit, references.head, Ref.toList(references.tail).toArray(new String[0])); }
+        public NameRef(final String reference, final String... references) { this(null, reference, references); }
+        public NameRef(final String reference, final SingleValueExpression limit) { this(limit, reference); }
+        public NameRef(final SingleValueExpression limit, final String reference, final String... references) { super(ParseValue::matches, limit, reference, references); }
     }
 
     public static class DefinitionRef extends Ref<Token> {
-        public DefinitionRef(final Token reference) { this(reference, null); }
-        public DefinitionRef(final Token reference, final SingleValueExpression limit) { super(reference, value -> value.definition.equals(reference), limit); }
+        public DefinitionRef(final SingleValueExpression limit, final ImmutableList<Token> references) { this(limit, references.head, Ref.toList(references.tail).toArray(new Token[0])); }
+        public DefinitionRef(final Token reference, final Token... references) { this(null, reference, references); }
+        public DefinitionRef(final Token reference, final SingleValueExpression limit) { this(limit, reference); }
+        public DefinitionRef(final SingleValueExpression limit, final Token reference, final Token... references) { super((value, ref) -> value.definition.equals(ref), limit, reference, references); }
     }
 
     @Override
@@ -79,7 +91,17 @@ public class Ref<T> implements ValueExpression {
     }
 
     private ImmutableList<Value> evalImpl(final ParseState parseState, final int limit) {
-        return wrap(getAllValues(parseState.order, predicate, limit), new ImmutableList<Value>()).computeResult();
+        return wrap(getAllValues(parseState.order, parseValue -> toList(references).stream().anyMatch(ref -> predicate.test(parseValue, ref)), limit), new ImmutableList<Value>()).computeResult();
+    }
+
+    private static <T> List<T> toList(final ImmutableList<T> allValues) {
+        final List<T> flatten = new ArrayList<>();
+        ImmutableList<T> tail = allValues;
+        while (!tail.isEmpty()) {
+            flatten.add(tail.head);
+            tail = tail.tail;
+        }
+        return flatten;
     }
 
     private static <T, U extends T> Trampoline<ImmutableList<T>> wrap(final ImmutableList<U> input, final ImmutableList<T> output) {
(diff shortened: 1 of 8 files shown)


core/src/main/java/io/parsingdata/metal/expression/value/reference/Ref.java
```diff
@@ -48,24 +52,36 @@
...
-        this.reference = checkNotNull(reference, "reference");
+    @SafeVarargs
+    private Ref(final BiPredicate<ParseValue, T> predicate, final SingleValueExpression limit, final T reference,  final T... references) {
         this.predicate = checkNotNull(predicate, "predicate");
         this.limit = limit;
+        this.references = create(checkContainsNoNulls(references, "references"))
+            .add(checkNotNull(reference, "reference"));
+    }
+
+    private Ref(final BiPredicate<ParseValue, T> predicate, final SingleValueExpression limit, final ImmutableList<T> references) {
```
mvanaken, line 67: This constructor is meant as a copy constructor to avoid converting an immutable list to an array. In combination with the `withLimit` method below, this constructor can be kept private. The `withLimit` is a copy method used in the Shorthand when ref is used in combination with last.

core/src/main/java/io/parsingdata/metal/expression/value/reference/Ref.java
```diff
@@ -79,7 +95,17 @@ public ImmutableList<Value> eval(final ParseState parseState, final Encoding enc
     }
 
     private ImmutableList<Value> evalImpl(final ParseState parseState, final int limit) {
-        return wrap(getAllValues(parseState.order, predicate, limit), new ImmutableList<Value>()).computeResult();
+        return wrap(getAllValues(parseState.order, parseValue -> toList(references).stream().anyMatch(ref -> predicate.test(parseValue, ref)), limit), new ImmutableList<Value>()).computeResult();
+    }
+
+    static <T> List<T> toList(final ImmutableList<T> allValues) {
```
mvanaken, line 101: Made this package private so I can write a test for it, see RefToListTest.

core/src/test/java/io/parsingdata/metal/AutoEqualityTest.java
```diff
@@ -319,6 +321,9 @@ private static Object[] generateObjectArrays(final Class<?> c) throws IllegalAcc
         cons.setAccessible(true);
         List<List<Supplier<Object>>> args = new ArrayList<>();
         for (Class<?> cl : cons.getParameterTypes()) {
+            if (!mapping.containsKey(cl)) {
+                throw new AssertionError("Please add a mapping for type " + cl.getSimpleName());
+            }
```
mvanaken, line 326: Added this error for our own convenience to save time in case of an error. If missing, the test somewhere upstream will cause a NullPointerException and we have to debug it to figure out it already went wrong here.

core/src/test/java/io/parsingdata/metal/EqualityTest.java
```diff
@@ -196,24 +196,28 @@ public void parseGraph() {
 
     @Test
     public void stringRef() {
-        final Ref object = new NameRef("name");
-        assertFalse(object.equals(null));
+        final Ref<String> object = new NameRef("name");
+        assertNotEquals(null, object);
         assertNotEquals("name", object);
         assertEquals(object, new NameRef("name"));
+        assertNotEquals(object, new NameRef("name", "name"));
```
mvanaken, line 203: I added this case, because I think that technically this should be equal. The result of these refs will always be the same, no matter how many times `"name"` was added as a reference. We could consider storing it in a Set first and then convert it to an ImmutableList.

core/src/test/java/io/parsingdata/metal/EqualityTest.java
```diff
@@ -196,24 +196,28 @@ public void parseGraph() {
...
+        assertNotEquals(object, new NameRef("name", "otherName"));
         assertNotEquals(object, new NameRef("otherName"));
         assertNotEquals(object, new DefinitionRef(any("name")));
-        assertNotEquals(object, new NameRef("name", con(1)));
+        assertNotEquals(object, new NameRef( con(1), "name"));
```
mvanaken, line 207: Swapped the arguments here to avoid using deprecated methods in our tests.

core/src/test/java/io/parsingdata/metal/EqualityTest.java
```diff
@@ -196,24 +196,28 @@ public void parseGraph() {
 
     @Test
     public void stringRef() {
-        final Ref object = new NameRef("name");
-        assertFalse(object.equals(null));
+        final Ref<String> object = new NameRef("name");
+        assertNotEquals(null, object);
```
mvanaken, line 200: IntelliJ was nagging about this line stating that `assertFalse(object.equals(null))` should be converted `assertNotEquals(null, object)`. Since the `AutoEqualityTest` also explicitly test for `equals` methods, I obliged IntelliJ here.

core/src/test/java/io/parsingdata/metal/ReferenceValueExpressionSemanticsTest.java
```diff
@@ -107,10 +108,22 @@ public static Collection<Object[]> data() {
...
+            { "[1, 2, 3, 0, 3] a, a, a, 0, sum(ref(1, a.definition))", limitedSum(ref(con(1), any("a"))), stream(1, 2, 3, 0, 3), enc(), true },
+            { "[1, 2, 3, 0, 5] a, a, a, 0, sum(ref(2, a.definition))", limitedSum(ref(con(2), any("a"))), stream(1, 2, 3, 0, 5), enc(), true },
+            { "[1, 2, 3, 0, 6] a, a, a, 0, sum(ref(3, a.definition))", limitedSum(ref(con(3), any("a"))), stream(1, 2, 3, 0, 6), enc(), true },
+            { "[1, 2, 3, 0, 7] a, a, a, 0, sum(ref(4, a.definition))", limitedSum(ref(con(4), any("a"))), stream(1, 2, 3, 0, 6), enc(), true },
+            { "[1, 2, 3, 0, 3] a, a, a, 0, sum(ref(a.definition, 1))", limitedSum(ref(any("a"), con(1))), stream(1, 2, 3, 0, 3), enc(), true },
+            { "[1, 2, 3, 0, 5] a, a, a, 0, sum(ref(a.definition, 2))", limitedSum(ref(any("a"), con(2))), stream(1, 2, 3, 0, 5), enc(), true },
+            { "[1, 2, 3, 0, 6] a, a, a, 0, sum(ref(a.definition, 3))", limitedSum(ref(any("a"), con(3))), stream(1, 2, 3, 0, 6), enc(), true },
+            { "[1, 2, 3, 0, 7] a, a, a, 0, sum(ref(a.definition, 4))", limitedSum(ref(any("a"), con(4))), stream(1, 2, 3, 0, 6), enc(), true }
```
mvanaken, line 126: I added the deprecated methods variants for testing here to achieve 100% test coverage and so it is easy to remove these methods, when we remove the deprecated shorthands in the future.

core/src/test/java/io/parsingdata/metal/expression/value/RefEdgeCaseTest.java
```diff
@@ -29,36 +30,31 @@
 import static io.parsingdata.metal.util.ParseStateFactory.stream;
 import static io.parsingdata.metal.util.TokenDefinitions.any;
 
-import org.junit.Before;
-import org.junit.Rule;
-import org.junit.Test;
-import org.junit.rules.ExpectedException;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
```
mvanaken, line 34: Changes this test class from JUnit 4 to JUnit 5.

core/src/test/java/io/parsingdata/metal/EqualityTest.java
Carly-1, line 207: There is a rogue space before the `con(1)` ;)

Carly-1 (approved):

core/src/test/java/io/parsingdata/metal/EqualityTest.java
rdvdijk: Also remove the trailing white space:

```java
assertNotEquals(object, new NameRef( con(1), "name"));
                                    ^
```

core/src/main/java/io/parsingdata/metal/expression/value/reference/Ref.java
```diff
@@ -91,19 +117,19 @@ private static <T, U extends T> Trampoline<ImmutableList<T>> wrap(final Immutabl
...
-        return getClass().getSimpleName() + "(" + reference + (limit == null ? "" : "," + limit) + ")";
+        return getClass().getSimpleName() + "(" + references + (limit == null ? "" : "," + limit) + ")";
     }
 
     @Override
     public boolean equals(final Object obj) {
         return Util.notNullAndSameClass(this, obj)
-            && Objects.equals(reference, ((Ref)obj).reference)
+            && Objects.equals(references, ((Ref)obj).references)
             && Objects.equals(limit, ((Ref)obj).limit);
```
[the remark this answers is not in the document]
rdvdijk: And although unrelated, similar casts in `equals` can be improved in `ImmutableList` and `ImmutablePair`.

core/src/test/java/io/parsingdata/metal/EqualityTest.java
mvanaken: Removed! :)

core/src/main/java/io/parsingdata/metal/expression/value/reference/Ref.java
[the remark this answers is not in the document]
mvanaken: Sure, done!

core/src/main/java/io/parsingdata/metal/expression/value/reference/Ref.java
```diff
@@ -48,24 +52,36 @@
...
-        public NameRef(final String reference, final SingleValueExpression limit) { super(reference, value -> value.matches(reference), limit); }
+        public NameRef(final String reference, final String... references) { this(null, reference, references); }
+        public NameRef(final SingleValueExpression limit, final String reference, final String... references) { super(ParseValue::matches, limit, reference, references); }
     }
 
     public static class DefinitionRef extends Ref<Token> {
-        public DefinitionRef(final Token reference) { this(reference, null); }
-        public DefinitionRef(final Token reference, final SingleValueExpression limit) { super(reference, value -> value.definition.equals(reference), limit); }
+        public DefinitionRef(final Token reference, final Token... references) { this(null, reference, references); }
+        public DefinitionRef(final SingleValueExpression limit, final Token reference, final Token... references) { super((value, ref) -> value.definition.equals(ref), limit, reference, references); }
```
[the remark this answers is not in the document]
mvanaken: Ah, similar as done for NameRef. Sure!

However, did you really meant to write `this.definition == definition`? Or was that pseudocode? ;)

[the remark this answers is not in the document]
mvanaken: Lol, I forgot about my own questions... ;)

I kept it as an overload method for `matches` and implemented the method with `this.definition.equals(definition)`.

Issue #394, opened 2023-07-31
Convert all tests from JUnit4 to JUnit5.


mvanaken:
We are mixing up JUnit4 and JUnit 5 tests. To avoid confusion what to import and use, convert all tests to JUnit 5 and remove the JUnit4 dependencies.

PR #395, opened 2023-07-31
#394 Convert all test to JUnit5.

mvanaken:
To avoid confusion about which import to pick, decided to convert all tests to JUnit5. The steps consists of the following removals:

* Removal of the @Rule annotations and replaced it with assertThrows() test methods instead
* Removal of the @RunWith(Parameterized) for classes and replaced it with @ParameterizedTest on test methods.
* Changing imports from org.junit.(...) to org.juni.jupiter.api.Assertions.(...).

These steps required some additional changes to be made, especially around the [ParameterizedParse](https://github.com/parsingdata/metal/pull/395/files#diff-c7383c4315c5343a8d14883a0f87de903a6a11883d5d5c1b56b42b27348ac3a2) class.

Note: the old test suite before this branch contained 1147+16 tests and so does this run, so all tests are accounted for.

- [x] Find out why there are 27 more tests. - Checked the wrong commit. :see_no_evil: 
- [x] Change destination to master when #393 is merged.

### DIFF
--- a/core/src/test/java/io/parsingdata/metal/AutoEqualityTest.java
+++ b/core/src/test/java/io/parsingdata/metal/AutoEqualityTest.java
@@ -24,9 +24,9 @@ import static java.util.function.Predicate.not;
 import static java.util.stream.Collectors.toList;
 import static java.util.stream.Collectors.toSet;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertNotEquals;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNotEquals;
 
 import static io.parsingdata.metal.Shorthand.TRUE;
 import static io.parsingdata.metal.Shorthand.con;
@@ -64,10 +64,10 @@ import java.util.function.Supplier;
 import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
 
-import org.junit.Test;
-import org.junit.runner.RunWith;
-import org.junit.runners.Parameterized;
-import org.junit.runners.Parameterized.Parameter;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
 
 import io.parsingdata.metal.data.ByteStream;
 import io.parsingdata.metal.data.ByteStreamSource;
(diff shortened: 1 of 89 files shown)


core/src/test/java/io/parsingdata/metal/token/TokenRefTest.java
rdvdijk, line 1: Minor and unrelated: the method `createNamedTokens` can use some `final`s for the arguments.

Feel free to ignore this, I noticed that there are quite a few more places we could improve this.

core/src/test/java/io/parsingdata/metal/AutoEqualityTest.java
```diff
@@ -361,32 +355,36 @@ private static List<Object> instantiate(final List<Supplier<Object>> suppliers)
         return output;
     }
 
-    @Test
-    public void notEqualsNull() {
+    @ParameterizedTest
+    @MethodSource("data")
+    public void notEqualsNull(final Object object, final Object same, final Object[] other) {
         assertFalse(object.equals(null));
```
rdvdijk, line 361: IntellIJ IDEA suggests `assertNotEquals(null, object);` here (and in a few other places in this test class, `ConstantSliceTest` and `EqualityTest`).

rdvdijk (approved):
Looks good!

core/src/test/java/io/parsingdata/metal/ImmutableListTest.java
```diff
@@ -146,6 +146,7 @@ public void sizeEmpty() {
         assertEquals(0, new ImmutableList<ParseValue>().size);
     }
 
+
```
Carly-1, line 149: I think this change can be reverted.

Carly-1 (approved):
Nice improvement :+1:

core/src/test/java/io/parsingdata/metal/AutoEqualityTest.java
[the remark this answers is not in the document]
mvanaken: I'll run a "Optimize imports" on entire project and then hoping my import settings for Metal or correct. :crossed_fingers: :grimacing:

formats/src/test/java/io/parsingdata/metal/format/FormatTest.java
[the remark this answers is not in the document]
mvanaken: Done!

core/src/test/java/io/parsingdata/metal/AutoEqualityTest.java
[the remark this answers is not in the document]
mvanaken: Two files needed adjusting. Changed it! Tnx!

core/src/test/java/io/parsingdata/metal/AutoEqualityTest.java
mvanaken: The `equals` method is the method being tested here, so it is deliberate that the test is written like this. We want to ignore any other checks that is possibly checked when calling `assertNotEquals`. This is the test that really checks if the `equals` method returns a result as expected, so I am going to keep it as is.

core/src/test/java/io/parsingdata/metal/data/callback/CallbackTest.java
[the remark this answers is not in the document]
mvanaken: Yes, there are many places where this is the case. I think I am going to leave that for a different issue. :)

core/src/test/java/io/parsingdata/metal/token/TokenRefTest.java
mvanaken: Agreed! I'll ignore it for now. :wink: I'll see if there are many other places and if it is a lot, I'll create a separate issue for it.

PR #397, opened 2023-08-17
Added message digest value expression

gertjanal:
Validate a message digest for large data streams

### DIFF
--- a/core/src/main/java/io/parsingdata/metal/Util.java
+++ b/core/src/main/java/io/parsingdata/metal/Util.java
@@ -17,9 +17,16 @@
 package io.parsingdata.metal;
 
 import static java.math.BigInteger.ZERO;
+import static java.math.BigInteger.valueOf;
+import static java.security.MessageDigest.getInstance;
+
+import static io.parsingdata.metal.data.Slice.createFromBytes;
+import static io.parsingdata.metal.data.Slice.createFromSource;
 
 import java.io.ByteArrayOutputStream;
 import java.math.BigInteger;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
 import java.util.Locale;
 import java.util.Optional;
 import java.util.zip.DataFormatException;
@@ -81,7 +88,7 @@ public final class Util {
 
     public static String bytesToHexString(final byte[] bytes) {
         checkNotNull(bytes, "bytes");
-        char[] hexChars = new char[bytes.length * 2];
+        final char[] hexChars = new char[bytes.length * 2];
         for (int j = 0; j < bytes.length; j++) {
             final int v = bytes[j] & 0xFF;
             hexChars[j * 2] = HEX_ARRAY[v >>> 4];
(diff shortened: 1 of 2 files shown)


gertjanal:
I have no idea why the JDK11 build fails:

```
[ERROR] Failures: 
[ERROR]   AutoEqualityTest.data:262->generateObjectArrays:314->generateObjectArrays:325 Please add a mapping for type Object
```

core/src/main/java/io/parsingdata/metal/Util.java
```diff
@@ -111,6 +118,33 @@ public Optional<Value> eval(final Value value, final ParseState parseState, fina
...
+            @Override
+            public Optional<Value> eval(final Value value, final ParseState parseState, final Encoding encoding) {
+                try {
+                    final Slice slice = value.slice();
+                    final BigInteger bufferSize = valueOf(512);
+                    final MessageDigest digester = getInstance(algorithm);
+
+                    for (BigInteger i = ZERO; i.compareTo(slice.length) < 0; i = i.add(bufferSize)) {
+                        final BigInteger length = bufferSize.min(slice.length.subtract(i));
+                        final Optional<Slice> subslice = createFromSource(slice.source, slice.offset.add(i), length);
```
mvanaken, line 132: Instead of calculating the exact slice you can also create a slice where only the offset is moved.
I.e.:
Your implementation uses this as slices: 
```
|***********|
|***|
   |***|
      |***|
         |**|
```
But what if we create slices as:
```
|***********|
|***********|
   |********|
      |*****|
         |**|
```
In combination with the `getData(final int limit)` method on `slice`, the calculations would be a easier.

core/src/main/java/io/parsingdata/metal/Util.java
```diff
@@ -111,6 +118,33 @@ public Optional<Value> eval(final Value value, final ParseState parseState, fina
...
+                        if (subslice.isEmpty()) {
+                            return Optional.empty();
+                        }
+                        digester.update(subslice.get().getData());
```
mvanaken, line 136: Can we also do without this `get()`?

Also, I think the windowing part is very interesting and other implementations may benefit from it was well. We may consider to extract the windowing part to a separate value expression so we can reuse it for other implementations.

core/src/main/java/io/parsingdata/metal/Util.java
```diff
@@ -111,6 +118,33 @@ public Optional<Value> eval(final Value value, final ParseState parseState, fina
         };
     }
 
+    public static ValueExpression digest(final String algorithm, final ValueExpression target) {
+        return new UnaryValueExpression(target) {
```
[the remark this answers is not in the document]
gertjanal: Created package a suggested

core/src/main/java/io/parsingdata/metal/Util.java
gertjanal: Can you give a code example? I don't know how to do this. `createFromSource` is the only public way to create new slices, right?

gertjanal: @mvanaken

mvanaken: I was thinking about something like this:
```java
Slice slice = value.slice();
while (slice != null) {
    digester.update(slice.getData(bufferSize));
    slice = createFromSource(slice.source, slice.offset.add(bufferSize), slice.length.subtract(bufferSize)).orElse(null);
}
return Optional.of(new CoreValue(createFromBytes(encoding.byteOrder.apply(digester.digest())), encoding));
```
if you want to differentiate between "end of slicing" and "slice should exist but isn't", you could extend it with:
```java
Slice slice = value.slice();
while (slice != null && slice.length.compareTo(bufferSize) > 0) {
    digester.update(slice.getData(bufferSize));
    slice = createFromSource(slice.source, slice.offset.add(bufferSize), slice.length.subtract(bufferSize)).orElse(null);
}
if (slice == null) {
    return Optional.empty();
}
digester.update(slice.getData(bufferSize));
return Optional.of(new CoreValue(createFromBytes(encoding.byteOrder.apply(digester.digest())), encoding));
```
This is all untested, so I may have missed something.
There is not much difference between the second option and your code, but it does avoid the Optional.get() call.

Issue #399, opened 2023-10-24
Add cache for more efficient refs


rdvdijk:
Using `ref` is sometimes extremely expensive when referencing values parsed very "early" in large `ParseGraph`s. This is a common pitfall in many tokens. We can improve this by creating a cache for all parsed values, which is used by evaluating a `ref`.

An example: the CFB format contains a 'sector size' in its header, which is used throughout the token. The lookup of this sector size becomes extremely slow in CFB with large sector chains.

PR #401, opened 2023-11-02
#399 Introducing ParseValueCache.

mvanaken:
A cache is added to the parse state as a lookup map for values by name.
The cache is used where possible. If it is unable to use the cache, the normal lookup through the parse graph is used, which is the original behaviour.

This only creates a lookup map by name, not by token.
@ reviewers, do we want to include that?


core/src/test/java/io/parsingdata/metal/ToStringTest.java
```diff
@@ -173,18 +174,26 @@ public void encoding() {
...
+        final String parseStateString = "ParseState(source:ByteStreamSource(InMemoryByteStream(2));offset:0;order:pg(EMPTY);cache:size=0)";
         assertEquals(parseStateString, parseState.toString());
+
         final ParseState parseStateWithIterations = parseState.addBranch(rep(def("a",1))).iterate();
-        final String parseStateWithIterationsString = "ParseState(source:ByteStreamSource(InMemoryByteStream(2));offset:0;order:pg(pg(terminator:Rep),pg(EMPTY),true);iterations:>Rep(Def(a,Const(0x01)))->1)";
+        final String parseStateWithIterationsString = "ParseState(source:ByteStreamSource(InMemoryByteStream(2));offset:0;order:pg(pg(terminator:Rep),pg(EMPTY),true);iterations:>Rep(Def(a,Const(0x01)))->1;cache:size=0)";
         assertEquals(parseStateWithIterationsString, parseStateWithIterations.toString());
+
+        final ParseState parseStateWithoutCache = parseStateWithIterations.withOrder(parseStateWithIterations.order);
+        final String parseStateWithoutCacheString = "ParseState(source:ByteStreamSource(InMemoryByteStream(2));offset:0;order:pg(pg(terminator:Rep),pg(EMPTY),true);iterations:>Rep(Def(a,Const(0x01)))->1;no-cache)";
```
mvanaken, line 185: Included this parse state variant for the 'no-cache' coverage.

core/src/main/java/io/parsingdata/metal/data/ParseState.java
```diff
@@ -34,64 +35,66 @@
...
     }
 
     public Optional<ParseState> seek(final BigInteger newOffset) {
-        return newOffset.compareTo(ZERO) >= 0 ? Optional.of(new ParseState(order, source, newOffset, iterations, references)) : Optional.empty();
+        return newOffset.compareTo(ZERO) >= 0 ? Optional.of(new ParseState(order, cache, source, newOffset, iterations, references)) : Optional.empty();
     }
 
     public ParseState withOrder(final ParseGraph order) {
-        return new ParseState(order, source, offset, iterations, references);
+        return new ParseState(order, NO_CACHE, source, offset, iterations, references);
```
mvanaken, line 93: `withOrder` is used when using scopes. Since scopes define a limitation to the search region, which we do not track within the cache, so we should disable it. Note that it is only disabled when evaluating the expression.

core/src/test/java/io/parsingdata/metal/data/ParseValueCacheTest.java
```diff
@@ -0,0 +1,173 @@
...
+            arguments("scoped definitionRef", scope(ref(pv2Definition), con(1)), false),
+            arguments("scoped definitionRef with limit", scope(ref(con(1), pv2Definition), con(1)), false),
+            arguments("scoped multi definitionRef", scope(ref(pv2Definition, pv3Definition), con(1)), false),
+            arguments("scoped multi definitionRef with limit", scope(ref(con(1), pv2Definition, pv3Definition), con(1)), false)
+        );
+    }
+
+    @ParameterizedTest(name="{2} - {0}")
+    @MethodSource
+    public void cacheUsageTest(final String testName, final ValueExpression expression, final boolean shouldUseCache) {
```
mvanaken, line 166: This tests if the cache is used. The other tests within this class make sure that the cache works as expected. Other tests outside this class, tests if the results are expected. This tests shows in which cases the cache is actually used.

rdvdijk:
I haven't reviewed this yet (I will, Soon™), but I have a question: Does the `ParseValueCache` work nicely with `Scope`?

Last week we (successfully) used `Scope`, and I immediately thought of this new cache.

(You might already have added a test for this, please ignore me if you have.)

mvanaken:
> I haven't reviewed this yet (I will, Soon™), but I have a question: Does the `ParseValueCache` work nicely with `Scope`?
> 
> Last week we (successfully) used `Scope`, and I immediately thought of this new cache.
> 
> (You might already have added a test for this, please ignore me if you have.)

Yes and no. :blush: 
Yes, because everything still works as expected when using scope.
And no, because the cache is turned off when evaluating scoped value expressions, see also my [comment here](https://github.com/parsingdata/metal/pull/401#discussion_r1380776720).

I have added a test, see [my other comment here](https://github.com/parsingdata/metal/pull/401/files#r1381478099), to test when the cache is used and when it is not. For all scoped variants it is turned off.

core/src/main/java/io/parsingdata/metal/data/ParseValueCache.java
```diff
@@ -0,0 +1,85 @@
...
+        this.cache = cache;
+    }
+
+    public Optional<ImmutableList<Value>> find(final String scopeName, int limit) {
+        if (this == NO_CACHE) {
+            return Optional.empty();
+        }
+        final String s = shortName(scopeName);
+        ImmutableList<ParseValue> valueImmutableList = cache.getOrDefault(s, new ImmutableList<>());
+        ImmutableList<Value> result = new ImmutableList<>();
```
jvdb, line 41: With these two mutable variables we are letting go of the mostly functional style Metal was written in. I don't mind necessarily, it's just that we should be aware that we're changing our style.

core/src/main/java/io/parsingdata/metal/expression/value/reference/Ref.java
```diff
@@ -50,7 +51,7 @@
  * the amount of returned results.
  * @param <T> The type of reference to match on.
  */
-public class Ref<T> implements ValueExpression {
+public abstract class Ref<T> implements ValueExpression {
```
jvdb, line 54: Did we just forget to mark this `abstract`? I try to remember if we had concrete plans to actually use this class in a stand-alone fashion...

core/src/test/java/io/parsingdata/metal/data/ParseValueCacheTest.java
```diff
@@ -0,0 +1,173 @@
...
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+import org.junit.jupiter.params.provider.ValueSource;
+
+import io.parsingdata.metal.expression.value.Value;
+import io.parsingdata.metal.expression.value.ValueExpression;
+import io.parsingdata.metal.token.Token;
+
+class ParseValueCacheTest {
```
jvdb, line 36: Awesome to have these highly specific unit tests!

core/src/test/java/io/parsingdata/metal/expression/value/reference/RefMultiReferencesTest.java
```diff
@@ -0,0 +1,113 @@
...
+import org.junit.jupiter.params.provider.MethodSource;
+
+import io.parsingdata.metal.data.Environment;
+import io.parsingdata.metal.data.ImmutableList;
+import io.parsingdata.metal.data.ParseState;
+import io.parsingdata.metal.expression.value.Value;
+import io.parsingdata.metal.expression.value.ValueExpression;
+import io.parsingdata.metal.token.Token;
+
+class RefMultiReferencesTest {
```
jvdb, line 37: Also great addition to test suite!

jvdb (approved):
Looks good!

core/src/main/java/io/parsingdata/metal/expression/value/reference/Ref.java
mvanaken: Good point! I think the abstract modifier could have been added earlier. Since we added an abstract method in this PR, it is mandatory now.

core/src/main/java/io/parsingdata/metal/data/ParseValueCache.java
mvanaken: Yes, good point. For consistency changed it to use trampolines.

Carly-1 (approved):
Looks good to me 😊

rdvdijk (approved):
Looks great to me! And: excellent tests 👍

Issue #402, opened 2023-11-08
Improve hash implementations for immutable objects.


mvanaken:
Some hash implementations in Metal are expensive and we should improve that to boost performance of Metal.

The general contract (JavaDoc) of hashCode() states:

1. Whenever it is invoked on the same object more than once during an execution of a Java application, hashCode() must consistently return the same value, provided no information used in equals comparisons on the object is modified. This value doesn’t need to stay consistent from one execution of an application to another execution of the same application.
2. If two objects are equal according to the equals(Object) method, calling the hashCode() method on each of the two objects __must__ produce the same value.
3. It is _not_ required that if two objects are unequal according to the [equals(java.lang.Object)](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/Objects.html#equals(java.lang.Object,java.lang.Object)) method, then calling the hashCode method on each of the two objects must produce distinct integer results. However, the programmer should be aware that producing distinct integer results for unequal objects may improve the performance of hash tables. As much as is reasonably practical, the hashCode method defined by class Object does return distinct integers for distinct objects.

In addition, see e.g. [University of California, Berkeley](https://inst.eecs.berkeley.edu/~cs61bl/r//cur/hashing/good-hashcodes.html?topic=lab19.topic&step=2&course=), a good hash function meets the following criteria:

4. All objects that are .equals, must have the same hashCode. => equal to contract criteria 2
5. hashCode values should be spread as evenly as possible over all ints.
6. hashCode should be relatively quick to compute.
7. hashCode must be deterministic (not random). => equal to contract criteria 1

Criteria 1 and 7 are equal, criteria 2 and 4 are equal, and criteria 5 is met by using the Object.hash() method as much as possible which helps to spread the hashcodes evenly.

So, to implement good hash codes, we should consider the following simplified criteria:
- hashCode must be deterministic (not random).
- All objects that are .equals, must have the same hashCode.
- It is _not_ required that if two objects are unequal, the hashCode is unequal.
- hashCode should be relatively quick to compute.

All of the hash function implementations within metal meet these criteria, except for the last. Not all hashCodes are quick to compute.
The `DataExpressionSource` calculates the hash of the whole parseState it is dependent on and this source is created when using `Tie`s in your tokens.
When Values from a Tie of a large parsegraph are placed within a hashmap, this may cause a significant performance drop.

There are multiple solutions possible:
[...]

PR #403, opened 2023-11-08
#402: Hash calculation improvements

mvanaken:
See also issue #402 , which explains the problem and possible solutions.

This PR implements hash caching for all immutable objects, which basically means for all objects within metal, so many files are adjusted. In short:
* [ImmutableObject](https://github.com/parsingdata/metal/pull/403/files#diff-81f3e3696155918279f269adae29f82bb381dcfbe181b589286a6d269a4d86e5) caches the hash so the hash is only calculated once.
* All immutable objects within metal extends this object
* [A test](https://github.com/parsingdata/metal/pull/403/files#diff-2de103cc58b5476f76502f919b02ba47b83ff2528554d554d9022e8ff6319fd3) is added to test that the hash is indeed only calculated once, and it contains a performance test, which will not even finish without caching.

We could also simplify the calculation of certain classes as well, e.g. ParseState, but I left that out for now. See [my comment here](https://github.com/parsingdata/metal/pull/403/files#r1386546508) for the performance comparison.

The build is :x:, because Sonar is not content  with this change, see [my comment here](https://github.com/parsingdata/metal/pull/403#issuecomment-1801855645). I would like to hear your thought about this as well.

Update: marked the Sonar remarks about unimplemented hashCode as false positives, since they are implemented, but Sonar is unable to detect it.
Update2: removed unused imports, and now we have 100% coverage again.


core/src/main/java/io/parsingdata/metal/data/ParseState.java
```diff
@@ -116,7 +117,7 @@ public boolean equals(final Object obj) {
     }
 
     @Override
-    public int hashCode() {
+    public int cachingHashCode() {
         return Objects.hash(getClass(), order, offset, source, iterations, references);
```
mvanaken, line 121: We may consider doing something here like:
```suggestion
        return Objects.hash(getClass(), order.size, offset, source, iterations.size, references.size);
```
to keep hash calculations simple and quick.

I ran the performance test within ImmutableObjectTest for a `dataBlockCount` of 512 and switched between hash caching on/off and parsestate hash simplified on/off. That resulted into the following performance matrix:
|      | No hash caching  | Hash caching |
| --- | ------------- | ------------- |
| **Original ParseState hash** | undetermined | 730ms |
| **Simplified ParseState hash** | 3600ms | 700ms |

Since the hash caching contributes most to the performance gain, I implemented that first.

mvanaken:
> So, how do we shut Sonar up? ;)

Discussed with @jvdb: we should mark these as false positives, because Sonar is unable to detect that the hashcode is also implemented. The ImmutableObject makes it mandatory.

core/src/main/java/io/parsingdata/metal/ImmutableObject.java
```diff
@@ -0,0 +1,26 @@
...
+ * It will improve performance if these objects caches their hash.
+ * <p>
+ * This is a lazy implementation, instead of calculating the hash once within the constructor, to avoid
+ * performance decrease during parsing. In most implementations the hash is not actually used/needed.
+ */
+public abstract class ImmutableObject {
+
+    private Integer hash;
+
+    public abstract int cachingHashCode();
```
jvdb, line 14: Maybe this should have a name that more reflects how an implementer should think about it? Such as `fixedHashCode()` or 'initFixedHashCode()` or `permanentHashCode()`?

Maybe `cachingHashCode()` explains that well, I'm just not sure about it because it's also about what it's used for (which is a kind of documentation, but also orthogonal to its functioning?)

core/src/main/java/io/parsingdata/metal/data/ParseState.java
jvdb: Since the impact of hash caching itself is so huge, I would wait with further optimization until there is a clear practical case where that needs to be added. That said, I don't oppose this change if you want to add it.

core/src/test/java/io/parsingdata/metal/ImmutableObjectTest.java
```diff
@@ -0,0 +1,104 @@
...
+
+import io.parsingdata.metal.data.ImmutableList;
+import io.parsingdata.metal.data.ParseState;
+import io.parsingdata.metal.data.ParseValue;
+import io.parsingdata.metal.data.Selection;
+import io.parsingdata.metal.expression.value.Value;
+import io.parsingdata.metal.token.Token;
+import io.parsingdata.metal.util.InMemoryByteStream;
+
+class ImmutableObjectTest {
```
jvdb, line 35: Nice tests!

jvdb (approved):
Solid improvement!

core/src/main/java/io/parsingdata/metal/ImmutableObject.java
```diff
@@ -0,0 +1,26 @@
+package io.parsingdata.metal;
+
+/**
+ * When objects are immutable, that means their hashcode will stay the same the moment it is created.
+ * It will improve performance if these objects caches their hash.
```
Carly-1, line 5: Nitpick:
```suggestion
 * It will improve performance if these objects cache their hash.
```

core/src/main/java/io/parsingdata/metal/ImmutableObject.java
Carly-1: You could also add some javadoc to explain this more clearly? Not sure, maybe naming is enough.

core/src/test/java/io/parsingdata/metal/ImmutableObjectTest.java
```diff
@@ -0,0 +1,104 @@
...
+            @Override
+            public int cachingHashCode() {
+                counter.incrementAndGet();
+                return 15;
+            }
+        };
+        assertEquals(15, immutableObject.hashCode());
+        assertEquals(15, immutableObject.hashCode());
+        assertEquals(15, immutableObject.hashCode());
+        assertEquals(1, counter.get());
```
Carly-1, line 67: Nice 😄

Carly-1 (approved):
Very nice performance improvement!

core/src/main/java/io/parsingdata/metal/ImmutableObject.java
mvanaken: What do you think about `immutableHashCode` as alternative?

> jvdb: Ja, goed, goed!

Alrighty then!

core/src/main/java/io/parsingdata/metal/ImmutableObject.java
```diff
@@ -0,0 +1,26 @@
+package io.parsingdata.metal;
+
+/**
+ * When objects are immutable, that means their hashcode will stay the same the moment it is created.
```
rdvdijk, line 4:
```suggestion
 * When objects are immutable, that means their hash code will stay the same the moment it is created.
```

core/src/main/java/io/parsingdata/metal/ImmutableObject.java
```diff
@@ -0,0 +1,26 @@
...
+ * It will improve performance if these objects cache their hash.
```
rdvdijk, line 5:
```suggestion
 * It will improve performance if these objects cache their hash code.
```

core/src/main/java/io/parsingdata/metal/ImmutableObject.java
```diff
@@ -0,0 +1,26 @@
...
+ * <p>
+ * This is a lazy implementation, instead of calculating the hash once within the constructor, to avoid
+ * performance decrease during parsing. In most implementations the hash is not actually used/needed.
```
rdvdijk, line 8:
```suggestion
 * performance decrease during parsing. In most implementations the hash code is not actually used/needed.
```

core/src/test/java/io/parsingdata/metal/data/SelectionTest.java
```diff
@@ -41,6 +41,9 @@ private Source createSource() {
         return new Source() {
             @Override protected byte[] getData(BigInteger offset, BigInteger length) { return new byte[0]; }
             @Override protected boolean isAvailable(BigInteger offset, BigInteger length) { return true; }
+            @Override public int immutableHashCode() { return 0; }
+            @Override
+            public boolean equals(Object obj) {return obj == this;}
```
rdvdijk, line 46: Minor style consistency nitpick:

```suggestion
            public boolean equals(Object obj) { return obj == this; }
```

core/src/test/java/io/parsingdata/metal/data/selection/ByTypeTest.java
```diff
@@ -38,6 +38,9 @@ public class ByTypeTest {
     public static final Source EMPTY_SOURCE = new Source() {
         @Override protected byte[] getData(BigInteger offset, BigInteger length) { throw new IllegalStateException(); }
         @Override protected boolean isAvailable(BigInteger offset, BigInteger length) { return false; }
+        @Override public int immutableHashCode() { return -1; }
+        @Override
+        public boolean equals(Object obj) {return obj == this;}
```
rdvdijk, line 43: See [this suggestion](https://github.com/parsingdata/metal/pull/403/files#r1404326638).

core/src/test/java/io/parsingdata/metal/ImmutableObjectTest.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    @Test
+    void performanceTest() {
+        // This test would take way too much time without hash caching.
```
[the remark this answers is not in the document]
rdvdijk: Ah, `@Timeout`: https://www.baeldung.com/java-junit-5-timeout-annotation

Maybe add it, to make sure we never have a regression here?

core/src/main/java/io/parsingdata/metal/ImmutableObject.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+public abstract class ImmutableObject {
+
+    private Integer hash;
```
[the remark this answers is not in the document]
mvanaken: Sure! I'll change it!

rdvdijk (approved):
Added some comments for minor improvements.

core/src/main/java/io/parsingdata/metal/ImmutableObject.java
rdvdijk, line 12: Another consideration: This is an `Integer`, and not an `int`, because it can be `null`.

Using boxed primitives isn't "free", though. The `int` would be 4 bytes, but each boxed `Integer` is 16 or even 24 bytes (according to [this Stackoverflow answer](https://stackoverflow.com/a/9037808)).

One alternative here could be to have a primitive `int hashCode` and a `boolean initialized`.

Not sure if this is worth the effort, but since most classes now inherit from this one, and they are used all over the `ParseGraph`, it might be worth considering.

This would be an easy fix in the future, so for now I think it's okay to keep this current implementation.

core/src/test/java/io/parsingdata/metal/ImmutableObjectTest.java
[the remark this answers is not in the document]
mvanaken: I did add that, but I have removed it again for two reasons:
1. The timing specified is an absolute timing and this test is dependent on the agent on GitHub, which could fluctuate. So that number would be arbitrary.
2. The TimeOut annotation does not stop the test. It simply checks if the time it took exceeds the timeout value. That means that the run would still run a very long time in case the hash cache is turned off and that would most likely trigger a GitHub error sooner than it would trigger the timeout error.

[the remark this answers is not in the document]
mvanaken: Discussed it again with @jvdb. I'm going to tune the numbers in this test, such that without hash cashing it would take e..g 30 seconds - 1 minute to run, while with hash cashing in a few milliseconds. I could then add a TimeOut annotation with a limit of e.g. 10 seconds. Then we would not have an error build when there is a GitHub glitch or something, and the test would also not run for a very long time in case hash cashing is (accidentally) turned off.

core/src/main/java/io/parsingdata/metal/ImmutableObject.java
mvanaken: Ah, interesting point! I found the use of an object here quite elegant, opposed to a separate boolean, but I haven't considered how much memory this would need. There are two concerns here:
1. Performance
2. Memory usage

1. Performance: I have run the performance test for large numbers and there does not seem to be a difference when using int over Integer. I think the JVM does a lot of auto improvements here.
2. Memory usage: hashCodes are not calculated for all objects. They are only calculated when needed. Since they are initialized as null, they only require as much space as a pointer requires, which is 4 bytes like a primitive int. When we would use ints + booleans as fields, then this requires a little more memory to store the initialize boolean variable. Assuming a boolean requires a bit of space (i mean this literal ;)), and a Integer hashCode requires 160 bits additional bits when calculated, then that means that if we calculate less then 1/160th of the total number of ImmutableObjects, then the Integer implementation actually requires less memory.


Tl;dr;
So, in short, performance is not an issue and since memory usage depends on the implementation on how many objects you want to calculates hashes for (and the default is for 0 objects), it may be wise to keep this as is.

mvanaken:
I'm going to merge this branch even though the codecov analysis failed. That analysis is incorrect, claiming that we have less coverage, but we do not. The analysis on the main branch will be green. If not, pie for everyone, my treat!

Issue #405, opened 2023-11-28
Performance bug in ConcatenatedValueSource


mvanaken:
The ConcatenatedValueSource is very slow when it consists of very large blocks and reading small blocks at a time.
It will first copy the entire large block in memory again and then copy a small part out of it.

The following test runs in 1.5 seconds:

```java
    @Test
    public void concatenatedValueSourceRead() throws IOException {
        final int arraySize = 5_120_000;
        final byte[] bytes = new byte[arraySize];
        new Random().nextBytes(bytes);

        int parts = 4;
        ImmutableList<Value> values = new ImmutableList<>();
        for (int part = 0; part < parts; part++) {
            values = values.add(new CoreValue(Slice.createFromBytes(Arrays.copyOfRange(bytes, (arraySize / parts) * part, (arraySize / parts) * (part + 1))), Encoding.DEFAULT_ENCODING));
        }

        final Value bigValue = ConcatenatedValueSource.create(values)
            .flatMap(source -> createFromSource(source, ZERO, valueOf(arraySize)))
            .map(slice -> new CoreValue(slice, Encoding.DEFAULT_ENCODING)).get();

        long start, end;
        final int readSize = 512;

        final byte[] valueBytes = new byte[arraySize];
        start = System.currentTimeMillis();
        for (int part = 0; part < arraySize / readSize; part++) {
            final byte[] data = bigValue.slice().source.getData(bigValue.slice().offset.add(valueOf(readSize * part)), valueOf(readSize));
            System.arraycopy(data, 0, valueBytes, readSize * part, data.length);
        }
        end = System.currentTimeMillis();
        System.out.printf("Value read: %ss%n", (end - start) / 1000.0);

        assertArrayEquals(bytes, valueBytes);
    }
```

Added this test also to the branch of this issue.

PR #406, opened 2023-11-28
#405 ConcatenatedValueSource performance bug

mvanaken:
The ConcatenatedValueSource was very inefficient when reading smaller blocks of data from large values. It would copy the entire value in an array first and then copy a small portion in a new array again.

Resolved this by adding a `getData` method to the `Slice` class where you can specify the offset next to the limit. This allows to directly read only that data that is needed.

### DIFF
--- a/core/src/main/java/io/parsingdata/metal/data/ConcatenatedValueSource.java
+++ b/core/src/main/java/io/parsingdata/metal/data/ConcatenatedValueSource.java
@@ -17,6 +17,7 @@
 package io.parsingdata.metal.data;
 
 import static java.math.BigInteger.ZERO;
+import static java.math.BigInteger.valueOf;
 
 import static io.parsingdata.metal.Trampoline.complete;
 import static io.parsingdata.metal.Trampoline.intermediate;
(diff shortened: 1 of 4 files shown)


core/src/main/java/io/parsingdata/metal/data/ConcatenatedValueSource.java
```diff
@@ -79,13 +80,15 @@ private Trampoline<byte[]> getData(final ImmutableList<Value> values, final BigI
...
         }
-        if (currentOffset.add(values.head.slice().length).compareTo(offset) <= 0) {
-            return intermediate(() -> getData(values.tail, currentOffset.add(values.head.slice().length), currentDest, offset, length, output));
+        final BigInteger nextOffset = currentOffset.add(values.head.slice().length);
+        if (nextOffset.compareTo(offset) <= 0) {
+            return intermediate(() -> getData(values.tail, nextOffset, currentDest, offset, length, output));
         }
         final BigInteger localOffset = offset.subtract(currentOffset).compareTo(ZERO) < 0 ? ZERO : offset.subtract(currentOffset);
-        final BigInteger toCopy = length.compareTo(values.head.slice().length.subtract(localOffset)) > 0 ? values.head.slice().length.subtract(localOffset) : length;
-        System.arraycopy(values.head.slice().getData(), localOffset.intValueExact(), output, currentDest.intValueExact(), toCopy.intValueExact());
```
mvanaken, line 87: The `values.head.slice().getData()` was the culprit of the slow read. See also [ConcatenatedValueSourceTest](https://github.com/parsingdata/metal/pull/406/files#diff-cd46a952cb6841c5b74fb506794539df74cc0b9571907a8643c61cf8faefdaf7).

core/src/test/java/io/parsingdata/metal/data/ConcatenatedValueSourceTest.java
```diff
@@ -90,4 +100,41 @@ public void checkData(final String description, final int offset, final int leng
         }
     }
 
+    @Test
+    @Timeout(value=1)
+    public void concatenatedValueSourceRead() {
```
mvanaken, line 105: This test took 1.5 seconds on my computer without the fix, and 0.055 seconds with the fix. :tada:

core/src/main/java/io/parsingdata/metal/data/Slice.java
```diff
@@ -58,8 +58,19 @@ public byte[] getData() {
...
     public byte[] getData(final BigInteger limit) {
-        final BigInteger calculatedLength = checkNotNegative(limit, "limit").compareTo(length) > 0 ? length : limit;
-        return source.getData(offset, calculatedLength);
+        return getData(ZERO, limit);
+    }
+
+    /**
+     * Return a part of the data specified by the offset and limit.
+     * @param offset the offset to start reading the slice from
+     * @param limit the maximum number of bytes returned. Less bytes are returned if the end of slice is reached.
```
Carly-1, line 67: Nitpick:
```suggestion
     * @param limit the maximum number of bytes returned. Fewer bytes are returned if the end of slice is reached.
```

core/src/test/java/io/parsingdata/metal/data/ConcatenatedValueSourceTest.java
```diff
@@ -90,4 +98,37 @@ public void checkData(final String description, final int offset, final int leng
...
+        // Create a large array with random data
+        final int arraySize = 5_120_000;
+        final byte[] bytes = new byte[arraySize];
+        new Random().nextBytes(bytes);
+
+        // Split the data in separate CoreValues.
+        int parts = 4;
```
Carly-1, line 110:
```suggestion
        final int parts = 4;
```

Carly-1 (approved):
Nice improvement! And nice tests :)

rdvdijk (approved):

Issue #381, opened 2023-05-12
Using scope makes ref inefficient


mvanaken:
In order to find the requested scope it first finds all the scopes and then selects the correct scope from that list.
If we then call for ref it would then traverse that scoped graph twice.

This can be made more efficient if we keep track of the scope while requesting the values with ref, similar as the limit parameter in Ref.

In fact, scope is only used in combination with Ref, because it would not make sense in any other context. It would therefor make much more sense if scope is a parameter for Ref. For example, a scoped Ref could look something like:
last(ref("name", NO_LIMIT, con(0))

Since Ref already has a parameter for the limit (which is 1 in case of last), we should think about how to define the shorthand for a scoped ref.

PR #411, opened 2024-02-10
#381 Move Scope ValueExpressions as parameter of Ref.

mvanaken:
The Scope value expression makes no sense without a Ref. 
The Scope class is removed and the scopeLevel is moved to a parameter for Ref instead.
Changed tests accordingly.

Since the scope level and the limit level are now both available within a Ref, we only have to traverse the ParseGraph once and keep track of both parameters when collecting values. This may have a slight performance gain.
Also note, that previously the ParseValueCache was never used in combination with Scope class. Now, we can check the value of the scopeLevel to determine if we can use the cache or not.


core/src/main/java/io/parsingdata/metal/Shorthand.java
```diff
@@ -349,7 +348,11 @@ private Shorthand() {}
     public static BinaryValueExpression mapRight(final BiFunction<ValueExpression, ValueExpression, BinaryValueExpression> func, final SingleValueExpression leftExpand, final ValueExpression right) { return func.apply(exp(leftExpand, count(right)), right); }
 
     /** @see Bytes */ public static ValueExpression bytes(final ValueExpression operand) { return new Bytes(operand); }
-    /** @see Scope */ public static ValueExpression scope(final ValueExpression scopedValueExpression, final SingleValueExpression scopeSize) { return new Scope(scopedValueExpression, scopeSize); }
+
+    /** @see Ref */ public static NameRef scope(final NameRef operand) { return scope(operand, con(0)); }
```
mvanaken, line 352: Do we think these shorthands (with the default scope of 0) could be useful?

jvdb: Maybe, but I'm not sure about the name, maybe `scopedRef`? But maybe all `Ref`s that have the `scopeSize`-argument should be called that? I haven't given it much thought yet, but it might be worth considering.

jvdb (approved):
Looks good. Also worth noting: we've been using this change at Infix for quite a while now and it works really well!

core/src/main/java/io/parsingdata/metal/Shorthand.java
rdvdijk: What would be a typical use case for a `Ref` with scope of 0?

core/src/main/java/io/parsingdata/metal/data/ParseState.java
```diff
@@ -42,61 +42,63 @@ public class ParseState extends ImmutableObject {
     public final Source source;
     public final ImmutableList<ImmutablePair<Token, BigInteger>> iterations;
     public final ImmutableList<ParseReference> references;
+    public final int scopeDepth;
 
-    public ParseState(final ParseGraph order, final ParseValueCache cache, final Source source, final BigInteger offset, final ImmutableList<ImmutablePair<Token, BigInteger>> iterations, final ImmutableList<ParseReference> references) {
+    public ParseState(final ParseGraph order, final ParseValueCache cache, final Source source, final BigInteger offset, final ImmutableList<ImmutablePair<Token, BigInteger>> iterations, final ImmutableList<ParseReference> references, final int scopeDepth) {
```
rdvdijk, line 47: A constructor with seven arguments, oof :sweat_smile:

I don't think we should address this now, but we should discuss this when/if we refactor the `ParseGraph` implementation.

rdvdijk (approved):
Nice!

core/src/main/java/io/parsingdata/metal/Shorthand.java
mvanaken: That means the most local scope that is possible, i.e. only values in the current seq.
I think it is the scope level that will be used most, hence the shorthand for it.

core/src/main/java/io/parsingdata/metal/data/ParseState.java
mvanaken: I just created #415 , where we are going to move the scopeDepth to the ParseGraph instead. This way, the `withOrder` method can be used correctly and the ParseState will not hold an incorrect scopeDepth. I developed it in our fork for now, but will create a separate PR here.

core/src/main/java/io/parsingdata/metal/Shorthand.java
mvanaken: > Maybe, but I'm not sure about the name, maybe `scopedRef`? But maybe all `Ref`s that have the `scopeSize`-argument should be called that? I haven't given it much thought yet, but it might be worth considering.

If we would call it `scopedRef`, then usage would look like `scopedRef(ref("name"))`. I like to keep it `scope` shorthand, especially since it is now always a `ref`, so including `Ref` is redundant to me.

Issue #413, opened 2024-02-16
Missing test for DefUntil


mvanaken:
I am missing a test for DefUntil, where the arguments use lists instead of single values.

PR #414, opened 2024-02-16
#413: Add test to DefUntil to test the list semantics.

mvanaken:
This test shows how multiple values for initialSize, stepSize and maxSize are used together and which offsets in which order are used to parse the terminator.

This was missing in our test base. We had 100% test coverage, because also with single values, we process them as lists. Clear example that line coverage is not the same as test coverage. ;)


mvanaken:
Because I simply added a test, decided to merge it without any reviews.

Issue #415, opened 2024-03-13
Bug when using `withOrder` method in ParseState.


mvanaken:
When you would use the `withOrder` method in ParseState to create a ParseState with a different ParseGraph, then the scopeDepth that is stored within the ParseState does not match the ParseGraph anymore. This may result into an error when using scoped references, where no value is found, because the scopeDepth is too high to begin with.

With that in mind, isn't it better to store the scopeDepth within the ParseGraph, instead of ParseState? Since the scopeDepth can be determined by traversing the ParseGraph, it shows that it this bug can be fixed if we move the field.

PR #422, opened 2024-03-15
#415 Move ScopeDepth from ParseState to ParseGraph.

mvanaken:
See also issue #415.

When de scopeDepth is stored within the ParseState, then calling `withOrder` with a different parseGraph, should actually reset the scopeDepth. Since it is dependendant on the parseGraph it makes more sense to store it there.


core/src/test/java/io/parsingdata/metal/AutoEqualityTest.java
```diff
@@ -437,4 +436,11 @@ public void basicNoHashCollisions(final Object object, final Object same, final
         }
     }
 
+    private static void assertEquals(final Object o, final Object object) {
```
mvanaken, line 439: Decided to override the assertEquals methods, so we get more meaningful error messages when a test fails. You will now see which objects causes a test to fail, which previously could only be discovered by debugging the test, which takes much more time.

core/src/main/java/io/parsingdata/metal/data/ParseGraph.java
```diff
@@ -47,52 +48,61 @@ public class ParseGraph extends ImmutableObject implements ParseItem {
...
+            return new ParseGraph(head.asGraph().closeBranch(token), tail, definition, true, newScopeDepth);
+        }
+        if (!head.getDefinition().equals(token)) {
+            throw new IllegalStateException("Cannot close branch with token that does not match its head token.");
+        }
+        if (newScopeDepth != 0) {
+            throw new IllegalStateException("Cannot close parse graph that has a non zero scopeDepth.");
         }
-        return new ParseGraph(head, tail, definition, false);
+        return new ParseGraph(head, tail, definition);
```
jvdb, line 105: Nice change, to call the other constructor here (to make the `false` implicit).

core/src/main/java/io/parsingdata/metal/data/ParseGraph.java
```diff
@@ -140,6 +150,7 @@ public boolean equals(final Object obj) {
             && Objects.equals(head, ((ParseGraph)obj).head)
             && Objects.equals(tail, ((ParseGraph)obj).tail)
             && Objects.equals(branched, ((ParseGraph)obj).branched)
+            && Objects.equals(scopeDepth, ((ParseGraph)obj).scopeDepth)
```
jvdb, line 153: The `scopeDepth` is essentially _derivable_ right? Maybe it should be considered cached then as well just like `size`?

core/src/main/java/io/parsingdata/metal/data/ParseState.java
```diff
@@ -42,63 +42,61 @@ public class ParseState extends ImmutableObject {
     public final Source source;
     public final ImmutableList<ImmutablePair<Token, BigInteger>> iterations;
     public final ImmutableList<ParseReference> references;
-    public final int scopeDepth;
 
-    public ParseState(final ParseGraph order, final ParseValueCache cache, final Source source, final BigInteger offset, final ImmutableList<ImmutablePair<Token, BigInteger>> iterations, final ImmutableList<ParseReference> references, final int scopeDepth) {
```
jvdb, line 47: Cleans up `ParseState` a lot to move `scopeDepth` to `ParseGraph`!

core/src/test/java/io/parsingdata/metal/data/ParseGraphTest.java
```diff
@@ -236,4 +246,130 @@ public void testCurrent() {
         assertFalse(EMPTY.addBranch(NONE).current().isPresent());
     }
 
+    public static Stream<Arguments> scopeDepthTest() {
```
jvdb, line 249: Wow, amazing work on these extensive tests!

jvdb (approved):

core/src/main/java/io/parsingdata/metal/data/ParseGraph.java
mvanaken: With normal usage, it should be derivable. However, it is possible to create a token that alternates its scope delimiter. This makes it possible to have two identitcal ParseGraphs with different scope depths. This is only theoretical and I do not know why you would create a token like that, but for robustness it is better to keep this here. Other minor reason to keep it here is so we can test equality for the ParseGraph with the AutoEqualityTest.

